PR #5, opened 2017-06-13
Add `cacert_file` and `insecure` options to the provider.

richardc:
These settings can be used to configure the level of SSL verification
performed by the provider.

* `cacert_file` allows for the provision of an alternate ca cert, for
locally-signed or self-signed operations.
* `insecure` allows SSL verification to be turned off completely.  It is
not suggested as a first option.

Targets https://github.com/terraform-providers/terraform-provider-gitlab/issues/4

### DIFF
--- a/gitlab/config.go
+++ b/gitlab/config.go
@@ -1,18 +1,49 @@
 package gitlab
 
 import (
+	"crypto/tls"
+	"crypto/x509"
+	"io/ioutil"
+	"net/http"
+
 	"github.com/xanzy/go-gitlab"
 )
 
 // Config is per-provider, specifies where to connect to gitlab
 type Config struct {
-	Token   string
-	BaseURL string
+	Token      string
+	BaseURL    string
+	Insecure   bool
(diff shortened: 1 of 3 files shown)


richardc:
So I'm a little confused, if this has two approvals, why's it not been merged?

tombuildsstuff:
@richardc in general we tend to approve and then let the original contributor merge (when they have permission) - but I'll merge this now :)

richardc:
@tombuildsstuff is that flow documented somewhere?  I don't see it in the README.md and there's no CONTRIBUTING.md

tombuildsstuff:
@richardc I've found this in the main repository: https://github.com/hashicorp/terraform/blob/master/docs/maintainer-etiquette.md#merging (it should probably be linked from the CONTRIBUTING.md in each provider though) :)

roidelapluie:
https://github.com/hashicorp/terraform/blob/master/docs/maintainer-etiquette.md

PR #6, opened 2017-06-13
Fix bad boilerplate in README

richardc:
The clone source and destination should be in terms of
`terraform-providers` not `hashicorp`


stack72:
LGTM! thanks for this :)

PR #7, opened 2017-07-06
Meatball fix of gitlab_project_hook tests.

richardc:
Here we treat gitlab_project_hook#build_events as optional, which they
were before, it was just more succinct in the code to act as though they
weren't, as it's a simple boolean attribute..

Also we stop testing for it so it won't 500 the gitlab API between 9.3.0
and release containing
https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/12673

This won't fix the underlying issue that the gitlab 9.3.0 broke the
handling of the build_events attribute of project hooks, but it will
allow the CI to run green again.


richardc:
@stack72 this should fix the immediate ci failing, but leaves the user experience when attempting to manage gitlab_project_hook#build_events against unpatched gitlab to be one of a 500 error bubbled at runtime.  I have an idea about sniffing versions to allow for a better experience here, which will be another PR

roidelapluie:
This can be closed without being merged because it has been fixed by gitlab for a while

richardc:
Agreed; a later development to add a "don't try this between versions x and y" is probably still useful, but the fix I submitted has been released upstream I'm closing this.

PR #19, opened 2017-09-22
gitlab_project: wait until real deletion

roidelapluie:
Deleting a project in Gitlab is an async method.

By blocking until the project does not exist anymore, we
do not lie to the terraform end user about the status of a
deleted project.

As a side effet it fixes the acceptances tests.

Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>

PR #24, opened 2017-09-26
Update changelog with recent changes

roidelapluie:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,24 @@
-## 0.1.1 (Unreleased)
+# 0.2.0 (Unreleased)
+
+BACKWARDS INCOMPATIBILITIES:
+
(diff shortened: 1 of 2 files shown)


radeksimko:
Thanks for adding this retrospectively 👍 

The format which is recognized by our release script is `[GH-19]` rather than `(#19)` - mind changing that?

Just FYI for the future - as mentioned in the etiquette - it is totally fine to use the Github's web interface to update changelog. If the PR here becomes stale it would be in conflict with some other changelog PRs. This is also why we discourage anyone from ever modifying changelog as part of their PR. Reviews may take time and changelog should always be kept up to date + there isn't much that can go wrong and if it does we can fix easily, so the risk of bypassing review process in this context is low.

roidelapluie:
Ok, I'll edit gradually as we get the PR merged. Currently waiting for #19 to get a review after my last rewrite

PR #27, opened 2017-09-27
Fail when using api v3

roidelapluie:
This will help our users to figure out that we moved to the v4 api.

### DIFF
--- a/gitlab/provider.go
+++ b/gitlab/provider.go
@@ -1,6 +1,9 @@
 package gitlab
 
 import (
+	"fmt"
+	"strings"
+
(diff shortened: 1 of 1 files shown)


richardc:
The idea is sound, but I think this should be a ValidateFunc on the schema over in provider.go https://github.com/terraform-providers/terraform-provider-gitlab/blob/master/gitlab/provider.go#L20

https://github.com/hashicorp/terraform/blob/master/helper/schema/schema.go#L186

roidelapluie:
Thanks! I fixed it to use ValidateFunc

Issue #36, opened 2017-10-22
[Feature Request] Make Gitlab resources importable


blaggacao:
### Affected Resource(s)
- gitlab_*

I'm a latecomer in the terraform ecosystem. I wonder, without having done a deep dive if it would be as simple as just implementin those lines:
https://github.com/terraform-providers/terraform-provider-fastly/blob/2dc6f92829de934b54ba7bdcf7f4f7a7aae6a9a2/fastly/resource_fastly_service_v1.go#L24-L26

Docs on resource importer:
https://github.com/hashicorp/terraform/blob/351c6bed79abbb40e461d3f7d49fe4cf20bced41/helper/schema/resource_importer.go

If would be really nice to have that feature. I'm unfortunately not really firm in go programming, else I'd have already done a PR.

richardc:
I think that that was already done in #30 for `gitlab_project`, but reading #35 this seems like it was necessary but insufficient, so the fix in #35 is needed across all types.  I don't have a development environment configured to verify this right now, but I should have the bandwidth to look at this in the next few days.

blaggacao:
Cool, sounds promising! Thank you!

richardc:
In part this is a case of 'if it's not documented, it doesn't exist', as we had it in a form for project and group.

Here's what is left over from the existing resources:

- [x] `gitlab_project` (Documented and tweaked by #37)
- [ ] `gitlab_group` (Documented and tweaked by #38)
- [ ] `gitlab_deploy_key`
- [ ] `gitlab_label`
- [ ] `gitlab_project_hook`
- [ ] `gitlab_user`

exarkun:
It looks like #39 was merged, although I'm not sure it works:
```
$ terraform import gitlab_user.xxx_xxx 14                                    
Acquiring state lock. This may take a few moments...                                                                                             
gitlab_user.xxx_xxx: Importing from ID "14"...
                                                        
Error: rpc error: code = Unavailable desc = transport is closing
          

Releasing state lock. This may take a few moments...
```

timofurrer:
For the resources which return a secret / token only after creation we should implement the importer nonetheless, but omit that specific attribute only. See https://github.com/gitlabhq/terraform-provider-gitlab/pull/860 (or `gitlab_group_access_token` resource once the PR is merged :)

Issue #35, opened 2017-10-09
Incorrect ID when Importing Groups


rawkode:
### Terraform Version
v0.10.8

### Affected Resource(s)
- gitlab_group
- gitlab_project

### Expected Behavior
When importing a project or group, the ID should be set to the ID provided by GitLab

### Actual Behavior
The ID is set to the path from `statePassThrough`

### Steps to Reproduce
1. Create a GitLab group
2. Add resource to `.tf`
3. Import

### Important Factoids
This renders importing a group and then using it has a sub-group as useless, as it fails to convert the alphanumeric "id" to an integer.

rawkode:
Hi @radeksimko, what would be the correct way to fix this? I have a working copy that does as I expect; but it doesn't feel like the correct solution:

```
commit c5ac978854cc559a3dc89f4b7e117fd7f0b875c6 (HEAD -> bugfix/import-id)
Author: David McKay <david.andrew.mckay@gmail.com>
Date:   Fri Oct 20 14:54:23 2017 +0100

    Ensuring that the IDs are set properly

diff --git a/gitlab/resource_gitlab_group.go b/gitlab/resource_gitlab_group.go
index 5097bca..572cebf 100644
--- a/gitlab/resource_gitlab_group.go
+++ b/gitlab/resource_gitlab_group.go
@@ -109,6 +109,7 @@ func resourceGitlabGroupRead(d *schema.ResourceData, meta interface{}) error {
                return err
        }
 
+       d.SetId(fmt.Sprintf("%d", group.ID))
        d.Set("name", group.Name)
        d.Set("path", group.Path)
        d.Set("description", group.Description)
diff --git a/gitlab/resource_gitlab_project.go b/gitlab/resource_gitlab_project.go
index 1581ebf..ab56c6f 100644
--- a/gitlab/resource_gitlab_project.go
+++ b/gitlab/resource_gitlab_project.go
@@ -94,6 +94,7 @@ func resourceGitlabProject() *schema.Resource {
 }
 
 func resourceGitlabProjectSetToState(d *schema.ResourceData, project *gitlab.Project) {
+       d.SetId(fmt.Sprintf("%d", project.ID))
        d.Set("name", project.Name)
        d.Set("path", project.Path)
        d.Set("description", project.Description)

```

richardc:
That looks like the correct fix to me; because otherwise the state isn't going to be updated with the loaded id the underlying resource.

richardc:
[...]

PR #37, opened 2017-10-23
Fix and document importing projects

richardc:
As noted in #35, when importing a project we were not updating the ID to
be the primary key id, so if a project was imported with `terraform
import gitlab_project.foo user/foo` it was kept in the state as
`user/foo` rather than the underlying id.

Here we fix that up, and add some documentation indicating that this
resource is importable (towards #36)


roidelapluie:
lgtm

PR #39, opened 2017-10-24
Add support for importing users

richardc:
Here we add basic support for importing users to state.   Ideally we
would be able to import by username, as there is an api to find users by
username[1], but this is not currently exposed by go-gitlab[2]

Towards #36

[1] https://docs.gitlab.com/ee/api/users.html#list-users documents
	/users?username=:username which worked well in manual testing.

[2] https://github.com/xanzy/go-gitlab/blob/master/users.go#L97-L99


kkirsche:
Is this waiting on anything? Seems to have stalled

PR #41, opened 2017-11-23
Fixes error when updating default branch

ThatGerber:
Updating the default branch caused the build to fail because it was inadvertently set to update the branch to the project description.

I attempted to write some tests for this case, but it's quite difficult. Gitlab returns `null` for the default branch until the first commit is pushed up to a repo, and the default branch can't be changed to another branch until that branch exists in Gitlab. However, by default, the `master` branch is the default branch when creating a new project. All makes sense, but, writing a test case means that a project needs to be created, the repo needs to exist and it needs to have multiple branches. I can help try and write it, but it's hard to fit it all into a single test step.

### DIFF
--- a/gitlab/resource_gitlab_project.go
+++ b/gitlab/resource_gitlab_project.go
@@ -183,7 +183,7 @@ func resourceGitlabProjectUpdate(d *schema.ResourceData, meta interface{}) error
 	}
 
 	if d.HasChange("default_branch") {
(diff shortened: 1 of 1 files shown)


ThatGerber:
Is this project being maintained at all? This MR has been here for two months without a comment and it doesn't look like there's been anything done on the project since the last updates.

roidelapluie:
Thanks!

PR #70, opened 2018-12-08
fix: data resources

adam-janis:
1. **gitlab_project**
data parameters were not passed into terraform state, they are being set in `dataSourceGitlabProjectRead` but ignored if not in Schema
2. **gitlab_user**
`d.Set("userName"` needs to match Schema


roidelapluie:
Thanks!

adam-janis:
@roidelapluie do you have any idea when this could be released? how can we help to release a new version of this provider?

PR #65, opened 2018-10-16
WIP: Implement gitlab service slack

hypnoglow:
Fixes #32 

This is WIP for now, as I haven't added any tests/documentation yet. But I'm ready for an initial review.

Also I'm not sure I've correctly updated vendor packages, because I've no idea how to use govendor properly.

PR #96, opened 2019-03-04
Implement gitlab service slack

jorcau:
Hi,

Here is a new proposal for Slack Integration implementation. It reuses work started on #65. I've talked with @hypnoglow about doing another PR so I guess we could close #65 for this one.

It includes documentation and acceptance tests:
```
➜ make testacc TEST=./gitlab TESTARGS='-run=TestAccGitlabServiceSlack_*'
TF_ACC=1 go test ./gitlab -v -run=TestAccGitlabServiceSlack_* -timeout 120m
=== RUN   TestAccGitlabServiceSlack_basic
--- PASS: TestAccGitlabServiceSlack_basic (8.23s)
=== RUN   TestAccGitlabServiceSlack_import
--- PASS: TestAccGitlabServiceSlack_import (6.95s)
PASS
ok  	github.com/terraform-providers/terraform-provider-gitlab/gitlab	15.209s
```

Thanks,


roidelapluie:
FYI now the Acc tests are run in travis too.

roidelapluie:
Can you please add an ACC test with default values? Current tests are great but too explicit.

jorcau:
> FYI now the Acc tests are run in travis too.

This is awesome!

I added a test with only required field set as you asked.

website/docs/r/service_slack.html.markdown
```diff
@@ -0,0 +1,86 @@
...
+* `wiki_page_events` - (Optional) Enable notifications for wiki page events.
+
+* `wiki_page_channel` - (Optional) The name of the channel to receive wiki page events notifications.
+
+## Importing projects
+
+You can import a service_slack state using `terraform import <resource> <project>`.  The
+`project` is the id of a project:
+
+    terraform import gitlab_service_slack.awesome_project 1
```
roidelapluie: maybe use  `terraform import gitlab_service_slack.slack 1` would be less confusing?

website/docs/r/service_slack.html.markdown
```diff
@@ -0,0 +1,86 @@
...
+  visibility_level = "public"
+}
+
+resource "gitlab_service_slack" "slack" {
+  project                    = "${gitlab_project.awesome_project.id}"
+  webhook                    = "https://webhook.com"
+  username                   = "myuser"
+  push_events                = true
+  push_channel               = "push_chan"
+  issues_events              = true
```
roidelapluie: what happens if you set this and not issue_channel?

website/docs/r/service_slack.html.markdown
roidelapluie: Also please use a code block (```)

jorcau: This was a mistake! Good catch, thanks!

website/docs/r/service_slack.html.markdown
jorcau: This is valid and would work but you're right that it's not really clear, so I made the change.

gitlab/resource_gitlab_service_slack.go
```diff
@@ -0,0 +1,243 @@
...
+
+	log.Printf("[DEBUG] delete gitlab slack service for project %s", project)
+
+	_, err := client.Services.DeleteSlackService(project)
+	return err
+}
+
+func resourceGitlabServiceSlackImportState(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	d.Set("project", d.Id())
+
```
roidelapluie: shouldnt we fetch the service at this point from gitlab?

jorcau: Not sure what you mean here.

From what I understand, the `resourceGitlabServiceSlackImportState` function should just define the ID or necessary parameters so the READ function can then be called to actually fetch the resource's details.

rsrchboy:
@roidelapluie Are there any blockers to merging this?

jorcau:
Hi @roidelapluie, do you have time to review & merge this anytime soon? Thanks!

PR #100, opened 2019-03-08
Update index.html.markdown

cherwin:
Using the current provider with a base_url ending in api/v3/ will result in an error message Instructing the user to upgrade to v4.


roidelapluie:
Thanks

PR #102, opened 2019-03-08
Update user datasource: allow searching by id and add attributes

jorcau:
Hi,

This PR updates the *user* datasource.

It allows searching by id:
```hcl
data "gitlab_user" "foo" {
  user_id = "2387"
}
```

It also adds more attributes and better handles the search parameters to avoid using multiple ones at the same time.

**Note**: the search by id parameter is called `user_id` because `id` is a Terraform reserved keyword. Same goes for the `provider` attribute renamed into `user_provider`. I'm willing to change this if you have better suggestions.

Best,


roidelapluie:
Thanks!

PR #108, opened 2019-03-22
website: make example server URL non-clickable

nfagerlund:
The site's markdown processor tries to automatically turn URLs into links, but this one
doesn't go anywhere, so let's style it as a code span.


roidelapluie:
Thanks!

PR #127, opened 2019-05-06
add 'reset_password' arg to user resource

UrosSimovic:


roidelapluie:
Can we have "null" as default? so that it would be true on user creation.

UrosSimovic:
> Can we have "null" as default? so that it would be true on user creation.

Sorry, but don't know what you mean with that.

roidelapluie:
I would like that by default we do not change that option

UrosSimovic:
Please show me an example how/what to do.

UrosSimovic:
I can remove default value. But wouldn't be a bit inconsistent at the end, if you look at `is_external` parameter, which is also optional and omitempty? I just copied `is_external` and renamed it.

UrosSimovic:
I removed default value, so after upgrading the provider there won't be state change. 

Regarding tests, this field is somehow special, edge case. You can only set it on create, you can't update it, and you don't get it in response. So don't know what to test.

matejvelikonja:
@roidelapluie this cannot be tested

roidelapluie:
Can we still add two tests: creating a user with and without that attribute ?

thanks

roidelapluie:
even if we do not check the result because we can't

Toady00:
I was going to create an issue requesting this feature but I found this. Can I ask what the status is here? This feature would be really helpful. As of right now the only way we can get a new user into out Gitlab is to create a random password when creating the user, they then get an email, but they have to go through the forgotten password flow to login.

roidelapluie:
we just need an acc test

Toady00:
@UrosSimovic is it possible to add the 2 tests that @roidelapluie requested so this can get merged? Just a creating a user with and without that attribute?

UrosSimovic:
Will this be ok? I added the test with `reset_password = true` attribute. Without attribute is already done.

PR #130, opened 2019-05-07
Stop silently deleting objects from state

roidelapluie:

### DIFF
--- a/gitlab/resource_gitlab_branch_protection.go
+++ b/gitlab/resource_gitlab_branch_protection.go
@@ -87,14 +87,8 @@ func resourceGitlabBranchProtectionRead(d *schema.ResourceData, meta interface{}
 
 	log.Printf("[DEBUG] read gitlab branch protection for project %s, branch %s", project, branch)
 
-	pb, response, err := client.ProtectedBranches.GetProtectedBranch(project, branch)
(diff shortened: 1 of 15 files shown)


roidelapluie:
@attachmentgenie @jorcau can you review this?

jorcau:
I'm not sure to understand why you want to avoid this behavior. It seems okay to me to remove from state resources that no longer exist. Am I missing something?

roidelapluie:
Yes. Gitlab will return 404 if the user has no right so see the resources (instead of 404).

If people use their own oauth token to run this it will be an issue.

jorcau:
Ok, it makes sense then!

I'm just a bit worried about the behavior in case resources have been removed using the UI. If I'm not making a mistake, it means that all Terraform actions will fail unless something like a `terraform state rm <removed_resource>` is done, right?

Knowing this, I'd rather like to keep the custom error message on 404 errors and make it more explicit about the 404 / 403 Gitlab errors meaning, so users know what's happening.

jakubknejzlik:
@jorcau imho it's standard behaviour in other providers ("if you remove it manually, you also have to fix the TF state").

roidelapluie:
I will merge it and mark it as breaking in the release notes.

PR #132, opened 2019-05-09
Cleans up "project" resource schema and testing framework

ThatGerber:
Uses the *gitlab.Project struct and resource Schema to test the resulting values.

Removes the extra generate access token file and updates the access token value to be generated from a variable (with a default).

### DIFF
--- a/gitlab/resource_gitlab_project.go
+++ b/gitlab/resource_gitlab_project.go
@@ -11,6 +11,134 @@ import (
 	gitlab "github.com/xanzy/go-gitlab"
 )
 
+var resourceGitLabProjectSchema = map[string]*schema.Schema{
+	"name": {
+		Type:     schema.TypeString,
+		Required: true,
+	},
+	"path": {
+		Type:     schema.TypeString,
+		Optional: true,
+		DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+			if new == "" {
+				return true
+			}
+			return old == new
+		},
+	},
+	"namespace_id": {
+		Type:     schema.TypeInt,
+		Optional: true,
+		ForceNew: true,
+		Computed: true,
+	},
+	"description": {
+		Type:     schema.TypeString,
+		Optional: true,
+	},
+	"default_branch": {
+		Type:     schema.TypeString,
+		Optional: true,
+	},
+	"issues_enabled": {
+		Type:     schema.TypeBool,
+		Optional: true,
+		Default:  true,
+	},
+	"merge_requests_enabled": {
+		Type:     schema.TypeBool,
+		Optional: true,
+		Default:  true,
+	},
+	"approvals_before_merge": {
+		Type:     schema.TypeInt,
+		Optional: true,
+		Default:  0,
+	},
+	"wiki_enabled": {
+		Type:     schema.TypeBool,
+		Optional: true,
+		Default:  true,
+	},
+	"snippets_enabled": {
+		Type:     schema.TypeBool,
+		Optional: true,
+		Default:  true,
+	},
+	"container_registry_enabled": {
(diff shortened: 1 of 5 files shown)


scripts/start-gitlab.sh
```diff
@@ -1,16 +1,31 @@
 #!/bin/bash -e
-if [[ $MAKE_TARGET != "testacc" ]]; then echo "not starting gitlab!"; exit 0; fi
+test "$MAKE_TARGET" == "testacc" || { echo "not starting gitlab!"; exit 0; }
 echo "Starting gitlab container..."
-docker run -d -e GITLAB_ROOT_PASSWORD=adminadmin --rm -p 127.0.0.1:8080:80 --name gitlab gitlab/gitlab-ce
+docker run -d --rm --name gitlab \
+  -e GITLAB_ROOT_PASSWORD=adminadmin \
```
roidelapluie: Do you want this as a variable as well?

gitlab/resource_gitlab_project.go
```diff
@@ -11,6 +11,134 @@ import (
...
+			Schema: map[string]*schema.Schema{
+				"group_id": {
+					Type:     schema.TypeInt,
+					Required: true,
+				},
+				"group_access_level": {
+					Type:     schema.TypeString,
+					Required: true,
+					ValidateFunc: validation.StringInSlice([]string{
+						"no one", "guest", "reporter", "developer", "maintainer"}, false),
```
roidelapluie: Isn't there a slice in the gitlab library for this?

gitlab/resource_gitlab_project.go
```diff
@@ -11,6 +11,134 @@ import (
...
+	},
+	"container_registry_enabled": {
+		Type:     schema.TypeBool,
+		Optional: true,
+		Default:  true,
+	},
+	"visibility_level": {
+		Type:         schema.TypeString,
+		Optional:     true,
+		ValidateFunc: validation.StringInSlice([]string{"private", "internal", "public"}, true),
```
roidelapluie: Isn't there a slice in the gitlab library for this?

gitlab/resource_gitlab_project.go
```diff
@@ -11,6 +11,134 @@ import (
...
+		Default:      "private",
+	},
+	"merge_method": {
+		Type:         schema.TypeString,
+		Optional:     true,
+		ValidateFunc: validation.StringInSlice([]string{"merge", "rebase_merge", "ff"}, true),
```
roidelapluie: Isn't there a slice in the gitlab library for this?

ThatGerber: The values are stored in the library as type `gitlab.MergeMethodValue`. Would need a method to convert them to type `gitlab.MergeMethodValue` and then another possibly to check value against the type.

gitlab/resource_gitlab_project.go
ThatGerber: The values are stored in the library as type `gitlab.VisibilityValue`. Would need a method to convert them to type `gitlab.VisibilityValue` and then another to check value against the type.

gitlab/resource_gitlab_project.go
ThatGerber: The values are stored in the library as type `gitlab.AccessLevelValue`. Would need a method to convert them to type `gitlab.AccessLevelValue` and then another to check value against the type.

scripts/start-gitlab.sh
ThatGerber: It could potentially be. For the sake of automated tests it's not even necessary to set the root password since we're not even accessing the UI, but it might be useful if someone wasn't deleting the test resources and was trying to log in afterwards to view them.

roidelapluie:
Thanks

PR #137, opened 2019-05-24
Add support for managed clusters

bdimcheff:
The gitlab api has a "managed" field to disable cluster management features when adding a kubernetes cluster. This implements that field. Default is true if you don't supply it.

This can only be toggled at cluster creation time, so changes will delete and readd the integration.

https://docs.gitlab.com/ee/api/project_clusters.html#add-existing-cluster-to-project

This also upgrades go-gitlab to an unreleased version that includes this field, which may be problematic.

I've run this fork for a few days successfully, and it has behaved as I expect it to.

### DIFF
--- a/gitlab/resource_gitlab_project_cluster_test.go
+++ b/gitlab/resource_gitlab_project_cluster_test.go
@@ -21,7 +21,22 @@ func TestAccGitlabProjectCluster_basic(t *testing.T) {
 		Steps: []resource.TestStep{
 			// Create a project and cluster with default options
 			{
-				Config: testAccGitlabProjectClusterConfig(rInt),
+				Config: testAccGitlabProjectClusterConfig(rInt, true),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGitlabProjectClusterExists("gitlab_project_cluster.foo", &cluster),
+					testAccCheckGitlabProjectClusterAttributes(&cluster, &testAccGitlabProjectClusterExpectedAttributes{
+						Name:                        fmt.Sprintf("foo-cluster-%d", rInt),
+						Domain:                      "example.com",
+						EnvironmentScope:            "*",
+						KubernetesApiURL:            "https://123.123.123",
+						KubernetesCACert:            projectClusterFakeCert,
+						KubernetesAuthorizationType: "abac",
+					}),
+				),
+			},
+			// create an unmanaged cluster
+			{
+				Config: testAccGitlabProjectClusterConfig(rInt, false),
 				Check: resource.ComposeTestCheckFunc(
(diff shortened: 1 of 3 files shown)


roidelapluie:
I would love to see more test coverage.

bdimcheff:
@roidelapluie me too - I took a look and didn't see anything that was easy for me to base my tests off of. This field has similar semantics to "enabled", which is also untested. It is only usable in `AddClusterOptions` because it can only be set on create. Its value isn't exposed via the API, so there's no way to check if it was set properly. If there's another test from a different module that I should look at for ideas, let me know. I'm happy to try to write a test, I just couldn't figure out how to write a useful one.

matejvelikonja:
@bdimcheff  you are correct. this is impossible to test. I had the same issues with enabled flag.

You cannot event read the value, so you cannot test if it changed correctly.

matejvelikonja:
BTW: i suggest you split remove the vendor update commit, update will be covered by #143

bdimcheff:
@matejvelikonja will do, thanks. Should I rebase it out or do you avoid force pushes to feature branches?

bdimcheff:
ok this is just a single commit on top of  #143 now so the diff should be nice once that's merged.

website/docs/r/project_cluster.html.markdown
```diff
@@ -46,6 +46,8 @@ The following arguments are supported:
 
 * `enabled` - (Optional, boolean) Determines if cluster is active or not. Defaults to `true`.
 
+* `managed` - (Optional, boolean) Determines if cluster is managed by gitlab or not. Defaults to `true`.
```
roidelapluie: Please make a comment that we can not read that value in the doc

gitlab/resource_gitlab_project_cluster_test.go
```diff
@@ -85,7 +85,7 @@ func TestAccGitlabProjectCluster_import(t *testing.T) {
 				ResourceName:            "gitlab_project_cluster.foo",
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"enabled", "kubernetes_token"},
+				ImportStateVerifyIgnore: []string{"enabled", "kubernetes_token", "managed"},
```
roidelapluie: how hard is it to add a test that creates a cluster managed and one unmanaged, even if we can not read the status?

website/docs/r/project_cluster.html.markdown
bdimcheff: will do. FYI: enabled works this way too, so I'll add it there too.

gitlab/resource_gitlab_project_cluster_test.go
bdimcheff: ok I added a test but I'm not sure it really tests much

Issue #42, opened 2018-01-18
[feature request] support importing projects in the `gitlab_project` resource


ketzacoatl:
Hi,

I am looking at using the `gitlab` provider to manage a variety of projects, and wondering how reasonable it is to use this resource when the project needs to be imported. EG, in gitlab, when you create a new project, you can select to import that project from an HTTP or git/etc URL. When creating the project with the Terraform resource, I would like to be able to specify the URL the project should be initially imported from. This import is a one-time action, so IDK if it makes sense for the provider to support this, but it seems worth asking about.

Thanks!

danielriesner:
#142 resolves this.

baurmatt:
@roidelapluie Could you please close this issue? :)

PR #142, opened 2019-06-03
Add import_url attribute to project resource to allow importing repository from URL.

danielriesner:
This resolves issue #42 .

Although I succesfully tested it with a locally running gitlab instance, unfortunately for now acceptance tests for import_url attribute are missing. From my understanding the requirements for such tests to fit the other tests approach are:
1. Before creating actual test gitlab project, create dummy gitlab project, which can be imported in the actual test.
2. Run actual test.
3. Check if import succeeded.
4. Remove dummy gitlab project.

Apart from that an already existing git repo url could be used for the test e.g. this repo itself. Although still a check would be needed to see if import was successful. 

What do you guys suggest?

### DIFF
--- a/gitlab/resource_gitlab_project.go
+++ b/gitlab/resource_gitlab_project.go
@@ -76,6 +76,10 @@ var resourceGitLabProjectSchema = map[string]*schema.Schema{
 		ValidateFunc: validation.StringInSlice([]string{"private", "internal", "public"}, true),
 		Default:      "private",
 	},
+	"import_url": {
+		Type:     schema.TypeString,
+		Optional: true,
+	},
 	"merge_method": {
 		Type:         schema.TypeString,
 		Optional:     true,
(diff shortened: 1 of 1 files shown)


roidelapluie:
I would suggest that you maybe use the import feature with this repo, then check with the file api https://github.com/xanzy/go-gitlab/blob/master/repository_files.go that there is a gitlab/resource_gitlab_project.go  file in the repository?

danielriesner:
> I would suggest that you maybe use the import feature with this repo, then check with the file api https://github.com/xanzy/go-gitlab/blob/master/repository_files.go that there is a gitlab/resource_gitlab_project.go file in the repository?

Alright thanks for the input! I will try to do it that way and add it to this pr.

danielriesner:
This pr is superseded by pr https://github.com/terraform-providers/terraform-provider-gitlab/pull/164 .

PR #144, opened 2019-06-03
Avoid switching the default branch to null if omitted

randomswdev:
If a gitlab_project resource does not include the default_branch attribute, terraform would set the default branch to null.
The scenario is as follows

1. create a terraform file defining a gitlab_project without the default_branch attribute
2. execute terraform apply to create the project in GitLab
3. push code to the new project, thus creating the master branch. This would make the master branch the defualt branch.
4. run again terraform apply. Terraform would reset the default branch, setting it to null.

Setting the default branch equal to master in the terraform file, would make the first terraform apply fail. This is because terraform would attempt to set the defult branch to master, but this branch does not exist yet.

### DIFF
--- a/gitlab/resource_gitlab_project.go
+++ b/gitlab/resource_gitlab_project.go
@@ -39,6 +39,19 @@ var resourceGitLabProjectSchema = map[string]*schema.Schema{
 	"default_branch": {
 		Type:     schema.TypeString,
 		Optional: true,
+		DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+			// If the old default branch is empty, it means that the project does not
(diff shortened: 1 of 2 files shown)


matejvelikonja:
I suggest you add some tests for this functionality.

roidelapluie:
Can't we set that to "computed" instead?

roidelapluie:
Or setting the default to "master"

randomswdev:
I’ll add the tests and evaluate the possibility to make the field “computed”. But it will take me few days to resume work on this merge request.

randomswdev:
I changed the implementation with respect to my original idea. So, to avoid confusion, I'm going to close this merge request and open a new one.

Issue #141, opened 2019-06-01
[Feature request] Enable archiving of project instead of deletion when specified


dihedron:
## Short description
It would be nice if the Gitlab provider gave a way to *archive* an old project instead of always destroying it. 

## Rationale
Destroying a repository is an irreversible operation and Gitlab itself requires special handling such as typing the name of the repo to confirm that you really, *really* want to destroy it since there's no going back.
If you value what's inside a project, archiving it is the safest way of handling its removal; moreover I expect archiving to be *by far* the most common way of removing something, especially in enterprise contexts.

## How to address the issue
It would be nice if a project could be marked with a special attribute (say, "on_delete_archive"), whose default "false" preserves the current destructive behaviour and when explicitly set to true instructs the provider to archive a project that is found missing from the .tf file instead of destroying it forever.
The attribute could be toggled at will without affecting an existing project (it is not stored on the server, it's only in the tfstate); the last value before removing the "gitlab_project" HCL block would determine the provider's behaviour.
***NOTE: I have already implemented this feature, it's extremely simple and works as expected; I still want to run a few additional tests with no haste, and check if some kind of "DiffSuppress" on the new attribute would make the whole behaviour more streamlined, then I'll submit a PR.*** (see https://github.com/dihedron/terraform-provider-gitlab/commit/735e431c560a946b0b582d3d55c773e26cd730f7 )

## Terraform Version
Not an issue: independent of terraform version.

## Affected Resource(s)
Only affected resource is `gitlab_project`

## Terraform Configuration Files
Something like this:
```hcl
provider "gitlab" {
    base_url = "http://localhost/api/v4/"
    token = "<my token>"
}

# missing attribute: defaults to false, deleted (irreversibly destroyed) as usual
resource "gitlab_project" "sample_project_1" {
    name = "example_1"
}                                                                                                                                                                                    

# attribute explicitly set to "true", project archived upon removal
resource "gitlab_project" "sample_project" {
    name = "example_2"
    on_delete_archive = true
}                                                                                                                                                                                    
```

roidelapluie:
I think we would better do it by simply adding an "archived" parameter to the project. That way we also support unarchiving.

dihedron:
I think the `archived` attribute addresses a different use case, which is equally valuable but different.
Explicitly archiving a project is OK and I'm ready to implement it, but it does not save you from inadvertent removal of a valuable asset, which might happen cause people make mistakes. 

[...]

PR #146, opened 2019-06-05
Add "archived" attribute to project resource

dihedron:
As requested by @roidelapluie in the discussion of issue #141, I added the "archived" attribute to the Project resource and managed the lifecycle of the resource accordingly. This pull requests makes a pair with the one providing the possibility to "logically delete" a properly tagged project instead of irreversibly wiping it away from the Gitlab server.

### DIFF
--- a/gitlab/resource_gitlab_project.go
+++ b/gitlab/resource_gitlab_project.go
@@ -142,6 +142,12 @@ var resourceGitLabProjectSchema = map[string]*schema.Schema{
 			},
 		},
 	},
+	"on_delete_archive": {
+		Type:        schema.TypeBool,
+		Description: "Whether the project should be archived instead of being physically deleted.",
+		Optional:    true,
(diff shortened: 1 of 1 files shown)


roidelapluie:
Why do you want that much the on_delete_archived ? this is against terraform way of doing.

People who would really want this would: put archve true then terraform state rm

matejvelikonja:
I agree with @roidelapluie. The `on_delete_archive` is against terraform way.
Terraform should reflect status of your infrastructure.

If you want to archive project, set `archive` attribute instead.

If you are afraid of accidentally removing the project, use meta parameter `prevent_destroy=false` instead.

dihedron:
OK, thanks for your previous feedback. I'll rework this: apart from fixing the failed acceptance tests, I'll remove the `on_delete_archive` and only implement the management of the `archived` attribute. Thank you.

PR #148, opened 2019-06-06
Implement the `archived` attribute on Project data source and resource 

dihedron:
Hallo, I'm proposing in this PR a few changes to enable the management of the `archived` attribute on both the `gitlab_project` resource and data source.
In doing so I had to implement partial state update in the project's update method, since the existing project properties update logic and the new archiving/unarchiving operation require different API calls, which may fail independently. 
In the process I also noticed that the existing code to handle the `shared_with_groups` attribute in the `gitlab_project` resource method does not seem to handle errors which are swallowed altogether. Since this attribute, just like the `archived` attribute, requires its own different API call, I expected it to use partial state updates in order to avoid an inconsistent state in case of failure.
I updated the acceptance tests and the documentation for both resource and datasource.
Hope this helps!

### DIFF
--- a/gitlab/data_source_gitlab_project.go
+++ b/gitlab/data_source_gitlab_project.go
@@ -87,6 +87,11 @@ func dataSourceGitlabProject() *schema.Resource {
 				Optional: true,
 				Computed: true,
 			},
+			"archived": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true,
+			},
 		},
 	}
 }
@@ -118,5 +123,6 @@ func dataSourceGitlabProjectRead(d *schema.ResourceData, meta interface{}) error
 	d.Set("http_url_to_repo", found.HTTPURLToRepo)
 	d.Set("web_url", found.WebURL)
 	d.Set("runners_token", found.RunnersToken)
+	d.Set("archived", found.Archived)
 	return nil
 }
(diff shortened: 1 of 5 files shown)


roidelapluie:
Hello,

SetPartial is great and could be used in create too ?

dihedron:
I'll check about that!

dihedron:
> SetPartial is great and could be used in create too?

OK, as far as I understand, the resource is only committed to Terraform state once you set its ID in the `schema.ResourceData`, which happens quite (maybe too?) late in the `resourceGitlabProjectCreate` function. 

Partial state management shoud still fit in but I feel it is more effective if we move this line:

```golang
d.SetId(fmt.Sprintf("%d", project.ID))
``` 

above the `shared_with_groups` update code, like this:

```golang
	// from this point onwards no matter how we return, resource creation 
	// is committed to state since we set its ID
	d.SetId(fmt.Sprintf("%d", project.ID))

	if v, ok := d.GetOk("shared_with_groups"); ok {
		for _, option := range expandSharedWithGroupsOptions(v) {
			if _, err := client.Projects.ShareProjectWithGroup(project.ID, option); err != nil {
				return err
			}
		}
		d.SetPartial("shared_with_groups")
	}
```

This is a change to code I didn't write and I'd like your advice on this before I submit it. Let me know.
Cheers!

roidelapluie:
Fine for me!

dihedron:
OK, it's in the PR.
While testing that `shared_with_groups` was unaffected by my changes, I noticed the docs reported the `groups_access_level` as optional whereas it's `Required: true` in the schema. I fixed the docs, it's in the same PR.
Cheers.

PR #158, opened 2019-06-25
Do not enforce the default_branch in an empty repository

randomswdev:
Currently, when defining a repository resource, the provider attempts to set a default branch even if the repository is empty (no branch).

Consider the following scenario:
1. Define a gitlab repository resource and set the default_branch attribute to "master"
2. Run terraform apply; terraform creates the repository, but gitlab sets the default branch to null, because the repository is empty (no branch has been created yet)
3. run again terraform apply; the terraform provider refreshes the project state and discovers that the default branch is set to null, while the resource description requires master. It tries to update the repository, setting the default branch; but the master branch does not exist, thus the operation fails.

With the changes in this pull request, the provider discards any difference to the default_branch attribute until the attribute value is null on Gitlab.In fact a null value simply means that there is no branch in the Gitlab repository; as soon as the first branch gets created, Gitlab automatically sets the default branch to this initial branch. My understanding is that from that time on, while it will be possible to change the default branch, it will not be possible to change it back to a null value.

I also updated the unit tests to check that: setting default_branch = "master"
1. does not generate any failure during the updates
2. behaves as expected after creating the master branch (i.e. the first branch) in the repository.

### DIFF
--- a/gitlab/resource_gitlab_project.go
+++ b/gitlab/resource_gitlab_project.go
@@ -39,6 +39,19 @@ var resourceGitLabProjectSchema = map[string]*schema.Schema{
 	"default_branch": {
 		Type:     schema.TypeString,
 		Optional: true,
+		DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+			// If the old default branch is empty, it means that the project does not
+			// have a default branch. This can only happen if the project does not have
+			// branches, i.e. it is an empty project. In that case it is useless to
+			// try setting a specific default branch (because no branch exists).
+			// This code will defer the setting of a default branch to a time when the
+			// project is no longer empty.
+			if old == "" {
+				return true
+			}
+
+			return old == new
+		},
 	},
 	"issues_enabled": {
 		Type:     schema.TypeBool,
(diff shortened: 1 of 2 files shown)


randomswdev:
Do you have any feedback about this pull request? Let me know if you think that something has ti be changed.

roidelapluie:
why could not we set the attribute to "computed" instead, so that we can still override it on first run?

randomswdev:
> why could not we set the attribute to "computed" instead, so that we can still override it on first run?
While setting the attribute as computed works in the scenario where the user does not set the default_branch attribute, it does not solve the problem in case the user sets the attribute.

Consider this sequence of steps:

1. Create a Terraform file describing a new project, setting default_branch to "master"
2. Run Terraform to create the project. The creation completes successfully
3. Run Terraform again. Terraform fails.

At step 3., Terraform discovers that the default branch in the Terrafrom file is set to master, while on Gitlab it is set to the empty string. So Terrafrom attempts to fix the GitLab configuration by changing the value of the default branch attribute for the project; and it fails because the master branch does not exist.

The code in my pull request avoids this problem: it attempts to set the default_branch attribute of a GitLab project only if the value of the default_branch attribute retrieved by GitLab is not the empty string, i.e. only if at least one branch has been created, meaning that the project has been fully initialized.

randomswdev:
@roidelapluie let me know if, based on my last comment, do you think that the pull request requires some further changes.

roidelapluie:
Thanks!

PR #161, opened 2019-07-14
Add resource for looking up a list of projects

enieuw:
This MR adds two resources:
* gitlab_projects - used to return a list of projects that match specified parameters
* gitlab_group_projects - used to return a list of projects that are in a certain group

This allows you to specify branch protection rules for an entire gitlab group without referencing the individual projects.

### DIFF
--- a/gitlab/data_source_gitlab_group_projects.go
+++ b/gitlab/data_source_gitlab_group_projects.go
@@ -54,7 +54,7 @@ func dataSourceGitlabGroupProjects() *schema.Resource {
 func dataSourceGitlabGroupProjectsRead(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*gitlab.Client)
 
-	log.Printf("[INFO] Reading Gitlab projects")
+	log.Printf("[DEBUG] Reading Gitlab projects in group")
 
 	listGroupProjectsOptions, id, err := expandGitlabGroupProjectsOptions(d)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


gitlab/data_source_gitlab_projects.go
```diff
@@ -0,0 +1,119 @@
...
+				},
+			},
+		},
+	}
+}
+
+func dataSourceGitlabProjectsRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gitlab.Client)
+
+	log.Printf("[INFO] Reading Gitlab projects")
```
roidelapluie: DEBUG

gitlab/data_source_gitlab_group_projects.go
```diff
@@ -0,0 +1,124 @@
...
+				},
+			},
+		},
+	}
+}
+
+func dataSourceGitlabGroupProjectsRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gitlab.Client)
+
+	log.Printf("[INFO] Reading Gitlab projects")
```
roidelapluie: DEBUG

roidelapluie: + this is groups

roidelapluie:
Can you merge both and add an optional group_id to gitlab_groups that would then search only in the group?

wernerb:
We really need this PR. @enieuw are you going to continue this ? 
What needs to happen to make it work. @roidelapluie ? I see your latest comment wants the two commits merged and `group_id` as an optional parameter? This should be straightforward.

ringods:
Implemented in #279.

PR #166, opened 2019-07-30
Fix state not being updated for gitlab_branch_protection

enieuw:
The merge access level and push access level is not being mapped back to the textual forms in the state currently. This causes drift to not be detected by terrafrom refresh actions.

This fix uses the accessLevel map to translate the response from the API back to the format used by Terraform to store the state.

### DIFF
--- a/gitlab/resource_gitlab_branch_protection.go
+++ b/gitlab/resource_gitlab_branch_protection.go
@@ -94,8 +94,8 @@ func resourceGitlabBranchProtectionRead(d *schema.ResourceData, meta interface{}
 
 	d.Set("project", project)
 	d.Set("branch", pb.Name)
-	d.Set("merge_access_level", pb.MergeAccessLevels[0].AccessLevel)
-	d.Set("push_access_level", pb.PushAccessLevels[0].AccessLevel)
+	d.Set("merge_access_level", accessLevel[pb.MergeAccessLevels[0].AccessLevel])
+	d.Set("push_access_level", accessLevel[pb.PushAccessLevels[0].AccessLevel])
 
 	d.SetId(buildTwoPartID(&project, &pb.Name))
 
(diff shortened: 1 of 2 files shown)


roidelapluie:
Can you add a test for it?

enieuw:
Added a test

roidelapluie:
mmmh no, you can run acceptance tests that check for indempotency. Your test is too low level.

enieuw:
The bug is low level - adding a test at a higher level only increases the chance of flakyness of the tests. This test covers scope of the bug and will make sure it doesn't happen again for this resource. Unless you stub out the Gitlab API you'll run into unexpected errors without any changes. Gitlab returns 200 on API calls every now and then which are not done and this will make idempotency tests fail.

roidelapluie:
is there actually a bug then?

enieuw:
I would consider writing an integer(e.g. 20 for a reporter) to the state and trying to read it back as a string("reporter") later a bug yes.

roidelapluie:
Yes but how does that integer come in the state in the normal usage of the provider?

enieuw:
Because this https://github.com/terraform-providers/terraform-provider-gitlab/pull/166/files#diff-366717bfa5a6d0763715dd8a0a662f97L97 does not look up the corresponding string value in the accessLevel slice. During the read a lookup is done against the accessLevelID array here: https://github.com/terraform-providers/terraform-provider-gitlab/blob/f75bf730de02dc46d453136232269d060c198604/gitlab/resource_gitlab_branch_protection.go#L51 which translates it back to values that get sent to Gitlab.

roidelapluie:
Yes but, in which case does the api get us the number? all the cases?

enieuw:
All of them

md5:
We are also seeing this same issue. `gitlab_branch_protection` is basically only useful in its current form for setting up the initial values for the access levels. Conflicting made in the UI are not reversed by a `terraform apply`.

PR #167, opened 2019-07-30
Add project_share_group resource to share projects with groups

enieuw:
This resources allows you to share a project with a group. Not really fond of the name but it follows the description as given here: https://docs.gitlab.com/ee/api/projects.html#share-project-with-group


gordonbondon:
Does not `shared_with_groups` parameter on `gitlab_project` resource already solve this?

enieuw:
It can do the same yes but it requires the project to be in the state of terraform. This decouples the linking resources similar to resources like aws security groups with inline security rules and separate rules.

On 7 Aug 2019, at 20:23, Artem Iarmoliuk <notifications@github.com<mailto:notifications@github.com>> wrote:


Does not shared_with_groups parameter on gitlab_project resource already solve this?

gitlab/resource_gitlab_project_share_group.go
```diff
@@ -0,0 +1,157 @@
...
+		if groupId == v.GroupID {
+			resourceGitlabProjectShareGroupSetToState(d, v, &projectId)
+		}
+	}
+
+	return nil
+}
+
+func projectIdAndGroupIdFromId(id string) (string, int, error) {
+	projectId, groupIdString, err := parseTwoPartID(id)
```
roidelapluie: return directly if err != nil

gitlab/resource_gitlab_project_share_group.go
```diff
@@ -0,0 +1,157 @@
...
+	d.SetId(buildTwoPartID(&projectId, &groupIdString))
+	return resourceGitlabProjectShareGroupRead(d, meta)
+}
+
+func resourceGitlabProjectShareGroupRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gitlab.Client)
+	id := d.Id()
+	log.Printf("[DEBUG] read gitlab project projectMember %s", id)
+
+	projectId, groupId, e := projectIdAndGroupIdFromId(id)
```
roidelapluie: use `err` for errors

gitlab/resource_gitlab_project_share_group.go
```diff
@@ -0,0 +1,157 @@
...
+		return err
+	}
+	return resourceGitlabProjectShareGroupRead(d, meta)
+}
+
+func resourceGitlabProjectShareGroupDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gitlab.Client)
+
+	id := d.Id()
+	projectId, groupId, e := projectIdAndGroupIdFromId(id)
```
roidelapluie: use `err` for errors

gitlab/resource_gitlab_project_share_group.go
```diff
@@ -0,0 +1,157 @@
...
+	"log"
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/xanzy/go-gitlab"
+)
+
+func resourceGitlabProjectShareGroup() *schema.Resource {
+	acceptedAccessLevels := make([]string, 0, len(accessLevelID))
```
roidelapluie: len -1 if we exclude owner?

gitlab/resource_gitlab_project_share_group.go
```diff
@@ -0,0 +1,157 @@
...
+	for k := range accessLevelID {
+		if k != "owner" {
```
roidelapluie: is it needed?

gitlab/resource_gitlab_project_share_group.go
enieuw: Changed it, I now made this a hardcoded list as there were more invalid options in the accessLevelID list. The only options in the hardcoded list now are options which will actually get accepted by the Gitlab API

gitlab/resource_gitlab_project_share_group.go
enieuw: Same as above, hardcoded the list

enieuw:
Processed your comments, I copied quite a bit from the project membership resource which is where the non err variables came from.  Should be resolved now.

PR #169, opened 2019-08-15
Add support for more variable options (variable_type, masked, environment_scope)

### LINKED ISSUE #117: Support for masked variables (gitlab_project_variable)
Gitlab is currently rolling out a `masked` parameter for variables, which allows controlling whether a project variable is masked in the logs.

More info: https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/27236

Would be good to follow suit and add  a `masked` boolean, just like we already have `protected`. (I might work on this patch, no promises)

### LINKED ISSUE #153: [feature request] support group and project variables of type 'file'
GitLab allows settings group and project variables with type 'file'. GitLab CI creates a temporary file with the contents of the variable while the actual env variable gets set to the path to that file. [Gitlab docs](https://docs.gitlab.com/ee/ci/variables/#variable-types).

The gitlab API supports file variables for both [groups](https://docs.gitlab.com/ce/api/group_level_variables.html#update-variable) and [projects](https://docs.gitlab.com/ce/api/project_level_variables.html#create-variable).

Can Terraform support this feature? Would very useful :).

sudoapt-getclean:
closes #153
closes #117


gitlab/resource_gitlab_group_variable.go
```diff
@@ -34,6 +34,12 @@ func resourceGitlabGroupVariable() *schema.Resource {
 				Required:  true,
 				Sensitive: true,
 			},
+			"variable_type": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "env_var",
+				ValidateFunc: StringIsGitlabVariableType(),
```
roidelapluie: Why exposing that?

gitlab/util.go
```diff
@@ -124,6 +137,21 @@ func StringIsGitlabVariableName() schema.SchemaValidateFunc {
 	}
 }
 
+func StringIsGitlabVariableType() schema.SchemaValidateFunc {
+	return func(v interface{}, k string) (s []string, es []error) {
+		value, ok := v.(string)
```
roidelapluie: Expose the good function directly instead of returning it

sudoapt-getclean: Hey thanks for the suggestion the result is shorter indeed.
I did it like this because I imitated how it was done (that is obviously no excuse ;-) ) I took the liberty to correct those occasions also so that this anti-pattern will not creep in again.

PR #176, opened 2019-09-17
Support for Enabling pre-existing deploy key on a project

### LINKED ISSUE #64: [Feature Request] Enable Privately accessible deploy keys
### Terraform Version
Terraform v0.11.8

### Affected Resource(s)
gitlab_*

### [Feature Request] 
We are currently integrating GitLab with azure, and configuring the deploykeys and webhooks, have been working great up until we started using submodules.

It would be great to be able to enable previously added deploy keys with multiple projects.
Now, we have to go into the submodule, and under "privately accessible deploy keys" we  have to enable the deploy key for the azure resource for that project manual.

primeroz:
fixes #64 

Allow to enable already existing deploy key on a project by ID 

```hcl
variable keys_to_enable {
  type = list(string)
  default = [ "678", "901" ]
}

variable project {
  type = string
  default = "12345"
}

resource "gitlab_deploy_key_enable" "keys" {
  count   = length(var.keys_to_enable)
  project = var.project
  key_id  = element(var.keys_to_enable, count.index)
}
```


primeroz:
I did not realize the `CI` would run on a local dockerized gitlab ( which make sense ) so the key i used in the test is not availble.

will change to pre-create a key just for the test sake

gitlab/resource_gitlab_deploy_key_enable.go
```diff
@@ -0,0 +1,127 @@
...
+	project := d.Get("project").(string)
+	deployKeyID, err := strconv.Atoi(d.Get("key_id").(string))
+	if err != nil {
+		return err
+	}
+	log.Printf("[DEBUG] read gitlab deploy key %s/%d", project, deployKeyID)
+
+	deployKey, response, err := client.DeployKeys.GetDeployKey(project, deployKeyID)
+
+	// HTTP 404 indicates that the key is not enabled on the project, maybe it was removed through the UI
```
roidelapluie: We do not ignore 404 anymore. Please remove this.

gitlab/resource_gitlab_deploy_key_enable.go
```diff
@@ -0,0 +1,127 @@
...
+	deployKeyID, err := strconv.Atoi(d.Get("key_id").(string))
+	if err != nil {
+		return err
+	}
+	log.Printf("[DEBUG] Delete gitlab deploy key %s/%d", project, deployKeyID)
+
+	response, err := client.DeployKeys.DeleteDeployKey(project, deployKeyID)
+
+	// HTTP 204 is success with no body
+	if response.StatusCode == 204 {
```
roidelapluie: we should just do response.StatusCode/100 == 2

website/docs/r/deploy_key_enable.html.markdown
```diff
@@ -0,0 +1,38 @@
...
+
+This resource allows you to enable pre-existing deploy keys for your GitLab projects.
+
+**the GITLAB KEY_ID for the deploy key must be known**
+
+## Example Usage
+
+```hcl
+resource "gitlab_deploy_key_enable" "example" {
+  project = "12345"
```
roidelapluie: please provide an example where we use variables. (do not hesitate to add a small gitlab_project to the example)

gitlab/resource_gitlab_deploy_key_enable.go
primeroz: I think in this case we actually need it , see the following example workflow 

* Key is enabled  on project X through terraform , resource is not in the state so is not read at first and just created
* Resource is now in the state
* Human being disable the key by hand through the ui
* Terraform is rerun and get 404 back from gitlab

In this case unless i can control this response , and set the resource Id to "" , terraform will just error out instead of re-enabling the key 

This is my first ever terraform resource so i might just be wrong :) 

Am i missing something ?

gitlab/resource_gitlab_deploy_key_enable.go
primeroz: will do

website/docs/r/deploy_key_enable.html.markdown
primeroz: will do

primeroz:
@roidelapluie pushed some changes and added a comment. 

I would have expected Ci to trigger again but can't see it happening

primeroz:
@roidelapluie do you have any update on this ? thanks a lot

gitlab/resource_gitlab_deploy_key_enable.go
roidelapluie: That is the point. If someones removes it from the UI, terraform should error, then it must me removed from state (terraform state rm).

Things should not disappear from the state magically

primeroz: Ok, quite an _opinionated_ point but i actually like it, agree with it and probably will reuse it myself.

Will fix now

primeroz:
@roidelapluie all changes done.

thanks !

primeroz:
@roidelapluie let me know if there is anything more you want me to change on this, i just don't want to miss the next release if is good since i heavily rely on this feature and i am currently running a local build of the provider which is not ideal.

thanks!

roidelapluie:
thanks

PR #199, opened 2019-10-23
Adds project approval rules resource

ThatGerber:


ThatGerber:
Closing this PR until I can determine a fix. 

Currently, the Project Approval Rules API will return inconsistent responses under certain conditions. 

If a User ID and a Group ID is provided in the request, GitLab will coalesce all of the user ids from all of the groups and the individual users in the "eligible_approvers" array, but, return an empty "users" array. 

Because the users array is empty, you would need to generate a list of user ids from all of the groups and filter them from eligible approvers list to get the remaining list of user IDs.

I have submitted a ticket to GitLab about it. I may attempt the ugly fix in the mean time, since it's unknown how long it could be before they fix it, but I will open a new MR to do that.

PR #208, opened 2019-11-04
Adds project approval rules resource

ThatGerber:
This MR adds a Project Approval Rule resource. 

The test matrix was also updated. "GitLab CE" == "GitLab EE without a license". This updates the test matrix to allow the CE tests to fail and require EE to pass. If the EE tests are passing, then the CE tests should pass, as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,4 +30,5 @@ website/vendor
 !command/test-fixtures/**/*.tfstate
 !command/test-fixtures/**/.terraform/
 
-license/JulienPivotto.gitlab-license
+.license/JulienPivotto.gitlab-license
+.license/*.gitlab-license
--- a/.gitignore
+++ b/.gitignore
@@ -30,5 +30,4 @@ website/vendor
 !command/test-fixtures/**/*.tfstate
 !command/test-fixtures/**/.terraform/
 
-.license/JulienPivotto.gitlab-license
-.license/*.gitlab-license
+.license/
--- a/gitlab/helper_test.go
+++ b/gitlab/helper_test.go
@@ -26,13 +26,13 @@ func testAccCompareGitLabAttribute(attr string, expected, received *schema.Resou
 }
 
 func testAccIsSkippedAttribute(needle string, haystack []string) bool {
-	if needle == "" {
-		return false
-	}
-	for _, hay := range haystack {
-		if hay == needle {
-			return true
+	if needle != "" {
+		for _, hay := range haystack {
+			if hay == needle {
+				return true
+			}
 		}
 	}
+
 	return false
 }
--- a/gitlab/resource_gitlab_project_approval_rule.go
+++ b/gitlab/resource_gitlab_project_approval_rule.go
@@ -324,7 +324,7 @@ func getApproversNotInProject(client *gitlab.ProjectMembersService, userIDs []in
 
 	if len(members) != 0 {
 		for _, member := range members {
-			if member.AccessLevel >= 30 {
+			if member.AccessLevel >= gitlab.DeveloperPermissions {
 				memberIDs.Add(member.ID)
 			}
 		}
--- a/gitlab/resource_gitlab_project_approval_rule.go
+++ b/gitlab/resource_gitlab_project_approval_rule.go
@@ -1,0 +1,425 @@
+package gitlab
+
+import (
+	"fmt"
+	"log"
+	"strconv"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	gitlab "github.com/xanzy/go-gitlab"
+)
+
+// https://docs.gitlab.com/ee/api/merge_request_approvals.html#create-project-level-rule
+
+var resourceGitLabProjectApprovalRuleName = "gitlab_project_approval_rule"
+
+var resourceGitLabProjectApprovalRuleSchema = map[string]*schema.Schema{
+	"project": {
+		Type:     schema.TypeString,
+		ForceNew: true,
+		Required: true,
+	},
+	"name": {
+		Type:     schema.TypeString,
+		ForceNew: true,
+		Required: true,
+	},
+	"approvals_required": {
+		Type:     schema.TypeInt,
+		Required: true,
+	},
+	"user_ids": {
+		Type:     schema.TypeSet,
+		Optional: true,
+		Elem:     &schema.Schema{Type: schema.TypeInt},
+		Set:      schema.HashInt,
+	},
+	"group_ids": {
+		Type:     schema.TypeSet,
+		Optional: true,
+		Elem:     &schema.Schema{Type: schema.TypeInt},
+		Set:      schema.HashInt,
+	},
+	"eligible_approvers": {
+		Type:     schema.TypeSet,
+		Computed: true,
+		Elem:     &schema.Schema{Type: schema.TypeInt},
+		Set:      schema.HashInt,
+	},
+}
+
+func resourceGitlabProjectApprovalRule() *schema.Resource {
+	return &schema.Resource{
+		Schema: resourceGitLabProjectApprovalRuleSchema,
+		Create: resourceGitlabProjectApprovalRuleCreate,
+		Read:   resourceGitlabProjectApprovalRuleRead,
+		Update: resourceGitlabProjectApprovalRuleUpdate,
+		Delete: resourceGitlabProjectApprovalRuleDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+	}
+}
+
+func resourceGitlabProjectApprovalRuleCreate(d *schema.ResourceData, meta interface{}) error {
+	var err error
+
+	options := gitlab.CreateProjectLevelRuleOptions{
+		Name:              gitlab.String(d.Get("name").(string)),
+		ApprovalsRequired: gitlab.Int(d.Get("approvals_required").(int)),
+		UserIDs:           expandApproverIds(d.GetOk("user_ids")),
+		GroupIDs:          expandApproverIds(d.GetOk("group_ids")),
+	}
+	project := d.Get("project").(string)
+
+	log.Printf("[DEBUG] Project %s create gitlab project-level rule %+v", project, options)
+
+	client := meta.(*gitlab.Client)
+	if rule, _, err := client.Projects.CreateProjectApprovalRule(project, &options); err == nil {
+		d.SetId(createApprovalRuleID(rule.ID, project))
+
+		return resourceGitlabProjectApprovalRuleRead(d, meta)
+	}
+
+	return err
+}
+
+func resourceGitlabProjectApprovalRuleRead(d *schema.ResourceData, meta interface{}) error {
+	var err error
+
+	log.Printf("[DEBUG] read gitlab project-level rule %s", d.Id())
+
+	d.Partial(true)
+
+	projectID, _, err := splitApprovalRuleID(d.Id())
+	if err != nil {
+		return err
+	}
+	d.Set("project", projectID)
+	d.SetPartial("project")
+
+	rule, err := getApprovalRuleByID(meta, d.Id())
+	if err != nil || rule == nil {
+		if err != nil {
+			return err
+		}
+		return fmt.Errorf("unable to read GitLab approvel rule %s", d.Id())
+	}
+
+	d.Set("name", rule.Name)
+	d.SetPartial("name")
+	d.Set("approvals_required", rule.ApprovalsRequired)
+	d.SetPartial("approvals_required")
+
+	groupIDs := flattenApprovalRuleGroupIDs(rule.Groups)
+	if err := d.Set("group_ids", groupIDs); err != nil {
+		return err
+	}
+	d.SetPartial("group_ids")
+
+	eligibleApprovers := flattenApprovalRuleUserIDs(rule.EligibleApprovers)
+	if err := d.Set("eligible_approvers", eligibleApprovers); err != nil {
+		return err
+	}
+	d.SetPartial("eligible_approvers")
+
+	userIDs, err := getProjectApprovalRuleUserIDs(meta.(*gitlab.Client), d.Id())
+	if err != nil {
+		return err
+	}
+
+	if err := d.Set("user_ids", userIDs); err != nil {
+		return err
+	}
+
+	d.Partial(false)
+
+	return nil
+}
+
+func resourceGitlabProjectApprovalRuleUpdate(d *schema.ResourceData, meta interface{}) error {
+	var err error
+
+	if project, ruleID, err := splitApprovalRuleID(d.Id()); err == nil {
+		options := gitlab.UpdateProjectLevelRuleOptions{
+			Name:              gitlab.String(d.Get("name").(string)),
+			ApprovalsRequired: gitlab.Int(d.Get("approvals_required").(int)),
+			UserIDs:           expandApproverIds(d.GetOk("user_ids")),
+			GroupIDs:          expandApproverIds(d.GetOk("group_ids")),
+		}
+
+		ruleIDInt, err := strconv.Atoi(ruleID)
+		if err != nil {
+			return err
+		}
+
+		log.Printf("[DEBUG] Project %s update gitlab project-level approval rule %s", project, *options.Name)
+
+		client := meta.(*gitlab.Client)
+		if _, _, err = client.Projects.UpdateProjectApprovalRule(project, ruleIDInt, &options); err == nil {
+			return resourceGitlabProjectVariableRead(d, meta)
+		}
+	}
+
+	return err
+}
+
+func resourceGitlabProjectApprovalRuleDelete(d *schema.ResourceData, meta interface{}) error {
+	var err error
+
+	if project, ruleID, err := splitApprovalRuleID(d.Id()); err == nil {
+		ruleIDInt, err := strconv.Atoi(ruleID)
+		if err != nil {
+			return err
+		}
+
+		log.Printf("[DEBUG] Project %s delete gitlab project-level approval rule %d", project, ruleIDInt)
+
+		client := meta.(*gitlab.Client)
+		_, err = client.Projects.DeleteProjectApprovalRule(project, ruleIDInt)
+	}
+
+	return err
+}
+
+// getApprovalRuleByID checks the list of rules and finds the one that matches our rule ID.
+func getApprovalRuleByID(meta interface{}, id string) (*gitlab.ProjectApprovalRule, error) {
+	var rules []*gitlab.ProjectApprovalRule
+	var err error
+
+	projectID, ruleIDStr, err := splitApprovalRuleID(id)
+	if err != nil {
+		return nil, err
+	}
+
+	client := meta.(*gitlab.Client)
+	if rules, _, err = client.Projects.GetProjectApprovalRules(projectID); err == nil {
+		var ruleIDInt int
+
+		if ruleIDInt, err = strconv.Atoi(ruleIDStr); err != nil {
+			return nil, err
+		}
+
+		for _, r := range rules {
+			if r.ID == ruleIDInt {
+				log.Printf("[DEBUG] found project-level rule %+v", r)
+				return r, nil
+			}
+		}
+	}
+
+	return nil, err
+}
+
+// createApprovalRuleID creates an ID of two parts: projectID:ruleID
+func createApprovalRuleID(ruleID int, projectID string) string {
+	return buildTwoPartID(gitlab.String(projectID), gitlab.String(strconv.Itoa(ruleID)))
+}
+
+// splitApprovalRuleID splits an approval rule into two parts, projectID:ruleID.
+// An error is returned if one occurs.
+func splitApprovalRuleID(ruleID string) (string, string, error) {
+	return parseTwoPartID(ruleID)
+}
+
+// flattenApprovalRuleUserIDs flattens a list of approval user ids into a list
+// of ints for storage in state.
+func flattenApprovalRuleUserIDs(users []*gitlab.BasicUser) []interface{} {
+	m := []interface{}{}
+	for _, user := range users {
+		m = append(m, user.ID)
+	}
+
+	return m
+}
+
+// flattenApprovalRuleGroupIDs flattens a list of approval group ids into a list
+// of ints for storage in state.
+func flattenApprovalRuleGroupIDs(groups []*gitlab.Group) []interface{} {
+	m := []interface{}{}
+	for _, group := range groups {
+		m = append(m, group.ID)
+	}
+
+	return m
+}
+
+// expandApproverIds Expands an interface into a list of ints to read from state.
+func expandApproverIds(ids interface{}, hasItems bool) []int {
+	m := []int{}
+	if hasItems {
+		for _, id := range ids.(*schema.Set).List() {
+			m = append(m, id.(int))
+		}
+	}
+
+	return m
+}
+
+// getProjectApprovalRuleUserIDs generates a list of approvers for a rule that
+// could not be found in the project or in any attached groups and returns a
+// list containing only the user IDs not found as developer <= higher in the
+// project, or developer <= in any of the rule groups.
+//
+// This method attempts to work around a bug in GitLab where there are instances
+// where the `users` attribute in the response will be empty under certain
+// conditions.
+func getProjectApprovalRuleUserIDs(client *gitlab.Client, ruleID string) ([]int, error) {
+	userIDs := newMemberList([]int{})
+
+	rule, err := getApprovalRuleByID(client, ruleID)
+	if err != nil {
+		return userIDs.List(), err
+	}
+	users := []int{}
+
+	if len(rule.Users) > 0 {
+		for _, i := range flattenApprovalRuleUserIDs(rule.Users) {
+			users = append(users, i.(int))
+		}
+
+		return users, nil
+	}
+
+	for _, i := range flattenApprovalRuleUserIDs(rule.EligibleApprovers) {
+		users = append(users, i.(int))
+	}
+
+	projectID, _, err := splitApprovalRuleID(ruleID)
+	if err != nil {
+		return users, err
+	}
+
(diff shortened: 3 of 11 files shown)


ThatGerber:
I'm running into a "Works on my computer" situation with this. My tests are passing locally with the same envs as Travis; only difference being Travis using Go 1.12.x instead of Go 1.13.x.

Go env:
```
GO111MODULE="on"
GOARCH="amd64"
GOBIN="/Users/chris.gerber/bin"
GOCACHE=".../Library/Caches/go-build"
GOENV=".../Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH=""
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="...go/1.13.4/libexec"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="...go/1.13.4/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="src/github.com/terraform-providers/terraform-provider-gitlab/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/lq/12rnd7453_v9gpkqxbd9fjn40000gq/T/go-build469647254=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

Go Version

```
go version go1.13.4 darwin/amd64
```

ThatGerber:
@roidelapluie I don't believe the `encrypted_4a4b21de8599_key` and `encrypted_4a4b21de8599_iv` variables are being set correctly. It looks like those two steps of `scripts/license.sh` are never being run, which is causing the EE tests to fail.

roidelapluie:
EE will always fail from Pull Requests.

It will only run when coming from the original repo. That is why EE is set as allowed failure

roidelapluie:
Otherwise you could open a PR and just cat the license file :)

learoyklinginsmith:
@ThatGerber Do you plan to come back to this after the new year?

ThatGerber:
It looks like I need to rebase and fix this. 

The problem though is the pipeline will never pass for two reasons: 

1. When verifying CE, it's running tests for features that are only available in EE
2. When verifying EE, it will fail because there isn't a license key available.

GitLab has stated that EE and CE are the same thing; "CE" is just GitLab Core, and is functionally the same as running GitLab EE without a license key. Is there any reason to keep testing against CE? Can we just test EE and expect that if EE passes that CE will be valid?

roidelapluie:
Please use https://github.com/terraform-providers/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_project_push_rules_test.go#L47

We have a license for the tests.

roidelapluie:
We can not test just EE as it would expose out EE key to anyone. And CE is most used than EE in the wild.

roidelapluie:
Please use skipFunc, I consider quite important that the tests pass on CE, even if EE tests are skipped

ThatGerber:
Updated to skip EE-specific tests if `GITLAB_LICENSE_FILE` is not set.

ThatGerber:
@roidelapluie Is there anything you need from me to get this merged? Thanks

.gitignore
```diff
@@ -30,4 +30,5 @@ website/vendor
 !command/test-fixtures/**/*.tfstate
 !command/test-fixtures/**/.terraform/
 
-license/JulienPivotto.gitlab-license
+.license/JulienPivotto.gitlab-license
```
roidelapluie: nip: we could ignore .license/*

gitlab/helper_test.go
```diff
@@ -26,13 +26,13 @@ func testAccCompareGitLabAttribute(attr string, expected, received *schema.Resou
 }
 
 func testAccIsSkippedAttribute(needle string, haystack []string) bool {
-	if needle == "" {
-		return false
-	}
-	for _, hay := range haystack {
-		if hay == needle {
-			return true
+	if needle != "" {
```
roidelapluie: Why did you change that? I think the less the code is indented, the best it is

scripts/start-gitlab.sh
```diff
@@ -1,31 +1,42 @@
...
-    img=gitlab/gitlab-ce
+test -z "$(docker ps -f 'name=gitlab' -q)" || { echo "Gitlab already running"; exit 0; }
+
+_gitlab_image="gitlab/gitlab-ce"
+extra=""
+if test -n "${GITLAB_LICENSE_FILE}"; then
+  _license_dir=${GITLAB_LICENSE_DIR:-.license}
+  _license_file=${GITLAB_LICENSE_FILE:-JulienPivotto.gitlab-license}
+
+  if ! test -f "${_license_dir}/${_license_file}"
```
roidelapluie: test -r

scripts/start-gitlab.sh
```diff
@@ -1,31 +1,42 @@
...
+  then
+      echo "No license"
```
roidelapluie:
```suggestion
      echo "No license at ${_license_dir}/${_license_file}"
```

scripts/start-gitlab.sh
```diff
@@ -1,31 +1,42 @@
...
+)
 
 echo -n "Waiting for gitlab to be ready "
 i=1
 until wget -t 1 127.0.0.1:8080 -O /dev/null -q
 do
     sleep 1
     echo -n .
-    if [[ $((i%3)) == 0 ]]; then echo -n ' '; fi
+    if test $((i%3)) -eq 0; then echo -n ' '; fi
```
roidelapluie: Why not using the bash internal? we enforce bash on #!

website/docs/r/project_approval_rule.html.markdown
```diff
@@ -0,0 +1,55 @@
...
+This resource allows you to create and manage approval rules for your GitLab 
+projects. For further information on approval rules, consult the [gitlab
+documentation](https://docs.gitlab.com/ee/api/merge_request_approvals.html#project-level-mr-approvals).
+
+*Note*: This feature requires a GitLab Starter account or above.
+
+## Example Usage
+
+```hcl
+resource "gitlab_project_approval_rule" "example" {
```
roidelapluie: terraform fmt

gitlab/resource_gitlab_project_approval_rule.go
```diff
@@ -0,0 +1,425 @@
...
+		Type:     schema.TypeSet,
+		Optional: true,
+		Elem:     &schema.Schema{Type: schema.TypeInt},
+		Set:      schema.HashInt,
+	},
+	"group_ids": {
+		Type:     schema.TypeSet,
+		Optional: true,
+		Elem:     &schema.Schema{Type: schema.TypeInt},
+		Set:      schema.HashInt,
```
roidelapluie: should we set ContflictsWith ? or you can set both?

website/docs/r/project_approval_rule.html.markdown
```diff
@@ -0,0 +1,55 @@
...
+
+The following arguments are supported:
+
+* `project` - (Required, string) The name or id of the project to add the approval rules.
+
+* `name` - (Required) The name of the approval rule.
+
+* `approvals_required` - (Required) The number of approvals required for this rule.
+
+* `user_ids` - (Required)  A list of specific User IDs to add to the list of approvers.
```
roidelapluie: It is set to Optional: true

gitlab/resource_gitlab_project_approval_rule.go
```diff
@@ -0,0 +1,425 @@
...
+
+	return err
+}
+
+func resourceGitlabProjectApprovalRuleRead(d *schema.ResourceData, meta interface{}) error {
+	var err error
+
+	log.Printf("[DEBUG] read gitlab project-level rule %s", d.Id())
+
+	d.Partial(true)
```
roidelapluie: Is this really needed?

gitlab/resource_gitlab_project_approval_rule.go
```diff
@@ -0,0 +1,425 @@
...
+}
+
+// getProjectApprovalRuleUserIDs generates a list of approvers for a rule that
+// could not be found in the project or in any attached groups and returns a
+// list containing only the user IDs not found as developer <= higher in the
+// project, or developer <= in any of the rule groups.
+//
+// This method attempts to work around a bug in GitLab where there are instances
+// where the `users` attribute in the response will be empty under certain
+// conditions.
```
roidelapluie: do you have an upstream bug url reference?

roidelapluie: I would add some [DEBUG] log in this method to know that we are workign around a gitlab bug

gitlab/resource_gitlab_project_approval_rule.go
```diff
@@ -0,0 +1,425 @@
...
+
+	memberIDs := newMemberList([]int{})
+	var members []*gitlab.ProjectMember
+	if members, _, err = client.ListProjectMembers(projectID, nil, nil); err != nil {
+		return memberIDs.List(), err
+	}
+
+	if len(members) != 0 {
+		for _, member := range members {
+			if member.AccessLevel >= 30 {
```
roidelapluie: instead of 30, isn't there a const ?

website/docs/r/project_approval_rule.html.markdown
```diff
@@ -0,0 +1,55 @@
...
+
+## Attributes Reference
+
+The resource exports the following additional attributes:
+
+* `eligible_approvers` - A list of approvers from the `user_ids` parameter, groups, and the project that are included in the approval rule.
+
+## Import
+
+GitLab project approval rules imported using an id made up of `project-id:rule-id`, e.g.
```
roidelapluie: can be imported / are imported

gitlab/helper_test.go
ThatGerber: I felt it's clearer with fewer exit conditions for the method. The method exits true if the conditions are met, false if they are not. I can revert this, but I felt it was clearer with 1 potential `true` exit path vs. multiple.

scripts/start-gitlab.sh
ThatGerber: Good call. I will add it.

scripts/start-gitlab.sh
ThatGerber: Both are considered built-in commands. I switched to `test` to make it consistent with the rest of the if statements. The two constructs are evaluated identically by Bash: https://www.gnu.org/software/bash/manual/bash.html#Bash-Conditional-Expressions

gitlab/resource_gitlab_project_approval_rule.go
ThatGerber: Because of the bug described below, Terraform may potentially run into a failure condition while updating the resource. Using partial tells Terraform that everything that's actually been updated so far in the run should be added to the stored state and everything else should not be discarded.

gitlab/resource_gitlab_project_approval_rule.go
ThatGerber: You can set both or none. https://docs.gitlab.com/ee/api/merge_request_approvals.html#create-project-level-rule

gitlab/resource_gitlab_project_approval_rule.go
ThatGerber: I'll add the ticket number as a comment above the method. The ticket has been opened, but I have not gotten any response on it. https://gitlab.com/gitlab-org/gitlab/issues/35008

scripts/start-gitlab.sh
roidelapluie: TIL :) thanks

gitlab/helper_test.go
roidelapluie: ok

roidelapluie:
One last remark

Why didn't you use isRunningInEE ?

this looks better than looking for an env variable as people could run it against other EE instances without setting our env variable.

roidelapluie:
And use SkipFunc: isRunningInCE

ThatGerber:
I can update to use that `isRunningInCE`. The main reason I didn't use it was that I didn't know it existed :)

The call that `isRunningInEE()` makes to determine version, `conn.Version.GetVersion()`, requires the user be authenticated with GitLab to be successful. Checking for the ENV does not require successful authentication. If you don't mind that the test could fail because of a bad access token, then I can switch and use `isRunningInCE()` to check.

roidelapluie:
I am updating it right now

gdubicki:
Hi!

I have read the code in this MR and - if I got it right - you do not test setting more than 1 user or more than 1 group as approvers in a rule, either new one or changing an existing one.

If you would do that then I think you would run into the same GitLab bug as I did in my WIP PR that adds the same API support to the GitLabForm app - https://github.com/egnyte/gitlabform/pull/80 .

Please consider adding such tests to verify if we do in fact have such bug.

Note that if bug exists then this feature of this provider will only work in trivial cases of setting 1 user and/or 1 group as approvers.

roidelapluie:
@ThatGerber have you seen that comment?

roidelapluie:
ping @ThatGerber

gdubicki:
@ThatGerber & @roidelapluie : I am sorry but it turned out that my issues with this API in GitLabForm were caused by using URL encoded form & content type set to `application/x-www-form-urlencoded` in requests to GitLab API instead of JSON & `application/json`.

Using URL encoding instead of JSON makes _almost_ all requests to work, but fail in _some_ cases, which caused more or less obscure bugs in my app (see https://github.com/egnyte/gitlabform/issues/90 for more info).

If this provider uses JSON and `application/json` content type then you should be good to go.

gitlab/resource_gitlab_project_approval_rule.go
```diff
@@ -0,0 +1,429 @@
...
+	if users, err := getApproversNotInProject(client.ProjectMembers, users, projectID); err != nil {
+		return users, err
+	}
+
+	groups := []int{}
+	for _, i := range flattenApprovalRuleGroupIDs(rule.Groups) {
+		groups = append(groups, i.(int))
+	}
+
+	if users, err := getApproversNotInGroups(client.Groups, users, groups); err != nil {
```
matt-demers, line 306: I tried out this PR myself, and ran into one issue. I had multiple groups each with multiple members, which applied cleanly, but then the terraform plan would always be dirty saying it was removing a user from the user_ids list. I dug into it and the error was that the `users` variable here was not getting set to the result of this function call. 

I think because `users` and `err` are already defined, you need to just use an `=` instead of `:=`. And the same thing on line 297 will probably fail for the same reason. That change fixed the plan for me.

baurmatt:
Hey, this looks great and I'm looking forward to seeing this being merged/released! 🎉  What's the current blocker? :)

gitlab/resource_gitlab_project_approval_rule.go
```diff
@@ -0,0 +1,429 @@
...
+	}
+	d.Set("project", projectID)
+	d.SetPartial("project")
+
+	rule, err := getApprovalRuleByID(meta, d.Id())
+	if err != nil || rule == nil {
+		if err != nil {
+			return err
+		}
+		return fmt.Errorf("unable to read GitLab approvel rule %s", d.Id())
```
aquavitae, line 106: Typo here: `approval`

gitlab/resource_gitlab_project_approval_rule.go
```diff
@@ -0,0 +1,429 @@
...
+		ruleIDInt, err := strconv.Atoi(ruleID)
+		if err != nil {
+			return err
+		}
+
+		log.Printf("[DEBUG] Project %s update gitlab project-level approval rule %s", project, *options.Name)
+
+		client := meta.(*gitlab.Client)
+		if _, _, err = client.Projects.UpdateProjectApprovalRule(project, ruleIDInt, &options); err == nil {
+			return resourceGitlabProjectVariableRead(d, meta)
```
aquavitae, line 160: `resourceGitlabProjectVariableRead`? Shouldn't it be `resourceGitlabProjectApprovalRuleRead`?

gitlab/resource_gitlab_project_approval_rule.go
```diff
@@ -0,0 +1,429 @@
...
+
+		client := meta.(*gitlab.Client)
+		_, err = client.Projects.DeleteProjectApprovalRule(project, ruleIDInt)
+	}
+
+	return err
+}
+
+// getApprovalRuleByID checks the list of rules and finds the one that matches our rule ID.
+func getApprovalRuleByID(meta interface{}, id string) (*gitlab.ProjectApprovalRule, error) {
```
armsnyder, line 186: Can you change the first argument to `client *gitlab.Client`?

gitlab/resource_gitlab_project_approval_rule.go
```diff
@@ -0,0 +1,429 @@
...
+	projectID, ruleIDStr, err := splitApprovalRuleID(id)
+	if err != nil {
+		return nil, err
+	}
+
+	client := meta.(*gitlab.Client)
+	if rules, _, err = client.Projects.GetProjectApprovalRules(projectID); err == nil {
+		var ruleIDInt int
+
+		if ruleIDInt, err = strconv.Atoi(ruleIDStr); err != nil {
```
armsnyder, line 199: You could move this error check to before the API call, when you parse the rule ID initially, to short circuit before making an API call.

gitlab/resource_gitlab_project_approval_rule.go
```diff
@@ -0,0 +1,429 @@
...
+			return nil, err
+		}
+
+		for _, r := range rules {
+			if r.ID == ruleIDInt {
+				log.Printf("[DEBUG] found project-level rule %+v", r)
+				return r, nil
+			}
+		}
+	}
```
armsnyder, line 209: Uh oh. If the rule ID does not exist, then the for loop is going to finish without returning, which means you will return a `nil` rule and a `nil` error. Looking at the places where you call this function, I don't see you handling the `nil` rule case, so this is a bug.

gitlab/resource_gitlab_project_approval_rule.go
armsnyder: ^ Yes please fix, if you need to even make this additional API call here. Though, you should already have all the data you need in the return value from your call to `UpdateProjectApprovalRule`, so I would prefer if you use that return object than to make another API call, to reduce the number of calls and to mitigate potential data consistency problems.

gitlab/resource_gitlab_project_approval_rule.go
```diff
@@ -0,0 +1,429 @@
...
+// where the `users` attribute in the response will be empty under certain
+// conditions.
+//
+// Reference: https://gitlab.com/gitlab-org/gitlab/issues/35008
+func getProjectApprovalRuleUserIDs(client *gitlab.Client, ruleID string) ([]int, error) {
+	log.Print("[DEBUG] Looking up project approval rule User IDs as a workaround for bug in GitLab.")
+
+	userIDs := newMemberList([]int{})
+
+	rule, err := getApprovalRuleByID(client, ruleID)
```
armsnyder, line 274: You shouldn't need to make this API call here, since you have already made this API call previously in the `resourceGitlabProjectApprovalRuleRead` function.

gitlab/resource_gitlab_project_approval_rule.go
```diff
@@ -0,0 +1,429 @@
...
+// could not be found in the project or in any attached groups and returns a
+// list containing only the user IDs not found as developer <= higher in the
+// project, or developer <= in any of the rule groups.
+//
+// This method attempts to work around a bug in GitLab where there are instances
+// where the `users` attribute in the response will be empty under certain
+// conditions.
+//
+// Reference: https://gitlab.com/gitlab-org/gitlab/issues/35008
+func getProjectApprovalRuleUserIDs(client *gitlab.Client, ruleID string) ([]int, error) {
```
armsnyder, line 269: I have to wonder if it's really our responsibility to work around this bug, since doing so adds such complexity and potential for strange behavior, like Matt has pointed out below. How often does it occur? What is the impact to this provider when it does? It seems to me like the impact is that it would show a diff when it shouldn't, but that's not a serious problem because making an update call with the same list of user IDs is safe.

gitlab/resource_gitlab_project_approval_rule_test.go
```diff
@@ -0,0 +1,416 @@
...
+			return resource.TestCheckResourceAttr(name, attribute, expectedValue)(state)
+		})
+	}
+
+	return resource.ComposeAggregateTestCheckFunc(checks...)
+}
+
+func (c *checkGitlabProjectApprovalRule) EqualLists(expected, received interface{}) error {
+	if expected.(*schema.Set).Len() != received.(*schema.Set).Len() {
+		return fmt.Errorf("Expected set count (%d) but received (%d)", expected.(*schema.Set).Len(), received.(*schema.Set).Len())
```
armsnyder, line 73: Minor issue but errors should not be capitalized. See also `Exists` line 85.

gitlab/resource_gitlab_project_approval_rule_test.go
```diff
@@ -0,0 +1,416 @@
...
+
+var gitlabProjectApprovalRuleSkipAttributes = []string{
+	"project", "user_ids", "group_ids", "eligible_approvers",
+}
+
+type checkGitlabProjectApprovalRule struct {
+	SkippedAttributes []string
+}
+
+func (c *checkGitlabProjectApprovalRule) Aggregate(name string, expected gitlab.ProjectApprovalRule) resource.TestCheckFunc {
```
armsnyder, line 24: These methods do not need to be exported (public). Also I see some more things that are exported and do not need to be -- the `SkippedAttributes` field, your methods on `memberList`, and fields of `gitlabProjectApprovalRuleFixtures`.

gitlab/resource_gitlab_project_approval_rule_test.go
```diff
@@ -0,0 +1,416 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	gitlab "github.com/xanzy/go-gitlab"
+)
+
+var gitlabProjectApprovalRuleSkipAttributes = []string{
+	"project", "user_ids", "group_ids", "eligible_approvers",
+}
+
+type checkGitlabProjectApprovalRule struct {
+	SkippedAttributes []string
```
armsnyder, line 21: What is the purpose of this field if it is always set to `gitlabProjectApprovalRuleSkipAttributes`?

gitlab/resource_gitlab_project_approval_rule_test.go
```diff
@@ -0,0 +1,416 @@
...
+func getValueFromGitLabApprovalRule(rule *gitlab.ProjectApprovalRule, key string) string {
+	var ret interface{}
+
+	switch key {
+	case "name":
+		ret = rule.Name
+	case "approvals_required":
+		ret = rule.ApprovalsRequired
+	case "group_ids":
+		ret = flattenApprovalRuleGroupIDs(rule.Groups)
```
armsnyder, line 169: In general can you avoid calling the same helper functions in your test as you do in your main code? You run the risk of testing for the same buggy behavior that you may have introduced. If there are bugs in these functions you will not catch them.

gitlab/resource_gitlab_project_approval_rule.go
armsnyder: Never mind. I see it is a recommended practice in the docs. Carry on!

gitlab/resource_gitlab_project_approval_rule.go
```diff
@@ -0,0 +1,429 @@
...
+	}
+	project := d.Get("project").(string)
+
+	log.Printf("[DEBUG] Project %s create gitlab project-level rule %+v", project, options)
+
+	client := meta.(*gitlab.Client)
+	if rule, _, err := client.Projects.CreateProjectApprovalRule(project, &options); err == nil {
+		d.SetId(createApprovalRuleID(rule.ID, project))
+
+		return resourceGitlabProjectApprovalRuleRead(d, meta)
```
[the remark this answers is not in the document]
armsnyder: Never mind. I see it is a recommended practice in the docs. Carry on!

armsnyder:
Hey @ThatGerber are you still working on this PR? It has some new comments and also needs rebasing.

Stromweld:
would love it if this could make it into 3.0 or 3.x soon #411

mgriffin13:
This is a pretty glaring feature omission by the provider. Would it be possible to get an update on when it might get merged and released?

ThatGerber:
@armsnyder @Stromweld @mgriffin13 and everyone else

I work as a consultant and the time and materials I spent contributing to this provider was graciously billable to my amazing client. Since I started this PR, my contract with that client has ended and I'm not currently in a role where I am using GitLab nor do not have the personal time to invest in contributing to the GitLab provider. I do not work for GitLab. I have talked with GitLab's salespeople and some devs encouraging them to take over the development of this provider instead of leaving it entirely to the OSS community, but I haven't gotten any traction, it would appear.

GitLab also lacks the mechanism of requesting license keys except directly through free trials or their sales channels, and I do not have the interest in making connections with yet another sales team in order to get the license key to verify the work.

The code is there for anyone to rebase and fix. If someone wants to provide me a license key on the _possibility_, no guarantee, I may have personal time to work on this, my email address is available in my GitHub profile. I cannot promise I will have the time or energy to work on this, but if by some miracle of fate the stars align and I am granted a license key and 8 hours of free time to setup local development and complete this PR, I will try to do it.

nicholasklick:
@ThatGerber Thanks for the backgrounder and for your contributions to this project.

As GitLab [has taken over this project](https://github.com/gitlabhq/terraform-provider-gitlab/issues/376) we can certainly help push the PR forward.

Thanks!

btyy77c:
FYI, I've picked up work on approval rules. The MR can be found: https://github.com/gitlabhq/terraform-provider-gitlab/pull/250

PR #216, opened 2019-11-17
Use Github Actions Workflows

roidelapluie:


roidelapluie:
Should we remove travis?

ringods:
@roidelapluie not yet. The acceptance tests, even for CE, are not correctly set up yet in Github Actions. Can you explain me what this line exactly does?

https://github.com/terraform-providers/terraform-provider-gitlab/blob/dda78c0a5043b7fb135d68c3119ceea99b6f0c3b/scripts/start-gitlab.sh#L39

It is the only line I see related to `GITLAB_TOKEN` but it doesn't work in Github Actions it seems.

ringods:
@roidelapluie Travis may be removed now that #285 is merged.

PR #218, opened 2019-11-18
Add group runners token to the group data source/resource

### LINKED ISSUE #181: Feature Request: Group Runner Token
### Use Case

The provider already exposes project-wide runner tokens, which has limited usefulness on an organization level. Most non-oss teams have multiple projects in Gitlab, including subgroups and what not, and it's common to register Gitlab Runners for the top group so they are available to all projects. This is a fairly common use case for the Kubernetes Gitlab Runner.

To enable this, the runner token for the group needs to be exposed by the `gitlab_group` resource and data source, so it can be piped into other Terraform resources such as `helm_release`.

jrdelarosa8:
Resolves #181

### DIFF
--- a/gitlab/data_source_gitlab_group.go
+++ b/gitlab/data_source_gitlab_group.go
@@ -2,9 +2,10 @@ package gitlab
 
 import (
 	"fmt"
+	"log"
+
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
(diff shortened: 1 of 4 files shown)


jrdelarosa8:
Hi @roidelapluie, do you have any additional requests for changes?

roidelapluie:
Thx

Issue #209, opened 2019-11-06
Manual removal of a resource on provider will raise a "404 Not Found" error without any recover


marimp:
Hi there,

### Terraform Version
not related to the terraform version

### Affected Resource(s)
all resources

### Terraform Configuration Files
not related to Configuration File

### Debug Output
N/A

### Panic Output
N/A

### Expected Behavior
A correct approach should be something like this (in this way the provider will be locally updated):

```go
	project, tag, err := projectAndTagFromID(d.Id())
	if err != nil {
		if strings.Contains(err.Error(), "404 Not Found") {
            d.SetId("")
            return nil
		}
		return err
	}
```

### Actual Behavior
there is a generic issue regarding the READ resource operation.
Considering as example the resource `resource_gitlab_tag_protection.go` (method `resourceGitlabTagProtectionRead`) code

```go
...
	project, tag, err := projectAndTagFromID(d.Id())
	if err != nil {
		return err
	}
...
```

In this way if the resource is not present on the server (404 Not Found response code), the provider return an error without deleting the local resource.
### Steps to Reproduce
N/A

### Important Factoids
N/A

### References
N/A

Is it possible to implement this approach in all resources?
Great Thanks for your attention
M.

borisceranic:
Hi @ymatsiuk (and others),

Related to the issue described above, and to solution implemented in #223, I'd like to report that this issue exists also on `gitlab_group_membership` resources.

I'm adding my own notes here because this issue is still open, and also judging by the context, which consists of this issue, #222 and #223, this issue is very much still relevant, despite the fact that #223 resolved the problem for a few resource types.

## Steps to reproduce

1. Write Terraform project that manages group members of a group with ID **6xxxxx2**.
2. Configure Terraform to use GitLab API key using GitLab user account with Owner role in group **6xxxxx2**, but this user must be different from user ID **3xxxxx1** (we will later remove this user from the group so Terraform will be unable to manage said group ID **6xxxxx2**).
3. Import existing group members into Terraform state (a few members, including user ID **3xxxxx1** with username **username123131**). 
4. Apply Terraform project to gitlab, applying the diff, possibly adding/removing members as a result.
5. GitLab group memberships are now fully managed by Terraform.
6. Remove member with user ID **3xxxxx1** from the group ID **6xxxxx2** out of band using GitLab UI.
7. Run `terraform plan`.

## Expected result

```
  # gitlab_group_membership.main_owners["username123131"] will be created
[...]

PR #222, opened 2019-11-22
Make read callbacks graceful

ymatsiuk:
Every time resource got changed manually, provider fails and the only fix is to remove failing resource from the state. If resource doesn't exist in gitlab, terraform should be informed and state cleaned, instead of failing with error.
Inspired by #209

### DIFF
--- a/gitlab/resource_gitlab_branch_protection.go
+++ b/gitlab/resource_gitlab_branch_protection.go
@@ -89,7 +89,9 @@ func resourceGitlabBranchProtectionRead(d *schema.ResourceData, meta interface{}
 
 	pb, _, err := client.ProtectedBranches.GetProtectedBranch(project, branch)
 	if err != nil {
-		return err
+		log.Printf("[DEBUG] failed to read gitlab branch protection for project %s, branch %s: %s", project, branch, err)
+		d.SetId("")
+		return nil
 	}
 
 	d.Set("project", project)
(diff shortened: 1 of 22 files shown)


roidelapluie:
This is the expected behaviour.

The provider uses gitlab ID's to identify resources. That ID can not change.

If a resource has a different id in gitlab, then it is not the resource that was created by terraform.

Note: in addition to that, if you do not have the rights to read resources, we also get a 404 from the API. It means that the state could be destroyed if someone launches the plug-in with bad credentials.

ymatsiuk:
@roidelapluie So in other words if I created protected branch with terraform and someone removed protection using gitlab admin interface it's expected for terraform to fail next plan by design?

I'm very curious, because [here](https://www.terraform.io/docs/extend/writing-custom-providers.html#implementing-read) it says:
> If the ID is updated to blank, this tells Terraform the resource no longer exists (maybe it was destroyed out of band). Just like the destroy callback, the `Read` function should gracefully handle this case.

roidelapluie:
For the protected branch we could adopt this.

However for git repos etc I would not do it because of the mentionned issue: if you do not have the rights to read resources, we also get a 404 from the API. It means that the state could be destroyed if someone launches the plug-in with bad credentials.

roidelapluie:
I would be happy if gitlab would return 403 but they return 404.

ymatsiuk:
`gitlab_project_share_group` causes the same amount of manual actions as `gitlab_branch_protection`. Shall I rebase for both resources and revert the rest?

roidelapluie:
Also gitlab_labels

ymatsiuk:
could you please re-open this PR?

PR #240, opened 2020-01-05
Force recreate gitlab_group when not found on read

stack72:
when a group was manually deleted, the read operation was throwing
an unrecoverable 404 error:

```
GET https://gitlab.com/api/v4/projects/16151185: 404 {message: 404 Project Not Found}
```

We should clear this from the state so that we can force recreation
of the group

```
▶ acctests gitlab TestAccGitlabGroup_
=== RUN   TestAccGitlabGroup_basic
--- PASS: TestAccGitlabGroup_basic (29.33s)
=== RUN   TestAccGitlabGroup_import
--- PASS: TestAccGitlabGroup_import (12.77s)
=== RUN   TestAccGitlabGroup_nested
--- PASS: TestAccGitlabGroup_nested (54.77s)
=== RUN   TestAccGitlabGroup_disappears
--- PASS: TestAccGitlabGroup_disappears (4.94s)
PASS
ok  	github.com/terraform-providers/terraform-provider-gitlab/gitlab	101.835s
```

### DIFF
--- a/gitlab/resource_gitlab_group.go
+++ b/gitlab/resource_gitlab_group.go
@@ -3,6 +3,7 @@ package gitlab
 import (
 	"fmt"
 	"log"
+	"net/http"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
@@ -117,8 +118,13 @@ func resourceGitlabGroupRead(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*gitlab.Client)
 	log.Printf("[DEBUG] read gitlab group %s", d.Id())
 
-	group, _, err := client.Groups.GetGroup(d.Id())
+	group, resp, err := client.Groups.GetGroup(d.Id())
 	if err != nil {
+		if resp != nil && resp.StatusCode == http.StatusNotFound {
+			log.Printf("[DEBUG] gitlab group %s not found so removing from state", d.Id())
+			d.SetId("")
+			return nil
+		}
 		return err
 	}
 
--- a/gitlab/resource_gitlab_group_test.go
+++ b/gitlab/resource_gitlab_group_test.go
@@ -156,6 +156,36 @@ func TestAccGitlabGroup_nested(t *testing.T) {
 	})
 }
 
+func TestAccGitlabGroup_disappears(t *testing.T) {
+	var group gitlab.Group
+	rInt := acctest.RandInt()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckGitlabGroupDestroy,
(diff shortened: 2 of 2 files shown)


gitlab/resource_gitlab_group.go
```diff
@@ -117,8 +118,13 @@ func resourceGitlabGroupRead(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*gitlab.Client)
 	log.Printf("[DEBUG] read gitlab group %s", d.Id())
 
-	group, _, err := client.Groups.GetGroup(d.Id())
+	group, resp, err := client.Groups.GetGroup(d.Id())
 	if err != nil {
+		if resp != nil && resp.StatusCode == http.StatusNotFound {
+			log.Printf("[DEBUG] gitlab group %s not found so removing from state", d.Id())
```
praneetloke: Sometimes a 404 can be due to the member losing access to the group, and not because a group is no longer there. I think only the user can truly tell if a GitLab Group is no longer there or if they just lost access. In the former's case, I think it is best to let the user manually remove the group from the state themselves. In a team-based environment where this resource is shared, this change would remove the group from the state accidentally while others still had access to it.

stack72: @praneetloke the normal convention is that if something is manually deleted then the TF provider should remove it from state. 404 means Not around - if it’s because of not access then I’d expect a different StatusCode

This is the correct fix and should actually be applied across all of the resources eventually IMO. This is how the aws provider works as well

praneetloke: > if it’s because of not access then I’d expect a different StatusCode

GitLab API returns 404 for groups that a user may not have access to. GitHub API does the same thing, too. It is similar to why when you try to access a private repo that you know exists, you receive a 404 instead of a 403 Forbidden.

stack72: So the question is, does the message that is returned any different between an access issue and a deleted repo? If not, that seems like a major upstream error in Gitlab.

praneetloke: > So the question is, does the message that is returned any different between an access issue and a deleted repo?

Nope. That's what I am trying to say :) But you can confirm this easily with the API.

roidelapluie: @stack72 the ambiguity of 404 in gitlab api's made us remove the 404 handling...

But I think lots of people complain the other way around.

I would be OK to re-add it for resources that support import.

praneetloke: Maybe a config option to let users pick if a 404 should be treated as a deleted resource? It would be global to this provider, of course and each resource that his behavior would have to honor that, of course.

ilpianista:
@stack72 FYI the test `TestAccGitlabGroup_disappears` fails randomly (apparently) on Travis.

ringods:
@ilpianista see #263 which is probably causing this.

PR #242, opened 2020-01-08
Changing notify_only_default_branch for branches_to_be_notified on slack

rin-amezquita:
Currently, `notify_only_default_branches` is deprecated. `branches_to_be_notified` is the replacement. Everything is documented [here](https://docs.gitlab.com/ee/api/services.html#slack-notifications)

This PR deletes the first and creates the second.

### DIFF
--- a/gitlab/resource_gitlab_service_slack.go
+++ b/gitlab/resource_gitlab_service_slack.go
@@ -37,7 +37,13 @@ func resourceGitlabServiceSlack() *schema.Resource {
 				Computed: true,
 			},
 			"notify_only_default_branch": {
-				Type:     schema.TypeBool,
+				Type:       schema.TypeBool,
+				Optional:   true,
+				Computed:   true,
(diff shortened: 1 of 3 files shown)


roidelapluie:
can we do it in a less disruptive way that would not force us to increase the major release flag?

rin-amezquita:
Then leaving both options should be the solution. Is there a prefered way of marking `notify_only_default_branches` as deprecated other than adding a comment in code?

roidelapluie:
Yes

```
    // When Deprecated is set, this attribute is deprecated.
    //
    // A deprecated field still works, but will probably stop working in near
    // future. This string is the message shown to the user with instructions on
    // how to address the deprecation.
    Deprecated string
```

https://godoc.org/github.com/hashicorp/terraform/helper/schema

rin-amezquita:
@roidelapluie Pinging you in case you missed the update.

roidelapluie:
Thanks!

PR #245, opened 2020-01-14
Update go-gitlab to v0.22.3

jrdelarosa8:
The changes here support #246

PR #246, opened 2020-01-14
Add support for commit_committer_check and reject_unsigned_commits in push rules

### LINKED ISSUE #202: [Bug] Unsupported argument: commit_committer_check
Hi there,

Getting error when `setting commit_committer_check` in `gitlab_project_push_rules`. According to [document](https://www.terraform.io/docs/providers/gitlab/r/project_push_rules.html#commit_committer_check), `commit_committer_check` should be supported.

### Terraform Version
+ Terraform v0.12.12
+ provider.gitlab v2.3.0

### Affected Resource(s)
- gitlab_project_push_rules

### Panic Output
```
Error: Unsupported argument

  on modules/gitlab_project/main.tf line 24, in resource "gitlab_project_push_rules" "push_rules":
  24:   commit_committer_check = true

An argument named "commit_committer_check" is not expected here.
```

jrdelarosa8:
Resolves #202, depends on #245

### DIFF
--- a/gitlab/resource_gitlab_project_push_rules.go
+++ b/gitlab/resource_gitlab_project_push_rules.go
@@ -57,6 +57,14 @@ func resourceGitlabProjectPushRules() *schema.Resource {
 				Type:     schema.TypeInt,
 				Optional: true,
 			},
+			"commit_committer_check": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+			"reject_unsigned_commits": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
 		},
 	}
 }
@@ -120,6 +128,8 @@ func resourceGitlabProjectPushRulesRead(d *schema.ResourceData, meta interface{}
 	d.Set("member_check", pushRules.MemberCheck)
 	d.Set("prevent_secrets", pushRules.PreventSecrets)
 	d.Set("max_file_size", pushRules.MaxFileSize)
+	d.Set("commit_committer_check", pushRules.CommitCommitterCheck)
+	d.Set("reject_unsigned_commits", pushRules.RejectUnsignedCommits)
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


roidelapluie:
There is not setter? is this computed? can you modify existing tests?

jrdelarosa8:
Hey @roidelapluie, I can definitely look at the existing tests. Can you clarify what you mean by:

> There is not setter? is this computed?

Thanks!

jrdelarosa8:
When creating a push rule resource via Terraform, we'll receive a 422 response because the push rule(s) already exist (new projects get push rules by default). 

This happens via [AddProjectPushRule](https://github.com/terraform-providers/terraform-provider-gitlab/blob/7dbdd83d27ebe937824de22f8b91ffbb71c897b9/gitlab/resource_gitlab_project_push_rules.go#L84) from the [client library](https://github.com/terraform-providers/terraform-provider-gitlab/blob/7dbdd83d27ebe937824de22f8b91ffbb71c897b9/vendor/github.com/xanzy/go-gitlab/projects.go#L1183).

It seems that we can get around this issue by not using the PUT/[ProjectPushRulesUpdate](https://github.com/terraform-providers/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_project_push_rules.go#L63) function and removing the create since this will continue to cause problems/confusion.

Any thoughts?

jrdelarosa8:
Are any maintainers available to take a look at my most recent question? @roidelapluie

roidelapluie:
Mmmh that's a tricky one. I will think about it and come back to you.

jrdelarosa8:
Do you have any additional thoughts?

armsnyder:
Can you update the test please?

armsnyder:
I like the idea of changing `AddProjectPushRule` to `EditProjectPushRule`, provided that it doesn't cause a problem where `EditProjectPushRule` also fails due to the push rule not existing, with the same exact race condition.

I also noticed in your change that you are reading these attributes, but you are not setting them in the create or update functions. That's probably what @roidelapluie meant.

armsnyder:
Closing this since we merged #422 where we added these attributes but moved the config into `gitlab_project`. Thanks!

PR #247, opened 2020-01-15
When changing the namespace_id, transfer project from old to new namespace

ringods:
This PR is work-in-progress and a request for comments.

Depends on: #245 

`go-gitlab` library supports project transferral between namespaces since version 0.22.3. With this support, it is possible to support this from the provider by checking the `namespace_id` property. If it is changed from the old value, the intent is to transfer the project.

Will add tests later, but comments on the production code is welcome.

PR #259, opened 2020-02-20
Fix for acceptance tests: lock Gitlab to v12.6.4

ringods:
After #247 I did some more exploratory testing with various Gitlab CE versions. My test setup is now based on a Vagrant setup I found [here](https://gist.github.com/cjtallman/b526d8c7d8b910ba4fd41eb51cd5405b).

Here is a timeline of the [Gitlab releases](https://about.gitlab.com/releases/categories/releases/), interleaved with the commits and Travis results of my previous PR:

09/01 - 12.6.3
13/01 - 12.6.4   

First commit on 15/01 - Travis build OK

22/01 - 12.7.0
24/01 - 12.7.2
30/01 - 12.7.4 + 12.6.6
31/01 - 12.7.5

Second commit on 31/01 - Rebased & Travis fails

13/02 - 12.7.6 + 12.6.7

So I started running the acceptance tests against different versions with these results:

12.6.4 = PASS
12.6.7 = FAIL
12.7.0 = FAIL

2nd round:
12.6.4 = PASS
12.6.6 = PASS

I'm locking the Gitlab version now to 12.6.4 as I have twice a success which makes it deterministic.

Probably there is a regression somewhere in Gitlab which makes the test results for the provider fail. While it can be fixed upstream, this provider has stable tests now.

### DIFF
--- a/scripts/start-gitlab.sh
+++ b/scripts/start-gitlab.sh
@@ -4,7 +4,7 @@ echo "Starting gitlab container..."
 if [[ -n $GITLAB_LICENSE_FILE ]]
 then
     extra="-v $PWD/license:/license -e GITLAB_LICENSE_FILE=/license/$GITLAB_LICENSE_FILE"
-    img=gitlab/gitlab-ee:12.6.4-ce.0
+    img=gitlab/gitlab-ee:12.6.4-ee.0
     if [[ ! -f license/$GITLAB_LICENSE_FILE ]]
     then
         echo No license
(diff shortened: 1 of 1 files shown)


roidelapluie:
Unfortunately I can't accept this. We first need to understand why it fails. maybe we need a new WaitForState function for example.

ringods:
@roidelapluie I understand.

But given my submission in #247 is not related to the test failures, would you mind cutting a release with that functionality included? I would like to go forward in my own projects with this without needing to commit custom provider binaries to the my setup.

ringods:
The provider is updated to work with soft-delete introduced in Gitlab Enterprise Edition v12.
Released in provider v2.5.1

PR #261, opened 2020-02-21
Allow to change user email without recreating the resource

ilpianista:
I imported an user into the terraform state and after I set the mandatory arguments the provider wanted to delete and recreate such user even if I didn't change anything.

This PRs maps the missing user attributes into the state.

### DIFF
--- a/gitlab/resource_gitlab_user.go
+++ b/gitlab/resource_gitlab_user.go
@@ -48,7 +48,6 @@ func resourceGitlabUser() *schema.Resource {
 			"email": {
 				Type:     schema.TypeString,
 				Required: true,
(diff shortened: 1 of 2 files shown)


ilpianista:
@roidelapluie this was bigger than I expected (there were a lot of problems under the hood).
Please take a look to each commit to understand the purpose of each change.

I [opened a bug to GitLab](https://gitlab.com/gitlab-org/gitlab/issues/207818) about the 'external' flag causing issues after update.

armsnyder (approved):
Thanks for fixing this, and fixing the test! If you have any ideas for how to make tests less prone to mistakes in the future, I'd be interested.

Issue #193, opened 2019-10-17
Error importing user from gitlab to terraform resource


Procras:
### Terraform Version
Terraform v0.12.10
+ provider.gitlab v2.3.0

### Affected Resource(s)
- gitlab_user 

### Terraform Configuration Files
```hcl
# Configure the GitLab Provider
provider "gitlab" {
  token    = "${var.gitlab_token}"
  base_url = "https://localhost/gitlab"
  insecure = true
}

resource "gitlab_user" "user" {
  name             = "firstname lastname"
  username         = "user"
  password         = "minimum8"
  email            = "email@email.io"
  projects_limit   = 100000
  can_create_group = true
}
```

### Debug Output
None

### Panic Output
None

### Expected Behavior
Terraform plan should output no changes

### Actual Behavior
Terraform wants to force email replacement on current user

```hcl
  # gitlab_user.user must be replaced
-/+ resource "gitlab_user" "user" {
        can_create_group  = true
      + email             = "email@email.io" # forces replacement
      ~ id                = "25" -> (known after apply)
      + is_admin          = false
      + is_external       = false
        name              = "firstname lastname"
      + password          = (sensitive value)
        projects_limit    = 100000
      + skip_confirmation = true
        username          = "user"
    }

Plan: 1 to add, 0 to change, 1 to destroy.
```

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Create new user in gitlab
2. Create resource representing user in a  .tf
3. `terraform import gitlab_user <user_id>`
4. `terraform fmt && terraform validate && terraform plan`

### Important Factoids
No

### References
None

mnology:
It should be noted that `terraform import` appears to create the user in the state with `email: null`. Importing with null will mean that imported user resources will always have to be recreated due to perceived email attribute changes.

I've been able to tweak the state manually to fix the email attr in order to get a sensible plan in a test environment, but this isn't feasible for most cases.

deweysasser:
Does this generate any interest from a developer?  The issue renders the gitlab provider substantially problematical to use.

PR #262, opened 2020-02-21
Make password and reset_password exclusive

ilpianista:
The documentation works when either `password` or `reset_password` is set.

### DIFF
--- a/gitlab/resource_gitlab_user_test.go
+++ b/gitlab/resource_gitlab_user_test.go
@@ -100,7 +100,7 @@ func TestAccGitlabUser_password_reset(t *testing.T) {
 			// Test that either password or reset_password is needed
 			{
 				Config:      testAccGitlabUserConfigWrong(rInt),
-				ExpectError: regexp.MustCompile(`\sAt least one of either password or reset_password must be defined`),
+				ExpectError: regexp.MustCompile("At least one of either password or reset_password must be defined"),
 			},
 			// Create a user without a password
 			{
--- a/gitlab/resource_gitlab_user_test.go
+++ b/gitlab/resource_gitlab_user_test.go
@@ -2,6 +2,7 @@ package gitlab
 
 import (
 	"fmt"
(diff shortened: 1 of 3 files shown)


roidelapluie:
Can we have an acceptance test with an empty password?

ilpianista:
@roidelapluie is this OK as test? I don't set a password during the creation, but then it's set during the user update test.

Still, I don't think we can check if the password has been set because the password is (obviously) never sent back by GitLab. Thus, I removed that part.

ilpianista:
Making the password optional has been solved by #193, so this PR now just ensure that either `password` or `reset_password` is set, not both.

armsnyder:
@ilpianista Can you add two new tests?
- One where you only set `reset_password`, just to ensure that it doesn't explode if `password` is undefined.
- One where you set neither field and expect an error.

ilpianista:
@armsnyder I implemented 2). To implement 1), I simply removed the `password` argument from the existing test.

gitlab/resource_gitlab_user_test.go
```diff
@@ -96,7 +97,12 @@ func TestAccGitlabUser_password_reset(t *testing.T) {
 		Providers:    testAccProviders,
 		CheckDestroy: testAccCheckGitlabGroupDestroy,
 		Steps: []resource.TestStep{
-			// Create a user
+			// Test that either password or reset_password is needed
+			{
+				Config:      testAccGitlabUserConfigWrong(rInt),
+				ExpectError: regexp.MustCompile(`\sAt least one of either password or reset_password must be defined`),
```
armsnyder, line 103: Nitpick but the `\s` made me scratch my head for a minute. I think the capital `A` alone serves the intended purpose with less head scratching. Thoughts?

```suggestion
				ExpectError: regexp.MustCompile("At least one of either password or reset_password must be defined"),
```

armsnyder (approved):
Thanks!

PR #265, opened 2020-02-25
Allow to add an user to a group by username

ilpianista:
This allows to add a member to an existing group also by username.

### DIFF
--- a/gitlab/resource_gitlab_group_membership.go
+++ b/gitlab/resource_gitlab_group_membership.go
@@ -1,6 +1,7 @@
 package gitlab
 
 import (
+	"fmt"
 	"log"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


armsnyder:
I appreciate the effort on this, but it goes against the design principle that [the resource schema should closely match the underlying API](https://www.terraform.io/docs/extend/hashicorp-provider-design-principles.html#resource-and-attribute-schema-should-closely-match-the-underlying-api) since the [API](https://docs.gitlab.com/ee/api/members.html) requires specifying a user ID. We have a [data source for resolving a username to an ID](https://www.terraform.io/docs/providers/gitlab/d/user.html) which serves this use case.

@roidelapluie What do you think?

roidelapluie:
Yess for this usecase the data source should be used.

armsnyder:
Alright I'll be the monster and close this PR. 👹 @ilpianista If you want to, you can always create an issue where we can discuss this some more. Thanks for your cleanly coded contribution as always!

ilpianista:
Sure, no problem!

PR #266, opened 2020-02-26
Do not fail when the group member already exists

ilpianista:
If a member already exists in the group (either because someone manually added it or because that user owns the `GITLAB_TOKEN`), the `gitlab_group_membership` resource creations fails.

Instead, we can handle that failure and update the existing member to match what we do expect.

### DIFF
--- a/gitlab/resource_gitlab_group_membership.go
+++ b/gitlab/resource_gitlab_group_membership.go
@@ -64,12 +64,27 @@ func resourceGitlabGroupMembershipCreate(d *schema.ResourceData, meta interface{
 	}
 	log.Printf("[DEBUG] create gitlab group groupMember for %d in %s", options.UserID, groupId)
 
-	groupMember, _, err := client.GroupMembers.AddGroupMember(groupId, options)
+	_, resp, err := client.GroupMembers.AddGroupMember(groupId, options)
 	if err != nil {
-		return err
+		if resp != nil && resp.StatusCode == http.StatusConflict {
+			options := gitlab.EditGroupMemberOptions{
+				AccessLevel: &accessLevelId,
+				ExpiresAt:   &expiresAt,
+			}
+			log.Printf("[DEBUG] update gitlab group membership %v for %s", userId, groupId)
+
+			_, _, err := client.GroupMembers.EditGroupMember(groupId, userId, &options)
+			if err != nil {
+				return err
+			}
+		} else {
+			return err
+		}
 	}
-	userIdString := strconv.Itoa(groupMember.ID)
+
+	userIdString := strconv.Itoa(userId)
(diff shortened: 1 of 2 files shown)


armsnyder:
Thanks for the contribution! Unfortunately this change goes against the standard Terraform provider design. Trying to create a resource that already exists _should_ result in an error. The user is expected to use `terraform import` to resolve the conflict themselves.

I'm going to close this PR for this reason, but if there is a usability problem you want to bring up, you can always create an issue! 😄

ilpianista:
@armsnyder maybe I didn't explain the behavior correctly. Let me try again.

If you have the following code:

```
resource "gitlab_group" "group" {
}

resource "gitlab_group_membership" "membership" {
  user_id = <ID of the user owning the GITLAB_TOKEN>
}
```

The creation of the resource `membership` will fail because GitLab always adds the owner of the `GITLAB_TOKEN` as `owner` of the group he creates.

Don't you consider this like a bug? I'm doing something wrong?

armsnyder:
Got it. Thanks for explaining. Yeah that sounds like a bug.

armsnyder:
However, is there a way to fix this PR so that it _does_ return an error if the user referenced in `gitlab_group_membership` is not the owner of the API token, but rather it is another user who already has a membership? It's important to return an error and let the Terraform user resolve it with `terraform import` in this case, because we don't want to silently mangle the membership that is already managed by the user out-of-band.

ilpianista:
> However, is there a way to fix this PR so that it _does_ return an error if the user_id referenced in `gitlab_group_membership` is not the owner of the API token, but rather it is another user who already has a membership? It's important to return an error and let the Terraform user resolve it with `terraform import` in this case, because we don't want to silently mangle the membership that is already managed out-of-band.

Please check it now.

ilpianista:
@armsnyder May you please review it?

ilpianista:
@armsnyder @nicholasklick Up.

armsnyder:
I think this needs to be discussed in an issue first, so we can assess the impact of it. I really appreciate the contribution. My hesitation is that this change makes the resource behave in a somewhat atypical way.
1. This change may go against the user's expectations, since Terraform is not designed to successfully create resources that already exist.
2. There may be other usage patterns that this change can adversely affect. One example: What if the user creates the group first, then applies, then creates the membership second after that? I would think this should fail and expect the user to import the membership resource.
3. Similar to point 2, the logic that you've added which checks that the Terraform user is the member can cause an adverse effect, since there is no guarantee that the group was created by the Terraform user in the first place.

timofurrer:
Since this MR would need some refactoring anyways, I'm going to close it for now and we can continue the discussion in #792. We can re-open and create a new one once we have a clear way forward.

@ilpianista thanks for the contribution and feel free to chime in in #792 🍰

PR #267, opened 2020-02-26
Do not fail when the group member does not exist

### LINKED ISSUE #254: User missing from group makes refresh fail
### Terraform Version
```
Terraform v0.12.18
+ provider.gitlab v2.5.0
```

### Affected Resource(s)
- gitlab_group_membership

Just a normal `gitlab_group_membership` instance.

### Terraform output
```
Error: GET https://gitlab.com/api/v4/groups/XXXXXXX/members/YYYYYYY: 404 {message: 404 Not found}
```
### Expected Behavior
When a user is missing from a group, gitlab_group_membership should just mark it as pending to add.

### Actual Behavior
The refresh process fails with a 404 not found, and terraform exits

### Steps to Reproduce
1. `terraform apply` a `gitlab_group_membership`
2. Remove the user from the group
3. `terraform refresh` (or any other command that tries to update the state in memory)

### Important Factoids
This is in a private gitlab subgroup.

ilpianista:
Same as discussed in #240: if the group member doesn't exist, then remove it from the state also.

Fixes #254

### DIFF
--- a/gitlab/resource_gitlab_group_membership.go
+++ b/gitlab/resource_gitlab_group_membership.go
@@ -2,6 +2,7 @@ package gitlab
 
 import (
 	"log"
+	"net/http"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


armsnyder (approved):
lgtm

PR #269, opened 2020-02-29
Delete and overwrite existing project push rules on new resource creation

awkaplan:
Currently, this resource fails on every apply, regardless of whether or not push rules actually exist for a given project.  With this change, existing push rules are overwritten with the parameters configured in Terraform if a new resource is being created.

### DIFF
--- a/gitlab/resource_gitlab_project_push_rules.go
+++ b/gitlab/resource_gitlab_project_push_rules.go
@@ -98,7 +98,17 @@ func resourceGitlabProjectPushRulesCreate(d *schema.ResourceData, meta interface
 
 	pushRules, _, err := client.Projects.AddProjectPushRule(project, options)
 	if err != nil {
-		return err
+		// Delete existing push rules
+		_, err := client.Projects.DeleteProjectPushRule(project)
+		if err != nil {
+			return err
+		}
+
+		// Add push rules
+		pushRules, _, err = client.Projects.AddProjectPushRule(project, options)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


awkaplan:
@ringods @roidelapluie can I please trouble you for a look?

gitlab/resource_gitlab_project_push_rules.go
```diff
@@ -98,7 +98,17 @@ func resourceGitlabProjectPushRulesCreate(d *schema.ResourceData, meta interface
 
 	pushRules, _, err := client.Projects.AddProjectPushRule(project, options)
 	if err != nil {
-		return err
+		// Delete existing push rules
```
armsnyder, line 101: You should check the cause of the error before anything else.

gitlab/resource_gitlab_project_push_rules.go
```diff
@@ -98,7 +98,17 @@ func resourceGitlabProjectPushRulesCreate(d *schema.ResourceData, meta interface
...
+		_, err := client.Projects.DeleteProjectPushRule(project)
```
armsnyder, line 102: It would be better to use the EditProjectPushRule API method, especially since some of the issues mention rules getting automatically recreated when they try to delete push rules manually.

armsnyder:
Hey @awkaplan are you still working on this PR? There are unresolved comments. 😄

armsnyder:
Closing this since we merged #422. Thanks!

PR #271, opened 2020-03-05
Add masked attribute to group variables

kirecek:
Related issue https://github.com/terraform-providers/terraform-provider-gitlab/issues/243

### DIFF
--- a/gitlab/resource_gitlab_group_variable.go
+++ b/gitlab/resource_gitlab_group_variable.go
@@ -45,6 +45,11 @@ func resourceGitlabGroupVariable() *schema.Resource {
 				Optional: true,
 				Default:  false,
 			},
+			"masked": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


kirecek:
Don't understand why the test is failing. I cannot replicate it locally (using gitlab:latest):

```
$ TESTARGS="-run TestAccGitlabGroup_disappears" make testacc
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestAccGitlabGroup_disappears -timeout 120m
?   	github.com/terraform-providers/terraform-provider-gitlab	[no test files]
=== RUN   TestAccGitlabGroup_disappears
--- PASS: TestAccGitlabGroup_disappears (0.42s)
PASS
ok  	github.com/terraform-providers/terraform-provider-gitlab/gitlab	0.431s
```

Also looking at code and does not make any sense to me why would the state be empty

emmanuelCarre:
Hi,
I am not familiar with travis ci, let me know what you think about my hypothesis:
Gitlab version is not fixed: https://github.com/terraform-providers/terraform-provider-gitlab/blob/master/scripts/start-gitlab.sh#L14; and I didn't see where we define pull strategy. I think CI use an too old image because argument is added into API into master code but it will be release in few days: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/25283.

A simple way to fix it is: add `docker pull $img` at line 16 of `start-gitlab.sh` but it will increase start time to everyone. A second way could be add argument to this script to force pull image when needed (in travis for example).

emmanuelCarre:
Hi,

There is something else to do to merge this merge request? May I help?
To be honnest, I am really interested about this feature :) 

Thank you for this development.

kirecek:
> There is something else to do to merge this merge request? May I help?

The issue I mentioned above was fixed after I rebased master. Only thing we can is to wait for a review.

Rebased again now to pull master.

sirlatrom:
@ringods Any chance you could give this a quick look?

PR #273, opened 2020-03-09
Update go-gitlab to v0.28.0

wernerb:
Updates go-gitlab to v0.28.0

This update has some jira integration changes that i will create a followup PR on that will reduce spamming. Right now jira integration by default creates a jira comment for every commit back to the jira issue.


wernerb:
There is a test failing but it seems to be unrelated to these changes. See https://github.com/terraform-providers/terraform-provider-gitlab/pull/271 for a pull request that has the same issue.

PR #277, opened 2020-03-12
Adding support for project templates

ahmet2mir:
Hello,

I added support to templates and custom templates.

But use_custom_template and group_with_project_templates_id are EE features.

I re-use the same test than "initializereadme" but I don't know how to write tests on EE features  because I have to:
* create a group
* create a template repo (simple git repo)
* declare the group as a template in admin settings

I don't found in tests how you test EE features, anyone know how to deal with ?

### DIFF
--- a/gitlab/resource_gitlab_project.go
+++ b/gitlab/resource_gitlab_project.go
@@ -208,6 +208,22 @@ var resourceGitLabProjectSchema = map[string]*schema.Schema{
 		Type:     schema.TypeBool,
 		Optional: true,
 	},
+	"template_name": {
+		Type:     schema.TypeString,
+		Optional: true,
+	},
+	"template_project_id": {
+		Type:     schema.TypeInt,
+		Optional: true,
+	},
+	"use_custom_template": {
+		Type:     schema.TypeBool,
+		Optional: true,
+	},
+	"group_with_project_templates_id": {
+		Type:     schema.TypeInt,
+		Optional: true,
+	},
 }
 
 func resourceGitlabProject() *schema.Resource {
@@ -299,6 +315,22 @@ func resourceGitlabProjectCreate(d *schema.ResourceData, meta interface{}) error
 
 	if v, ok := d.GetOk("import_url"); ok {
 		options.ImportURL = gitlab.String(v.(string))
+	}
+
+	if v, ok := d.GetOk("template_name"); ok {
+		options.TemplateName = gitlab.String(v.(string))
+	}
+
+	if v, ok := d.GetOk("template_project_id"); ok {
+		options.TemplateProjectID = gitlab.Int(v.(int))
+	}
+
+	if v, ok := d.GetOk("use_custom_template"); ok {
+		options.UseCustomTemplate = gitlab.Bool(v.(bool))
+	}
+
+	if v, ok := d.GetOk("group_with_project_templates_id"); ok {
+		options.GroupWithProjectTemplatesID = gitlab.Int(v.(int))
 	}
 
 	log.Printf("[DEBUG] create gitlab project %q", *options.Name)
--- a/gitlab/resource_gitlab_project_test.go
+++ b/gitlab/resource_gitlab_project_test.go
@@ -197,7 +197,79 @@ func TestAccGitlabProject_initializeWithReadme(t *testing.T) {
 				Config: testAccGitlabProjectConfigInitializeWithReadme(rInt),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckGitlabProjectExists("gitlab_project.foo", &project),
-					testAccCheckGitlabProjectInitializeWithReadme(&project, &testAccGitlabProjectExpectedAttributes{
+					testAccCheckGitlabProjectDefaultBranch(&project, &testAccGitlabProjectExpectedAttributes{
+						DefaultBranch: "master",
+					}),
+				),
+			},
+		},
+	})
+}
+
+func TestAccGitlabProject_templateName(t *testing.T) {
+	var project gitlab.Project
+	rInt := acctest.RandInt()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckGitlabProjectDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGitlabProjectConfigTemplateName(rInt),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGitlabProjectExists("gitlab_project.template-name", &project),
+					testAccCheckGitlabProjectDefaultBranch(&project, &testAccGitlabProjectExpectedAttributes{
+						DefaultBranch: "master",
+					}),
+				),
+			},
+		},
+	})
+}
+
+func TestAccGitlabProject_templateNameCustom(t *testing.T) {
+	if v := os.Getenv("GITLAB_EE"); v == "no" {
+		return
+	}
+	var project gitlab.Project
+	rInt := acctest.RandInt()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckGitlabProjectDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGitlabProjectConfigTemplateNameCustom(rInt),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGitlabProjectExists("gitlab_project.template-name-custom", &project),
+					testAccCheckGitlabProjectDefaultBranch(&project, &testAccGitlabProjectExpectedAttributes{
+						DefaultBranch: "master",
+					}),
+				),
+			},
+		},
+	})
+}
+
+func TestAccGitlabProject_templateProjectID(t *testing.T) {
+	if v := os.Getenv("GITLAB_EE"); v == "no" {
+		return
+	}
+	var project gitlab.Project
+	rInt := acctest.RandInt()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckGitlabProjectDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGitlabProjectConfigTemplateProjectID(rInt),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGitlabProjectExists("gitlab_project.template-id", &project),
+					testAccCheckGitlabProjectDefaultBranch(&project, &testAccGitlabProjectExpectedAttributes{
 						DefaultBranch: "master",
 					}),
 				),
@@ -487,10 +559,10 @@ func testAccCheckAggregateGitlabProject(expected, received *gitlab.Project) reso
 	return resource.ComposeAggregateTestCheckFunc(checks...)
 }
 
-func testAccCheckGitlabProjectInitializeWithReadme(project *gitlab.Project, want *testAccGitlabProjectExpectedAttributes) resource.TestCheckFunc {
+func testAccCheckGitlabProjectDefaultBranch(project *gitlab.Project, want *testAccGitlabProjectExpectedAttributes) resource.TestCheckFunc {
 	return func(s *terraform.State) error {
 		if project.DefaultBranch != want.DefaultBranch {
-			return fmt.Errorf("got description %q; want %q", project.DefaultBranch, want.DefaultBranch)
+			return fmt.Errorf("got default branch %q; want %q", project.DefaultBranch, want.DefaultBranch)
 		}
 
 		return nil
@@ -720,6 +792,44 @@ resource "gitlab_project" "foo" {
   path = "foo.%d"
   description = "Terraform acceptance tests"
   initialize_with_readme = true
+}
+	`, rInt, rInt)
+}
+
+func testAccGitlabProjectConfigTemplateName(rInt int) string {
+	return fmt.Sprintf(`
+resource "gitlab_project" "template-name" {
+  name = "template-name-%d"
+  path = "template-name.%d"
+  description = "Terraform acceptance tests"
+  template_name = "rails"
+  default_branch = "master"
+}
+	`, rInt, rInt)
+}
+
+func testAccGitlabProjectConfigTemplateNameCustom(rInt int) string {
+	return fmt.Sprintf(`
+resource "gitlab_project" "template-name-custom" {
+  name = "template-name-custom-%d"
+  path = "template-name-custom.%d"
+  description = "Terraform acceptance tests"
+  template_name = "myrails"
+  use_custom_template = true
+  default_branch = "master"
+}
+	`, rInt, rInt)
+}
+
+func testAccGitlabProjectConfigTemplateProjectID(rInt int) string {
+	return fmt.Sprintf(`
+resource "gitlab_project" "template-id" {
+  name = "template-id-%d"
+  path = "template-id.%d"
+  description = "Terraform acceptance tests"
+  template_project_id = 999
+  use_custom_template = true
+  default_branch = "master"
 }
 	`, rInt, rInt)
 }
--- a/vendor/github.com/xanzy/go-gitlab/client_options.go
+++ b/vendor/github.com/xanzy/go-gitlab/client_options.go
@@ -24,6 +24,16 @@ func WithCustomBackoff(backoff retryablehttp.Backoff) ClientOptionFunc {
 	}
 }
 
+// WithCustomLimiter injects a custom rate limiter to the client.
+func WithCustomLimiter(limiter RateLimiter) ClientOptionFunc {
+	return func(c *Client) error {
+		c.configureLimiterOnce.Do(func() {
+			c.limiter = limiter
+		})
+		return nil
+	}
+}
+
 // WithCustomRetry can be used to configure a custom retry policy.
 func WithCustomRetry(checkRetry retryablehttp.CheckRetry) ClientOptionFunc {
 	return func(c *Client) error {
--- a/website/docs/r/project.html.markdown
+++ b/website/docs/r/project.html.markdown
@@ -83,6 +83,14 @@ The following arguments are supported:
 
 * `initialize_with_readme` - (Optional) Create master branch with first commit containing a README.md file.
 
+* `template_name` - (Optional) When used without use_custom_template, name of a built-in project template. When used with use_custom_template, name of a custom project template.
+
+* `template_project_id` - (Optional)  When used with use_custom_template, project ID of a custom project template. This is preferable to using template_name since template_name may be ambiguous (enterprise edition).
+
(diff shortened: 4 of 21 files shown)


nicholasklick:
@armsnyder @roidelapluie if this is looking good can we get it merged?

sfang97:
@ahmet2mir This needed a rebase so I went ahead and kept both changes (testAccGitlabProjectConfigTemplateName from this PR and testAccGitlabProjectConfigImportURL from master), WDYT?

ahmet2mir:
LGTM :)

Le lun. 14 sept. 2020 à 22:09, Serena Fang <notifications@github.com> a
écrit :

sfang97:
The branch needed a rebase so I went ahead and rebased it, there was a small conflict that was easy to resolve (I kept both changes), which created a bunch of errors so I want to revert my merge. I'm getting this error though: " ! [remote rejected] project-template-support -> project-template-support (refusing to allow an OAuth App to create or update workflow `.github/workflows/acceptance-tests.yml` without `workflow` scope)
error: failed to push some refs to 'https://github.com/ahmet2mir/terraform-provider-gitlab'" 
@ahmet2mir or someone else, could you revert my b6d5e38 commit? Sorry for messing with your branch 😅

ahmet2mir:
I will recreate, too much conflicts and my changes are not so big

ahmet2mir:
@sfang97 MR recreated with https://github.com/gitlabhq/terraform-provider-gitlab/pull/414

ahmet2mir:
I added support to templates and custom templates (recreate from #277)

I re-use the same test than initializereadme (no state).

use_custom_template, template_project_id and group_with_project_templates_id are EE features
template_name is EE and CE feature.

I see that Gitlab will took probably this repo control #376 so they probably rework all tests to split EE and CE.
We could use tags with go test but i'm not confortable with that and we have to change everywhere

So to simplify EE tests, I added an environment variable GITLAB_EE, when we enter in this test, it's skipped when not EE.

on CE

=== RUN   TestAccGitlabProject_templateNameCustom
--- PASS: TestAccGitlabProject_templateNameCustom (0.00s)
=== RUN   TestAccGitlabProject_templateProjectID
--- PASS: TestAccGitlabProject_templateProjectID (0.00s)
In other hand, in start-gitlab, via rails, I:

create a group
create a template repo (simple git repo)
declare the group as a template in admin settings
Thoses lines works with CE too (so I don't remove them, if in future Gitlab move thoses EE features to CE, we just have to remove the GITLAB_EE test in tests)

sfang97:
> I will recreate, too much conflicts and my changes are not so big. I did a push force to remove the binary from history and now we have too much commits.

@ahmet2mir Thanks for reopening, sorry again for messing up your branch!

armsnyder:
@ahmet2mir Do you need the new `GITLAB_EE` variable? In other EE-only tests we use `SkipFunc: isRunningInCE` [as in here](https://github.com/gitlabhq/terraform-provider-gitlab/blob/6a7f6e359e4583c5da7a115cba67e0ea27e7cc06/gitlab/resource_gitlab_service_github_test.go#L67).

Also, I don't think we want to add any test-specific setup to the `start-gitlab.sh` file, because the tests should be able to run on existing GitLab instances. Can you create the template in the test code?[TestAccGitlabProject_importURL](https://github.com/gitlabhq/terraform-provider-gitlab/blob/de14c7da4485bdcadf0e5f3c31398165b576a39b/gitlab/resource_gitlab_project_test.go#L356) is an example of how to add some custom setup logic.

ahmet2mir:
@armsnyder I don't need the env var, SkipFunc: isRunningInCE is what I missed. thanks.

For `start-gitlab.sh` I didn't found another way to achieve this. This setting is defined in EE only and not available via API (or don't found it).

It's the param `custom_project_templates_group_id`, found on WEB https://gitlab.com/gitlab-org/gitlab/-/blob/master/ee/app/models/ee/application_setting.rb#L108 but not on API

I could only create the group, the project myrails based on rails, get the id and send to the tests but I can't change EE settings to declare the group as template.

gitlab/resource_gitlab_project.go
```diff
@@ -301,6 +317,22 @@ func resourceGitlabProjectCreate(d *schema.ResourceData, meta interface{}) error
 		options.ImportURL = gitlab.String(v.(string))
 	}
 
+	if v, ok := d.GetOk("template_name"); ok {
+		options.TemplateName = gitlab.String(v.(string))
```
sfang97, line 321: Where does `options.TemplateName` come from? It's not in the `CreateProjectOptions` struct or `resourceGitlabProjectSetToState`, does it need to be declared somewhere?

gitlab/resource_gitlab_project.go
```diff
@@ -301,6 +317,22 @@ func resourceGitlabProjectCreate(d *schema.ResourceData, meta interface{}) error
...
+	}
+
+	if v, ok := d.GetOk("template_project_id"); ok {
+		options.TemplateProjectID = gitlab.Int(v.(int))
```
sfang97, line 325: Same thing here with `options.TemplateProjectID`, should it be declared somewhere? Also bit of a warning: I tried opening up my own PR based on yours here: https://github.com/gitlabhq/terraform-provider-gitlab/pull/377 but when I had both `TemplateName` and `TemplateProjectID` in `CreateProjectOptions`, I kept getting an error that TemplateName and TemplateProjectID had to be mutually exclusive. Do you know how to fix that?

gitlab/resource_gitlab_project_test.go
```diff
@@ -197,7 +197,75 @@ func TestAccGitlabProject_initializeWithReadme(t *testing.T) {
...
+	rInt := acctest.RandInt()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckGitlabProjectDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGitlabProjectConfigTemplateName(rInt),
+				Check: resource.ComposeTestCheckFunc(
```
sfang97, line 220: This `Check` step is a bit confusing, since it only checks that the project exists with name `"template-name"` and the default branch is master. Could you add an additional check that verifies that the project was indeed created with attribute `template_name = "rails"`?

gitlab/resource_gitlab_project_test.go
```diff
@@ -197,7 +197,75 @@ func TestAccGitlabProject_initializeWithReadme(t *testing.T) {
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckGitlabProjectDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config:   testAccGitlabProjectConfigTemplateNameCustom(rInt),
+				SkipFunc: isRunningInCE,
+				Check: resource.ComposeTestCheckFunc(
```
sfang97, line 243: Same thing here, this test would be stronger if it checked that attribute `template_name` was set to  `"myrails"`

gitlab/resource_gitlab_project_test.go
```diff
@@ -197,7 +197,75 @@ func TestAccGitlabProject_initializeWithReadme(t *testing.T) {
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckGitlabProjectDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config:   testAccGitlabProjectConfigTemplateProjectID(rInt),
+				SkipFunc: isRunningInCE,
+				Check: resource.ComposeTestCheckFunc(
```
sfang97, line 266: Same thing here, this test would be stronger if it checked that attribute template_project_id was set to 999

vendor/github.com/xanzy/go-gitlab/client_options.go
```diff
@@ -24,6 +24,16 @@ func WithCustomBackoff(backoff retryablehttp.Backoff) ClientOptionFunc {
 	}
 }
 
+// WithCustomLimiter injects a custom rate limiter to the client.
```
sfang97, line 27: I'm not sure if you should be checking in the `go mod vendor` files, if you need changes from the upstream library that aren't in the newest package version you could use the replace directive to point to your own upstream branch.

https://golang.org/ref/mod#go-mod-file-replace

website/docs/r/project.html.markdown
```diff
@@ -83,6 +83,14 @@ The following arguments are supported:
 
 * `initialize_with_readme` - (Optional) Create master branch with first commit containing a README.md file.
 
+* `template_name` - (Optional) When used without use_custom_template, name of a built-in project template. When used with use_custom_template, name of a custom project template.
+
+* `template_project_id` - (Optional)  When used with use_custom_template, project ID of a custom project template. This is preferable to using template_name since template_name may be ambiguous (enterprise edition).
```
sfang97, line 88: I know this is straight from the Gitlab docs, but perhaps you could add a note in here about `template_name` and `template_project_id` being mutually exclusive.

gitlab/resource_gitlab_project.go
sfang97: I struggled with the mutually exclusive error for a while, if you figure it out please let me know how! If not, perhaps we can stick to either TemplateName or TemplateProjectID but not both- since the docs say TemplateProjectID is preferable, maybe just that? Ideally we get around the exclusivity error though.

sfang97 (commented):
Thanks for your work on this @ahmet2mir, I've left a few comments.

gitlab/resource_gitlab_project.go
ahmet2mir: In the PR I upgraded go gitlab package to 0.37.0.
You could found here https://github.com/xanzy/go-gitlab/blob/v0.37.0/projects.go#L505

gitlab/resource_gitlab_project.go
ahmet2mir: I already see in some codes a way to achieve this (probably in ValidateFunc), will check if possible otherwise I will update the doc (Gitlab doc doesn't said something like that but it's 'trivial')

gitlab/resource_gitlab_project_test.go
ahmet2mir: Like `initialize with readme`, init with a template is 'stateless'. So don't had the information that the project was created with a template.

gitlab/resource_gitlab_project_test.go
ahmet2mir: same upper

gitlab/resource_gitlab_project_test.go
ahmet2mir: same upper

vendor/github.com/xanzy/go-gitlab/client_options.go
ahmet2mir: It's not my changes but a sync from 0.37.0 (from 0.34.0)

website/docs/r/project.html.markdown
ahmet2mir: will check mutal exclusive first and will adapt

gitlab/resource_gitlab_project.go
sfang97: I _think_ `options.TemplateName` reads from the `options` struct in resource_gitlab_project.go :277? In which case the attribute needs to be set there

gitlab/resource_gitlab_project_test.go
sfang97: Hm okay. In that case would you mind sanity checking this then? By TF creating a project based on a template name, custom template, and project template ID, then manually checking that the template was applied?

gitlab/resource_gitlab_project.go
ahmet2mir: Ok found an example here, it's tested in code directly https://github.com/terraform-providers/terraform-provider-azurerm/blob/687e20310ee035694a83606a0815e702dd357f12/azurerm/internal/services/batch/batch_pool_resource.go#L542
So I added the same test and in acceptance testing that the project doesn't exists.

gitlab/resource_gitlab_project_test.go
ahmet2mir: To resolve this, I test that README.md is present, should be enough to ensure that our repo contains any template type

website/docs/r/project.html.markdown
ahmet2mir: Added a note

gitlab/resource_gitlab_project.go
ahmet2mir: I reads from `d ResourceData`, template options doesn't have default values (like `initialize_with_readme`) and are optional, it's better to no set something in a [json as null](https://golang.org/pkg/encoding/json/#:~:text=The%20JSON%20null%20value%20unmarshals,value%20and%20produces%20no%20error.) (we got some side effects on other providers)

sfang97: Ah I see, thanks for the explanation

sfang97:
@ahmet2mir This looks good to me, thanks for working on it! Might be worth it to get another set of eyes on this before it goes in. @armsnyder Could you take a look? 😄

armsnyder:
@ahmet2mir Thanks for the detailed response about the template group!

@roidelapluie What do you think about the change to `start-gitlab.sh`? Ahmet is using the rails console to configure the template project group, which is used in the test, because there's isn't a GitLab API for creating the template project group. Since you know more about the ways tests can be run, do you think it's ok, or do you have any suggestions there?

@sfang97 Sure I'll review.

gitlab/resource_gitlab_project.go
```diff
@@ -208,6 +208,22 @@ var resourceGitLabProjectSchema = map[string]*schema.Schema{
...
 		Optional: true,
 	},
+	"template_name": {
+		Type:     schema.TypeString,
+		Optional: true,
+	},
+	"template_project_id": {
+		Type:     schema.TypeInt,
+		Optional: true,
+	},
```
armsnyder, line 218: You can enforce these to be mutually exclusive using `ConflictsWith` ([source](https://github.com/hashicorp/terraform-plugin-sdk/blob/35d7fcb6c861606651dee1584261138dd71fb61b/helper/schema/schema.go#L196)).

```suggestion
	"template_name": {
		Type:          schema.TypeString,
		Optional:      true,
		ConflictsWith: []string{"template_project_id"},
	},
	"template_project_id": {
		Type:          schema.TypeInt,
		Optional:      true,
		ConflictsWith: []string{"template_name"},
	},
```

Then you can remove your `if options.TemplateName != nil && options.TemplateProjectID != nil` check in the create function.

gitlab/resource_gitlab_project.go
```diff
@@ -208,6 +208,22 @@ var resourceGitLabProjectSchema = map[string]*schema.Schema{
 		Type:     schema.TypeBool,
 		Optional: true,
 	},
+	"template_name": {
```
armsnyder, line 211: For attributes that cannot be updated after resource creation, like these, we use the `ForceNew: true` field ([source](https://github.com/hashicorp/terraform-plugin-sdk/blob/35d7fcb6c861606651dee1584261138dd71fb61b/helper/schema/schema.go#L140)). Could you add that to these new attributes?

gitlab/resource_gitlab_project_test.go
```diff
@@ -724,6 +857,58 @@ resource "gitlab_project" "foo" {
...
+	`, rInt, rInt)
+}
+
+func testAccGitlabProjectConfigTemplateNameCustom(rInt int) string {
+	return fmt.Sprintf(`
+resource "gitlab_project" "template-name-custom" {
+  name = "template-name-custom-%d"
+  path = "template-name-custom.%d"
+  description = "Terraform acceptance tests"
+  template_name = "myrails"
```
armsnyder, line 878: Can you add a comment in this test, either here or in the test above, explaining where this template is created?

gitlab/resource_gitlab_project_test.go
```diff
@@ -206,6 +206,129 @@ func TestAccGitlabProject_initializeWithReadme(t *testing.T) {
 	})
 }
 
+func TestAccGitlabProject_templateName(t *testing.T) {
```
armsnyder, line 209: Can you add these tests as additional TestSteps in the `basic` test above, instead of making them separate tests? That's how Terraform recommends doing it. That way we can reduce some of the testing overhead, and also test what happens when the user updates these attributes. Once you implement the `ForceNew: true` change I suggested in another comment, Terraform will handle destroying and creating the test project for you.

gitlab/resource_gitlab_project_test.go
```diff
@@ -206,6 +206,129 @@ func TestAccGitlabProject_initializeWithReadme(t *testing.T) {
...
+					testAccCheckGitlabProjectExists("gitlab_project.template-name", &project),
+					testAccCheckGitlabProjectDefaultBranch(&project, &testAccGitlabProjectExpectedAttributes{
+						DefaultBranch: "master",
+					}),
+					func(state *terraform.State) error {
+						client := testAccProvider.Meta().(*gitlab.Client)
+
+						projectID := state.RootModule().Resources["gitlab_project.template-name"].Primary.ID
+
+						_, _, err := client.RepositoryFiles.GetFile(projectID, "README.md", &gitlab.GetFileOptions{Ref: gitlab.String("master")}, nil)
```
armsnyder, line 230: Can you check for a file that is unique to your template, other than `README.md`? This test as written would also succeed for configs with `initialize_with_readme=true`.

gitlab/resource_gitlab_project_test.go
```diff
@@ -206,6 +206,129 @@ func TestAccGitlabProject_initializeWithReadme(t *testing.T) {
...
+}
+
+func TestAccGitlabProject_templateMutualExclusiveNameAndID(t *testing.T) {
+	var project gitlab.Project
+	rInt := acctest.RandInt()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
```
armsnyder, line 320: Usually for testing errors you would use `ExpectError` ([source](https://github.com/hashicorp/terraform-plugin-sdk/blob/35d7fcb6c861606651dee1584261138dd71fb61b/helper/resource/testing.go#L410)) ([example](https://github.com/gitlabhq/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_project_variable_test.go#L343)). Actually as this is written I think it would fail, because you are not setting the `ExpectError` field.

armsnyder (changes requested):

armsnyder:
Also, just as a reminder @sfang97, we should run this PR from a branch of this repository, so that the EE tests can run, before we merge. I made the mistake of forgetting to do that recently and it was a pain to fix haha.

ahmet2mir:
@armsnyder should be good now. Let me know if I forget something.

armsnyder (approved):
lgtm!

@roidelapluie What's your typical workflow for running EE tests?

nicholasklick:
@ahmet2mir looks like one conflicting file and then we can get this merged.

ahmet2mir:
@nicholasklick merged master to this branch

ahmet2mir:
@armsnyder @sfang97 @roidelapluie I remerged and pushed forced changes due to lot of conflicts with master. Could we merge this or need I rebase everyday :)

Issue #263, opened 2020-02-25
Gitlab introduced soft delete of projects and groups


ringods:
### Affected Resource(s)
Please list the resources as a list, for example:
- `gitlab_group`
- `gitlab_project`
- the acceptance tests

### Expected Behavior
Removing the block for a `gitlab_group` or `gitlab_project` should delete the resource.

### Actual Behavior
Depending on the subscription tier you are on, the deletion of the resource is either hard or soft version.

A delete is a soft-delete when you are working against:
* Gitlab.com: Silver or Gold
* Self-managed: Premium or Ultimate

As a user, when I delete a `gitlab_project` resource in my code and `tf apply`, the project becomes soft-deleted. If I re-add again within the "pending deletion" period and `tf apply`, the creation will fail because the project is actually still there.

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- GH-259
- [Prevent accidental deletions via soft delete for groups](https://gitlab.com/gitlab-org/gitlab/issues/33257)
- [Preventing accidental project deletion via soft delete](https://gitlab.com/gitlab-org/gitlab/issues/32935)
- [Default deletion adjourned period](https://docs.gitlab.com/ee/user/admin_area/settings/visibility_and_access_controls.html#default-deletion-adjourned-period-premium-only)
- https://travis-ci.com/pulumi/pulumi-gitlab/builds/142654401#L1345

nagyv:
I'd recommend setting the default deletion adjourned period in this case to immediate. 

The basic idea of Terraform is that every infrastructure management happens through Terraform, thus soft-delete should be either managed by Terraform or it should be switched off on the infrastructure side. As Terraform does not support soft-delete, the only option is to switch it off.

ringods:
@nagyv as a Gitlab.com user, I can't turn this off when on higher pricing tiers, can I?

YusDyr:
> I'd recommend setting the default deletion adjourned period in this case to immediate.
> 
It's just a ridiculous. Unless we could set up soft/hard delete on project/group level, you can't just propose to turn it off for a whole gitlab. Absolutetly clear there are many teams which work in one gitlab but different projects/groups and not all managed by terraform.

What I'm proposing to do is to add code checking "marked_for_deletion" as this is the only attribute which I seem to be returned in GET /project after delete it. It should be pretty simple, doesn't it?

kuahyeow:
[...]

PR #291, opened 2020-04-01
Support for Gitlab EE soft-delete of groups and projects

ringods:
Adding the support in the `gitlab_group` and `gitlab_project` resources for soft delete as defined in Gitlab Enterprise Edition since version 12.

Fixes: #263 

@weimeng and/or @kuahyeow I still could use some help with the Gitlab Enterprise Edition test suite. On Github Actions, I can't get them to run. First of all, they time out which means that the tests are hanging somewhere. I also have several failures.

Comparing this to the Gitlab CE test suite. These are the same tests, but they all succeed when running against the Gitlab CE edition. Also, when I run the test suite against Gitlab EE on my local mac in a Vagrant box, they run in about 11 minutes and I have 1 test failure which I think is related to #269.

Would one of you be open for a conference call where we could discuss things? With me knowing how TF tests work and you knowing how Gitlab and the API work, we have to bring this knowledge together to get this working.


ringods:
@weimeng and @kuahyeow, I have further analysed the logs from an acceptance test run with DEBUG output showing the requests and respones going to and coming from a Gitlab server. I detected an inconsistency between `groups` and `projects`:

* `groups`: soft-delete can be detected by field `marked_for_deletion_on`
* `projects`: soft-delete can be detected by field `marked_for_deletion_at`

Note the difference in the last two characters of the field name.

Additional question: is there even a `pending_delete` field in the `groups` or `projects` JSON returned by a GET operation?

kuahyeow:
> @weimeng and @kuahyeow, I have further analysed the logs from an acceptance test run with DEBUG output showing the requests and respones going to and coming from a Gitlab server. I detected an inconsistency between `groups` and `projects`:
> 
> * `groups`: soft-delete can be detected by field `marked_for_deletion_on`
> * `projects`: soft-delete can be detected by field `marked_for_deletion_at`
> 
> Note the difference in the last two characters of the field name.

Yeah, I have noticed this and opened https://gitlab.com/gitlab-org/gitlab/-/issues/212937

> Additional question: is there even a `pending_delete` field in the `groups` or `projects` JSON returned by a GET operation?

I don't think there's a `pending_delete` at all. It's `marked_for_deletion_at` / `marked_for_deletion_on` for projects and groups respectively as you have noticed

PR #297, opened 2020-04-12
Provide a way to specify client cert and key

dabio:
This will allows the option to specify a client certificate and key
when your GitLab instance runs behind a proxy that requires both.

PR #315, opened 2020-05-23
Provide a way to specify client cert and key

dabio:
This will allows the option to specify a client certificate and key
when your GitLab instance runs behind a proxy that requires both.


dabio:
This fixes #297.

PR #316, opened 2020-05-27
Skip tests when running against CE since only EE supports LDAP links.

jasoncarlsen:


ringods:
@jasoncarlsen tnx a lot for your swift action.

PR #324, opened 2020-06-03
Handling environmentScope as a key for project variables

### LINKED ISSUE #213: Incorrect project environment variable lookup
### Terraform Version
```
$ terraform -v
Terraform v0.12.13
+ provider.vault v2.6.0
```

### Affected Resource(s)
- gitlab_project_variable

```hcl
resource "gitlab_project_variable" "vault-role-id" {
  key               = "VAULT_ROLE_ID"
  value             = vault_approle_auth_backend_role.app-role.role_id
  project           = var.project.gitlab_project.id
  environment_scope = "development"
  protected         = var.gitlab_protected_environment
}

resource "gitlab_project_variable" "vault-role-id" {
  key               = "VAULT_ROLE_ID"
  value             = vault_approle_auth_backend_role.app-role.role_id
  project           = var.project.gitlab_project.id
  environment_scope = "production"
  protected         = var.gitlab_protected_environment
}
```

### Description
The GitLab project environment variable lookup is incorrect. In GitLab an environment variable can have the same name in multiple environment scopes. However, the GitLab api doesn't support this on the REST api that is used in the source code of this provider. So what happens is that the wrong environment variable is being read when refreshing the state.

Solving this would be to fetch the list of all project variable and find the variable with the corresponding name and environment_scope.
```
/api/v4/projects/<project_id>/variables
```

### LINKED ISSUE #235: Environment scope needs to be a part of the gitlab_project_variable ID
Hi,
I have two different terraform projects that create a gitlab_project_variable with the same name but with different environment_scope. Everything works when I run terraform apply and the variables are created correctly. But if I run terraform apply again for the first project it what's to update the variable being created from the second project.

I have found the problem being that the resource ID is based on `projectID:variableName` but since you can have multiple instance of the same variable name inside a gitlab project this needs to change to perhaps `projectID:scopeName:variableName` 

### Terraform Version
Terraform v0.12.17

### Affected Resource(s)
- gitlab_project_variable

### LINKED ISSUE #298: gitlab_project_variable uses a non unique identifier
### Terraform Version
Terraform v0.12.24
...
+ provider.gitlab v2.6.0
...
### Affected Resource(s)
- gitlab_project_variable

### Expected Behavior
What should have happened?
I have multiple time
```
resource "gitlab_project_variable" "AWS_ACCESS_KEY" {
  project           = data.gitlab_project.project.id
  key               = "AWS_ACCESS_KEY_ID"
  value             = var.AWS_KEY
  masked            = true
  environment_scope = terraform.workspace
}
```
Should only sync that variable but instead a synced a different one with the same key.
```
   # module.sai_gitlab_aws.gitlab_project_variable.AWS_ACCESS_KEY will be updated in-place
   ~ resource "gitlab_project_variable" "AWS_ACCESS_KEY" {
       ~ environment_scope = "production*" -> "development"
         id                = "10:AWS_ACCESS_KEY_ID"
         key               = "AWS_ACCESS_KEY_ID"
       ~ masked            = false -> true
         project           = "10"
         protected         = false
       ~ value             = (sensitive value)
         variable_type     = "env_var"
     }
```
### Actual Behavior
It should use a different identifier to map gitlab_project_variable as key is not unique

techthumb:
Gitlab introduced the `environmentScope` attribute to ProjectVariables.

Within Gitlab, ProjectVariables are currently keyed by `key`.

Since the introduction of `environmentScope`, ProjectVariables should be keyed by `key+environmentScope`.

This PR now allows for the provider to be keyed by `key+environmentScope`.

The current implementation is a workaround for a bug within Gitlab API.
https://gitlab.com/gitlab-org/gitlab/-/issues/20661

Brute-force approach to workaround the shortcoming of the Gitlab API issue:

1. Fetch all ProjectVariables
2. Filter ProjectVariables for the key that is being updated or deleted
3. Issue n Delete api calls for ProjectVariable key where n is the number of ProjectVariables matching the key
4. Re-create without change ProjectVariables for key & environmentScope that are not being updated/deleted
5. If Update, re-create with change ProjectVariable for key & environmentScope that is being updated
5. If Delete, no API call for the ProjectVariable for key & environmentScope that is should be deleted

Fixes #213
Fixes #298 
Fixes #235

### DIFF
--- a/gitlab/resource_gitlab_project_variable.go
+++ b/gitlab/resource_gitlab_project_variable.go
@@ -13,6 +13,7 @@ func resourceGitlabProjectVariable() *schema.Resource {
 		Read:   resourceGitlabProjectVariableRead,
 		Update: resourceGitlabProjectVariableUpdate,
 		Delete: resourceGitlabProjectVariableDelete,
+		Exists: resourceGitlabProjectVariableExists,
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
@@ -59,6 +60,35 @@ func resourceGitlabProjectVariable() *schema.Resource {
 	}
 }
 
+func resourceGitlabProjectVariableExists(d *schema.ResourceData, meta interface{}) (bool, error) {
+	client := meta.(*gitlab.Client)
+
+	project := d.Get("project")
+
+	log.Printf("[DEBUG] list gitlab project variable %s", project)
+
+	options := gitlab.ListProjectVariablesOptions{Page: 1, PerPage: 9999}
+	projectVariables, _, err := client.ProjectVariables.ListVariables(project, &options)
+	if err != nil {
+		return false, err
+	}
+
+	log.Printf("[DEBUG] gitlab project variables: %s", projectVariables)
+
+	key := d.Get("key")
+	environmentScope := d.Get("environment_scope")
+
+	for _, projectVariable := range projectVariables {
+		if projectVariable.Key == key && projectVariable.EnvironmentScope == environmentScope {
+			log.Printf("[DEBUG] Variable matching key and environment scope exists: %s:%s", key, environmentScope)
+			return true, nil
+		}
+	}
+
+	log.Printf("[DEBUG] Variable matching key and environment scope does NOT exist: %s:%s", key, environmentScope)
+	return false, nil
+}
+
 func resourceGitlabProjectVariableCreate(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*gitlab.Client)
 
@@ -78,14 +108,14 @@ func resourceGitlabProjectVariableCreate(d *schema.ResourceData, meta interface{
 		Masked:           &masked,
 		EnvironmentScope: &environmentScope,
 	}
-	log.Printf("[DEBUG] create gitlab project variable %s/%s", project, key)
+	log.Printf("[DEBUG] create gitlab project variable %s/%s/%s", project, key, environmentScope)
 
 	_, _, err := client.ProjectVariables.CreateVariable(project, &options)
 	if err != nil {
 		return err
 	}
 
-	d.SetId(buildTwoPartID(&project, &key))
+	d.SetId(buildThreePartID(&project, &key, &environmentScope))
 
 	return resourceGitlabProjectVariableRead(d, meta)
 }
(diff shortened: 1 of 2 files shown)


techthumb:
Fixes #213

gitlab/resource_gitlab_project_variable.go
```diff
@@ -78,80 +108,162 @@ func resourceGitlabProjectVariableCreate(d *schema.ResourceData, meta interface{
...
+			return err
+		}
+	}
+	return nil
+}
+
+func deleteProjectVariables(projectVariables []*gitlab.ProjectVariable, project interface{}, client *gitlab.Client) error {
+	for _, projectVariable := range projectVariables {
+		log.Printf("[DEBUG] Delete gitlab project variable %s/%s", project, projectVariable.Key)
+		_, err := client.ProjectVariables.RemoveVariable(project, projectVariable.Key)
```
woz5999, line 246: the gitlab api still has no way to deterministically target a variable for modification or deletion, even if you're able to do an accurate lookup. i don't believe the EnvironmentScope option in this context is used by the API to identify a variable. Rather, the API seems to use the variable Key exclusively and the additional options are just configuring the variable's properties. 

My belief is that if this call will just delete a variable with the specified key but and a random environment scope.

techthumb: We are aligned in our understanding of the Gitlab API behavior.

Hence, the code applies a brute force approach here.

1. Fetch all ProjectVariables
2. Filter ProjectVariables for the key that is being updated or deleted
3. Issue n Delete api calls for ProjectVariable key
4. Re-create without change ProjectVariables for key & environmentScope that are not being updated/deleted
5. If Update, re-create with change ProjectVariable for key & environmentScope that is being updated
5. If Delete, no API call for the ProjectVariable for key & environmentScope that is should be deleted

woz5999: Clever workaround. Might be dicey touching variables potentially not managed by terraform or managed by a different project. But as a workaround I agree it's better than nothing at this point.

techthumb:
Signed-off-by: Kunal Parikh Kunal.parikh@nasdaq.com

techthumb:
@roidelapluie Could you please provide feedback to get this merged into master?

techthumb:
@ringods Could you please provide feedback to get this merged into master?

Nosmoht:
Can we please merge this ASAP as the current behaviour destroys most of my pipeline?

Nosmoht:
Hi all,

i've just tested the branch with some existing Gitlab project environment scoped variables and got following error while `terraform refresh`:

```shell
...
Error: Unexpected ID format ("781:GOOGLE_APPLICATION_CREDENTIALS"). Expected project:key:env
...
```

Is there anything planed on how to transform existing variables with old structure to new one?

techthumb:
> Is there anything planed on how to transform existing variables with old structure to new one?

I didn't code for this. 
I did recently learn about `StateUpgrader` & will consider using it here.

> Should we?
> * Create new variables with ID format project:key:env
> * Update existing variable with format project:key to project:key:env with env default * All (Default)
> * Use env * All (Default) as the default

I don't know the correct answer to 👆, because there is a matrix of non-trivial cases to consider here:

1.  One-to-one mapping between `project:key` and `env`
2. One-to-many mapping between `project:key` and `env`
3. Is a user requested new change being applied?

Based on my (limited) understanding of the TF plugin api, and the missing information in the state file, I suspect that we can't answer these questions to automate a fix.

@Nosmoht Can you validate my thinking here?
There is a good chance that I may be missing some concepts that may provide an automated solution.

Nosmoht:
Fair question, which i'm unable to answer. Can we get someone from Gitlab into this?

woz5999:
@Nosmoht here's the relevant gitlab issue. it looks like they're about to release an api update shortly: https://gitlab.com/gitlab-org/gitlab/-/issues/9912

I imagine once that's live the provider will need to be updated to support the new changes.

enkov:
Any news here? Gitlab released a fix for environment variables.

thapakazi:
If we are good, lets merge this, we couldn't update our CI variables because of this 😞

woz5999:
This probably shouldn't get merged unless and it'll it's checked against the gitlab api changes released recently to address the underlying api issue. people in #213 are saying this resource is currently broke entirely for them.

techthumb:
Hi @nicholasklick, 

Thanks for looking into this.

As mentioned in the issue description, this implementation applies a brute force approach.

Given that the api now supports filtering by environment scope, this implementation is perhaps invalid.

nicholasklick:
@techthumb good to know. @sfang97 will be investigating this further and help moving this toward merge.

sfang97:
@techthumb Thanks for your work on this, a lot of people are looking forward to this being merged! 😄 
Since GitLab has fixed the bug in the API, would you be willing to modify this PR to use the fix? The API implementation is not in `go-gitlab` yet,  but perhaps @remil1000's solution here: https://github.com/xanzy/go-gitlab/issues/746#issuecomment-671364797 could be used to simplify this PR? 
Environment scope bug fix MR: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/34490

techthumb:
The scope of this issue is getting large:

1. migrate state file
2. detect feature flag
3. implementation based on the value of the feature flag
4. testing with feature flag on and off

@sfang97 I don't have the bandwidth ATM to work on this.
I am happy to review and manually test changes though.

armsnyder:
If you can it would be good to include a new test that creates two different variables using the same key but different scopes. That would verify the new behavior, and the bug reported in #396 in particular.

nicholasklick:
@armsnyder @sfang97  how can we work to move this through?

1. It seems like the original committer doesnt have time to complete updating the PR
1. Can we better leverage `go-gitlab` as mentioned here to simplify this PR?

armsnyder:
I agree with @sfang97 that the current hacky solution of iterating over all variables is good enough. The thing that's missing for me is validating that the provider version upgrade is safe. Changing a resource's ID format is listed as a [potentially breaking change](https://www.terraform.io/docs/extend/best-practices/versioning.html#example-major-number-increments) by Terraform. It isn't clear to me from Terraform's docs if implementing [state migration](https://www.terraform.io/docs/extend/resources/state-migration.html) makes it non-breaking, or if Terraform still wants it to be a major version upgrade. Does anyone know?

roidelapluie:
Anyway major version upgrade is not an issue. We can release a 3.0.0 to be safe.

nicholasklick:
@armsnyder @roidelapluie @sfang97  I suggest that we merge this and bump to a major version update. 
This seems to be blocking a lot of people. 
We can [iterate](https://about.gitlab.com/handbook/values/#iteration) and improve this going forward.

nicholasklick:
Seems like there are conflicts now.

armsnyder:
I can take a stab at a new PR, unless @sfang97 or someone else has already started on one.

sfang97:
@armsnyder Go for it! Thanks for working on this :)

abelmokadem:
Is it possible to get the GitLab provider from this PR and run it in my own project?

armsnyder:
@abelmokadem It is possible. You can clone this project and build the provider yourself with `make build`, then [install the binary as a third party Terrraform plugin](https://www.terraform.io/docs/plugins/basics.html#installing-plugins). Beware that there is the potential for a breaking change for you when you switch back to an official release of this provider, after we get a fix in and released.

armsnyder:
Closing this since we merged #409

PR #330, opened 2020-06-09
fix: set project_key as required in gitlab_service_jira

solidnerd:
This will set the project key also required according to the current docs.

### DIFF
--- a/gitlab/resource_gitlab_service_jira.go
+++ b/gitlab/resource_gitlab_service_jira.go
@@ -47,8 +47,7 @@ func resourceGitlabServiceJira() *schema.Resource {
 			},
 			"project_key": {
 				Type:     schema.TypeString,
-				Optional: true,
-				Default:  "",
+				Required: true,
 			},
 			"username": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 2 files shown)


ringods:
Hello @solidnerd, can you also fix the acceptance tests accordingly?

sysadmiral:
Hello,

Something that threw me about this being "required" is that in the Gitlab UI for the Jira Integration there is no reference to the project key at all:

image

There is a mismatch between the fields in the UI and the ones used by the terraform provider that could lead to confusion at import time.

(API Url is another example field - present in UI but not mentioned in provider)

Is this a case of documentation needing updating or perhaps the underlying go-gitlab library needs updating to reflect newer gitlab versions?

I would be happy to help if you feel there are some actions that need to happen based on my comment.

Many Thanks

timofurrer:
@armsnyder How do you suggest to move forward with this one here?

armsnyder:
Let's close this PR for now. There is risk of it being a breaking change, plus it does not add functionality and there is no issue. We can re-open or move discussion to an issue if it is needed.

PR #337, opened 2020-06-16
Clean the state for jira if project doesn't exist

### LINKED ISSUE #336: gitlab_service_jira must clean the state if it can't read project
### Terraform Version
```
Terraform v0.12.26
+ provider.gitlab v2.10.0
```

### Affected Resource(s)
- gitlab_service_jira

```hcl
provider "gitlab" {
  base_url = "http://localhost:8080/api/v4/"
  version  = "~> 2.10.0"
}

resource "gitlab_group" "example" {
  name        = "example"
  path        = "example"
  description = "An example group"
}
  name         = "example"
  description  = "An example project"
  namespace_id = gitlab_group.example.id
}
data "gitlab_projects" "example" {
  order_by = "id"
  group_id = gitlab_group.example.id
  simple   = true
}
resource "gitlab_service_jira" "example" {
  for_each                 = { for project in data.gitlab_projects.example.projects : project.name => project.id }
  project                  = each.value
  url                      = "https://example.atlassian.net"
  project_key              = "EXAMPLE"
  username                 = "example"
  password                 = "example"
  comment_on_event_enabled = false
  jira_issue_transition_id = 12
}
```

### Expected Behavior
If project doesn't exist anymore I would expect `gitlab_service_jira` to be dropped from state as well.

### Actual Behavior
```
Error: GET http://localhost:8080/api/v4/projects/38/services/jira: 404 {message: 404 Project Not Found}
```

### Steps to Reproduce
I can imagine that this is our very specific use case and we end up here due to multiple migrations.
Half of the state was actually imported as we started to grow and maintaining `gitlab` manually became too painful.
But we are here and these are the steps:
1. `terraform apply -target data.gitlab_projects.example -target gitlab_group.example -target gitlab_project.example`
2. `terraform apply`
3. `terraform destroy -target gitlab_project.example`
4. `terraform apply`

* of course we don't run so many steps, as in our case we always `apply`, all these are needed to bootstrap/reproduce the state
* the idea is that we link resources using `data source` instead of reference to a specific `resource` which hides dependencies under data source abstraction.

### Important Factoids
none

### References
Will update afterwards.

ymatsiuk:
Fixes #336

### DIFF
--- a/gitlab/resource_gitlab_service_jira.go
+++ b/gitlab/resource_gitlab_service_jira.go
@@ -3,6 +3,7 @@ package gitlab
 import (
 	"fmt"
 	"log"
+	"net/http"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	gitlab "github.com/xanzy/go-gitlab"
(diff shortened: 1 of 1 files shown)


armsnyder:

armsnyder:
Thanks!

PR #339, opened 2020-06-19
Import group labels

aquavitae:
This simply adds support for importing group labels, e.g.:

```sh
$ terraform import gitlab_deploy_key_enable.example 12345:fixme
```

### DIFF
--- a/gitlab/resource_gitlab_group_label.go
+++ b/gitlab/resource_gitlab_group_label.go
@@ -1,7 +1,9 @@
 package gitlab
 
 import (
+	"fmt"
 	"log"
+	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	gitlab "github.com/xanzy/go-gitlab"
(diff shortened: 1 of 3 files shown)


gitlab/resource_gitlab_group_label.go
```diff
@@ -123,3 +129,25 @@ func resourceGitlabGroupLabelDelete(d *schema.ResourceData, meta interface{}) er
...
+	group, _, err := client.Groups.GetGroup(parts[0])
+	if err != nil {
+		return nil, err
+	}
+
+	if err := d.Set("group", strconv.Itoa(group.ID)); err != nil {
+		return nil, err
+	}
+
+	err = resourceGitlabGroupLabelRead(d, meta)
```
armsnyder, line 150: Is it necessary to read here? Doesn't Terraform perform a read for you before the next operation anyway? I don't think it changes the provider behavior.

armsnyder: Never mind. I see it is a recommended practice in Terraform docs. Carry on!

aquavitae: Possibly... I was modelling this on how all the other Importers were implemented?

roidelapluie: Yes I think it makes sense to read and make sure the resource you try to import exists.

PR #343, opened 2020-06-22
Fix Terraform trying to set `default_branch` to `null`

### LINKED ISSUE #299: `default_branch` is reset to null when not set in .tf
### Terraform Version
```
$ terraform -v
Terraform v0.12.24
+ provider.gitlab v2.5.0
```

### Affected Resource(s)
- gitlab_project

```hcl
provider "gitlab" {
}

resource "gitlab_project" "default" {
  name             = "dnf-go-gui"
  description      = "DNF What's New for upcoming package updates."
  visibility_level = "public"
}
```

### Expected Behavior
Expected [`default_branch` argument](https://www.terraform.io/docs/providers/gitlab/r/project.html#default_branch) to be optional. Running `terraform plan` should not reset `default_branch` for the project if it is not set in `.tf` file.

### Actual Behavior
```
...
        container_registry_enabled                       = true
      - default_branch                                   = "master" -> null
        description                                      = "DNF What's New for upcoming package updates."
...
```

### Steps to Reproduce
1. `terraform plan`

### References

https://github.com/terraform-providers/terraform-provider-gitlab/blob/a9f493f125f95ca4432c80e65f73faf36d326a83/gitlab/resource_gitlab_project.go#L38

abitrolly:
Fixes #299 where Terraform tries to reset the `default_branch` to null for non-empty repositories if not set in .tf

The `default_branch` defaults to `master` on GitLab side, but it is only set if the repository is not empty. Otherwise it is `null` on remote side.

---

The debug in 24d43d8fb30e2e4038dcbac545db5ec10f9766ad showed no trace of `null` value from remote.

    2020/06/19 20:55:55 [DEBUG] gitlab.Project default_branch (string) ""
    2020/06/19 20:55:55 [DEBUG] d.Get default_branch (string) ""
    2020/06/19 20:55:55 [DEBUG] d.GetOk default_branch (string) "" - false
    2020/06/19 20:55:55 [DEBUG] d.HasChange default_branch false
    2020/06/19 20:55:55 [DEBUG] d.GetChange default_branch (string) "", (string) ""

But the actual response shows that `null` was returned in JSON.

    2020/06/19 20:55:55 [DEBUG] GitLab API Response Details:
    ...
     "default_branch": null,

---

The fix is to skip the change if TF is trying to set null value for remote side. This PR also adds a warning if `default_branch` is specified in .tf file, but the remote repo is empty. And also removes more specific check for `initilize_with_readme` towards the generic case above. `default_branch` should not make assumptions on how the repo was created.


abitrolly:
Setting default branch on an empty repo results in an error.
```
Error: PUT https://gitlab.com/api/v4/projects/19526446: 400 {message: {base: [Could not change HEAD: branch 'sdfdf' does not exist]}}
```

armsnyder:
It would be nice to see a [regression test](https://www.terraform.io/docs/extend/best-practices/testing.html#regression-tests) for this, since it's difficult to tell just by looking at the change how it will behave.

horjulf:
This looks simple enough, can we get this merged please ?
Currently the provider breaks in all cases where you don't want to track or enforce a default branch in Terraform code.

@nicholasklick

abitrolly:
With the ongoing governmental terror in Belarus, I am emotionally drained, and can not think clearly to write the tests, sorry.

roidelapluie:
@armsnyder WDYT about this?

@abitrolly no worries. take care!!!!

armsnyder:
@roidelapluie The PR looks like an improvement, so I'm fine with merging. I think it could be improved further at a later time by using the `Default` or `Computed` fields of the attribute, and I was also hoping for a test. But on visual inspection it looks ok.

nicholasklick:
Merging due to customer request and Maintainer approval

roidelapluie:
> Merging due to customer request and Maintainer approval

Thanks to* I hope :)

Issue #136, opened 2019-05-23
Release a 0.12 compatible version


valdisrigdon:
Trying to upgrade to Terraform 0.12, but this provider doesn't have a compatible version.

```
No available provider "vault" plugins are compatible with this Terraform version.

From time to time, new Terraform major releases can change the requirements for
plugins such that older plugins become incompatible.

Terraform checked all of the plugin versions matching the given constraint:
    (any version)

Unfortunately, none of the suitable versions are compatible with this version
of Terraform. If you have recently upgraded Terraform, it may be necessary to
move to a newer major release of this provider. Alternatively, if you are
attempting to upgrade the provider to a new major version you may need to
also upgrade Terraform to support the new version.

Consult the documentation for this provider for more information on
compatibility between provider versions and Terraform versions.


Error: no available version is compatible with this version of Terraform
```

roidelapluie:
Hello, when #135 gets merged we will do a release!

roidelapluie:
2.0.0 is out!

PR #344, opened 2020-06-23
add support for importing project push rules

aquavitae:
Adds support for importing project push rules. Since there is a 1-to-1 relationship with the project, all that is needed is the project name or id.


ziporah:
This MR is valid, I believe the tests are failing for non-code related issue, can someone please fix this, so we can at least import the rules for gitlab cloud repo's

armsnyder:
I apologize, but I just realized that my recent PR #360 is a duplicate of this one. I didn't see this before because it wasn't mentioned in any of the reported issues. I'm going to keep mine open since it includes a test.

armsnyder:
I think there's a bug in this implementation -- since you don't set the ID in the import function, and the ID also isn't set in the read function, then that means that created vs imported resources will have different ID types.

roidelapluie:
I have merged #136 
Thanks to you both!

PR #346, opened 2020-06-23
Updated environment_scope reference

rpfinneran:
Leaving environment_scope off (optional) results in environment value of 0, it should rather be set to asterisk. If the scope is in GitLab as 0, it will not be accessible to runners in any environment (except 0 of course, which would be a very unusual name for an environment).


armsnyder:
Closing this since we fixed the default for the upcoming 3.0.0 release via #409

Issue #205, opened 2019-10-27
Destroy acceptance tests not tested properly + duplicated code


kirecek:
Hello, I noticed that many destroy tests do checks only for the absence of a parent resource (i.e. gitlab_project and gitlab_group) and delete function for a particular resource is not tested at all (they are only called by terraform).

What's worse, most of these destroy tests are duplicated code. For example, the following tests are the same.

- [x] [testAccCheckGitlabPipelineScheduleDestroy](https://github.com/terraform-providers/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_pipeline_schedule_test.go#L130)
- [x] [testAccCheckGitlabPipelineTriggerDestroy](https://github.com/terraform-providers/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_pipeline_trigger_test.go#L99)
- [x] [testAccCheckGitlabProjectPushRulesDestroy](https://github.com/terraform-providers/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_project_push_rules_test.go#L165)
- [x] [testAccCheckGitlabProjectHookDestroy](https://github.com/terraform-providers/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_project_hook_test.go#L155)
- [x] [testAccCheckGitlabServiceJiraDestroy](https://github.com/terraform-providers/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_service_jira_test.go#L106)
- [x] [testAccCheckGitlabServiceSlackDestroy](https://github.com/terraform-providers/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_service_slack_test.go#L126)
- [x] [testAccCheckGitlabLabelDestroy](https://github.com/terraform-providers/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_label_test.go#L114)
- [x] [testAccCheckGitlabProjectVariableDestroy](https://github.com/terraform-providers/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_project_variable_test.go#L111)

I think it deserves some refactoring. Does it make sense to take some action here?

armsnyder:
Still a weak point of our tests

PR #358, opened 2020-07-14
feature: add project mirror resources to provider

lukemgriffith:
Implemented functionality for project mirrors, as per issue https://github.com/terraform-providers/terraform-provider-gitlab/issues/150

This was recently added to the go-gitlab library by myself and released in v0.33.0

https://github.com/xanzy/go-gitlab/pull/879

### DIFF
--- a/gitlab/resource_gitlab_project_mirror.go
+++ b/gitlab/resource_gitlab_project_mirror.go
@@ -1,0 +1,176 @@
+package gitlab
+
+import (
+	"log"
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/xanzy/go-gitlab"
+)
+
+func resourceGitlabProjectMirror() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGitlabMirrorCreate,
+		Read:   resourceGitlabProjectMirrorRead,
+		Update: resourceGitlabMirrorUpdate,
+		Delete: resourceGitlabMirrorDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"project": {
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+			},
+			"mirror_id": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+			"url": {
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+			},
+			"enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+			"only_protected_branches": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+			"keep_divergent_refs": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+		},
+	}
+}
+
+func resourceGitlabMirrorCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gitlab.Client)
+
+	projectID := d.Get("project").(string)
+	URL := d.Get("url").(string)
+	enabled := d.Get("enabled").(bool)
+	onlyProtectedBranches := d.Get("only_protected_branches").(bool)
+	keepDivergentRefs := d.Get("keep_divergent_refs").(bool)
+
+	options := &gitlab.AddProjectMirrorOptions{
+		URL:                   &URL,
+		Enabled:               &enabled,
+		OnlyProtectedBranches: &onlyProtectedBranches,
+		KeepDivergentRefs:     &keepDivergentRefs,
+	}
+
+	log.Printf("[DEBUG] create gitlab project mirror for project %v", projectID)
+
+	mirror, _, err := client.ProjectMirrors.AddProjectMirror(projectID, options)
+	if err != nil {
+		return err
+	}
+	d.Set("mirror_id", mirror.ID)
+
+	mirrorID := strconv.Itoa(mirror.ID)
+	d.SetId(buildTwoPartID(&projectID, &mirrorID))
+	return resourceGitlabProjectMirrorRead(d, meta)
+}
+
+func resourceGitlabMirrorUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gitlab.Client)
+
+	mirrorID := d.Get("mirror_id").(int)
+	projectID := d.Get("project").(string)
+	enabled := d.Get("enabled").(bool)
+	onlyProtectedBranches := d.Get("only_protected_branches").(bool)
+	keepDivergentRefs := d.Get("keep_divergent_refs").(bool)
+
+	options := gitlab.EditProjectMirrorOptions{
+		Enabled:               &enabled,
+		OnlyProtectedBranches: &onlyProtectedBranches,
+		KeepDivergentRefs:     &keepDivergentRefs,
+	}
+	log.Printf("[DEBUG] update gitlab project mirror %v for %s", mirrorID, projectID)
+
+	_, _, err := client.ProjectMirrors.EditProjectMirror(projectID, mirrorID, &options)
+	if err != nil {
+		return err
+	}
+	return resourceGitlabProjectMirrorRead(d, meta)
+}
+
+// Documented remote mirrors API does not support a delete method, instead mirror is disabled.
+func resourceGitlabMirrorDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gitlab.Client)
+
+	enabled := false
+
+	mirrorID := d.Get("mirror_id").(int)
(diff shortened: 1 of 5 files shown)


gitlab/resource_gitlab_project_mirror.go
```diff
@@ -0,0 +1,170 @@
...
+		Create: resourceGitlabMirrorCreate,
+		Read:   resourceGitlabProjectMirrorRead,
+		Update: resourceGitlabMirrorUpdate,
+		Delete: resourceGitlabMirrorDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"project_id": {
```
armsnyder, line 23: "project" seems to be more common in this provider than "project_id", although I do prefer "project_id" myself. Is that a conscious choice?

gitlab/resource_gitlab_project_mirror.go
```diff
@@ -0,0 +1,170 @@
...
+		log.Printf("[DEBUG] project mirror found %v", m.ID)
+		if m.ID == mirrorID {
+			mirror = m
+			found = true
+			return nil
+		}
+	}
+
+	if !found {
+		return fmt.Errorf("unable to find mirror %v on project %s", mirrorID, projectID)
```
armsnyder, line 156: Instead of returning an error, can you `setId("")`? That seems to be the best practice for handling missing upstream resources.

gitlab/resource_gitlab_project_mirror.go
lukemgriffith: I copied the schema from gitlab project member, so not conscious. Happy to change

gitlab/resource_gitlab_project_mirror.go
```diff
@@ -0,0 +1,173 @@
...
+	var mirror *gitlab.ProjectMirror
+	found := false
+
+	for _, m := range mirrors {
+		log.Printf("[DEBUG] project mirror found %v", m.ID)
+		if m.ID == mirrorID {
+			mirror = m
+			found = true
+			return nil
+		} else {
```
armsnyder, line 152: Looks like a bug. This for loop will never make it past the first iteration.

armsnyder:
Can you add a test for the importer as well? You can copy an _import test from another resource without much change.

lukemgriffith:
Will do, late GMT so will do tomorrow. Thanks for reviewing and patience much appreciated.

lukemgriffith:
Having issues adding the importer at the moment, and quite make sense of the error either. Still working on this, any input would be great - im not too familiar with the import tests, or the workflow it runs. Trying to pin down where the call to gitlab is being made.

armsnyder:
The importer test is fine. The problem occurs after the import, during the read. Your read function is depending on some attributes that you are not setting in your import function.

Check the docs for `ImportStatePassthrough`

https://www.terraform.io/docs/extend/resources/import.html#importer-state-function

> Where possible, prefer using schema.ImportStatePassthrough as the Importer State function. This function requires the Read function to be able to refresh the entire resource with d.Id() ONLY. Sometimes it is possible to adjust the resource Read function to replace d.Get() use with d.Id() if they exactly match or add a function that parses the resource ID into the necessary attributes.

jeremad:
You can now rebase on master, go-gitlab has been updated

lukemgriffith:
I've got the acceptance tests working, testing manually resource imports fine also.

lukemgriffith:
rebased and caused a lot of mess. Resouce is working as expected. Do you require the commit history to be cleared up any further or is this good?

gitlab/resource_gitlab_project_mirror.go
```diff
@@ -0,0 +1,176 @@
...
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/xanzy/go-gitlab"
+)
+
+func resourceGitlabProjectMirror() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGitlabMirrorCreate,
+		Read:   resourceGitlabProjectMirrorRead,
```
armsnyder, line 15: Inconsistent function naming scheme here

gitlab/resource_gitlab_project_mirror.go
```diff
@@ -0,0 +1,176 @@
...
+	return nil
+}
+
+func resourceGitlabProjectMirrorSetToState(d *schema.ResourceData, projectMirror *gitlab.ProjectMirror, projectID *string) {
+
+	mirrorID := strconv.Itoa(projectMirror.ID)
+	d.Set("enabled", projectMirror.Enabled)
+	d.Set("only_protected_branches", projectMirror.OnlyProtectedBranches)
+	d.Set("keep_divergent_refs", projectMirror.KeepDivergentRefs)
+	d.SetId(buildTwoPartID(projectID, &mirrorID))
```
armsnyder, line 175: Nitpick, but this line is unnecessary since in all cases the ID will have already been set before calling this function.

gitlab/resource_gitlab_project_mirror_test.go
```diff
@@ -0,0 +1,213 @@
+package gitlab
+
+// TODO convert to handle project mirror.
```
armsnyder, line 3: I can't tell what this means.

gitlab/resource_gitlab_project_mirror_test.go
```diff
@@ -0,0 +1,213 @@
...
+		CheckDestroy: testAccCheckGitlabProjectMirrorDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGitlabProjectMirrorConfig(rInt),
+			},
+			{
+				ResourceName:            "gitlab_project_mirror.foo",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"enabled", "mirror_id", "project", "url"},
```
armsnyder, line 85: Why was this needed? The import should result in a fully synced state, ideally.

gitlab/resource_gitlab_project_mirror_test.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+type testAccGitlabProjectMirrorExpectedAttributes struct {
+	URL                   string
+	Enabled               bool
+	OnlyProtectedBranches bool
+	KeepDivergentRefs     bool
+}
+
+func testAccCheckGitlabProjectMirrorAttributes(mirror *gitlab.ProjectMirror, want *testAccGitlabProjectMirrorExpectedAttributes) resource.TestCheckFunc {
```
armsnyder, line 133: Check your error message strings. They look copy-pasted.

armsnyder:
If you could squash this MR that could be good, because GitHub looks confused. Maybe it's just me but it's showing changes in the diff that aren't yours.

lukemgriffith:
That occured once I performed `git rebase master` and it interlaced the more recent commits. 
I will tidy this up from master and address the more recent comments. 

Re: import test. Testing manually the resource is imported fine using the terraform cli, the import test does not show any fields. I've copied the test from other resources unsure why this is causing problems as the functionality works.

lukemgriffith:
Import test success, I wasn't setting all properties

gitlab/resource_gitlab_project_mirror_test.go
```diff
@@ -0,0 +1,210 @@
...
+			return fmt.Errorf("got only protected branches %t; want %t", mirror.OnlyProtectedBranches, want.OnlyProtectedBranches)
+		}
+		if mirror.KeepDivergentRefs != want.KeepDivergentRefs {
+			return fmt.Errorf("got keep divergent refs %t; want %t", mirror.KeepDivergentRefs, want.KeepDivergentRefs)
+		}
+		return nil
+	}
+}
+
+func testAccCheckGitlabProjectMirrorDestroy(s *terraform.State) error {
```
armsnyder, line 149: Just an fyi, a "destroy" test should be testing the behavior of your own resource destroy function, i.e. checking that the mirror gets disabled. There's an open issue about this generally #205 so we can come back and do this later.

armsnyder:
Can you add a new documentation page for this new resource?

lukemgriffith:
added docs page.

gitlab/resource_gitlab_project_mirror_test.go
lukemgriffith: sure, I will not make any changes in this PR for this.

armsnyder:
Looks good, but can you add the new doc page to the index (like so https://github.com/gitlabhq/terraform-provider-gitlab/pull/391)

lukemgriffith:
ack i actually searched for an index, didn't realise it was the erb file. should be done.

lukemgriffith:
@armsnyder is there anything I can do to speed this up getting merged? I have some work in the next week that i'd be keen to use this resource for... unsure on the release cadence or the process?

armsnyder:
I don't have merging powers, but @nicholasklick may be able to give an update.

lukemgriffith:
Hi @nicholasklick  any timeframe on this getting merged?

PR #331, opened 2020-06-09
Add support for subgroup_creation_level and project_creation_level

maxenced:
* `make test` passed
* tested on real internal gitlab

This has also been released/ tested on our claranet fork.

PR #257, opened 2020-02-12
[WIP] More group options

justinkillen:

PR #362, opened 2020-07-19
More group options

### LINKED ISSUE #210: Support for project_creation_level on groups
This is a feature request to add the `project_creation_level` parameters to the `gitlab_groups` resource.

### Terraform Version
Terraform v0.12.13
+ provider.gitlab v2.3.0

### Affected Resource(s)
- gitlab_groups

```hcl
resource "gitlab_group" "mygroup" {
  name        = "mygroup"
  path        = "mygroup"
  description = "My group description"
  visibility_level = "internal"
  project_creation_level = "noone"
}
```
In this example, I added the inexistent `project_creation_level`. Gitlab supports these values:
- `noone`
- `maintainer`
- `developer`

justinkillen:
Supersedes #257, #331 
Fixes #210 

Add support for gitlab group fields:

Fields added:

ShareWithGroupLock (bool)
RequireTwoFactorAuth (bool)
TwoFactorGracePeriod (int)
ProjectCreationLevel (string: "noone", "maintainer", "developer")
SubGroupCreationLevel (string: "owner", "maintainer")
AutoDevopsEnabled (bool)
EmailsDisabled (bool)
MentionsDisabled (bool)


justinkillen:
For https://github.com/terraform-providers/terraform-provider-gitlab/issues/210

justinkillen:
I'm new to golang, and it's showing.  Could use some help on setting up the tests correctly.

justinkillen:
@roidelapluie ready for review

roidelapluie:
@armsnyder is ther amy chance you can review this? thanks!

gitlab/resource_gitlab_group.go
```diff
@@ -61,9 +61,51 @@ func resourceGitlabGroup() *schema.Resource {
 			"visibility_level": {
 				Type:         schema.TypeString,
 				Optional:     true,
-				Computed:     true,
+				Default:      "private",
```
armsnyder, line 64: What's the reason for changing the default?

website/docs/r/group.html.markdown
```diff
@@ -48,6 +48,32 @@ enable users to request access to the group.
   Valid values are `private`, `internal`, `public`.
   Groups are created as private by default.
 
+* `share_with_group_lock` - (Optional) Boolean, defaults to false.  Prevent sharing
+a project with another group within this group.
+
+* `project_creation_level` - (Optional), defaults to Maintainer
```
armsnyder, line 54: Missing punctuation after `Maintainer`

```suggestion
* `project_creation_level` - (Optional), defaults to Maintainer.
```

website/docs/r/group.html.markdown
```diff
@@ -48,6 +48,32 @@ enable users to request access to the group.
...
+
+* `auto_devops_enabled` - (Optional) Boolean, defaults to false.  Default to Auto
+DevOps pipeline for all projects within this group.
+
+* `emails_disabled` - (Optional) Boolean, defaults to false.  Disable email notifications
+
+* `mentions_disabled` - (Optional) Boolean, defaults to false.  Disable the capability
+of a group from getting mentioned
+
+* `subgroup_creation_level` - (Optional), defaults to Owner
```
armsnyder, line 67: Missing punctuation after `Owner`

```suggestion
* `subgroup_creation_level` - (Optional), defaults to Owner.
```

armsnyder (changes requested):
Thanks for the high quality contribution and the improvements to the test. I left one question and two docs suggestions.

armsnyder:
> I'm new to golang, and it's showing.

😮

gitlab/resource_gitlab_group.go
justinkillen: The gitlab api doesn't require this field.  When not provided, the api will provide a default, but it varies depending on what version of gitlab you're using.  Adding a default here forces deterministic behavior.

armsnyder: Cool, thanks.

armsnyder (approved):

gitlab/resource_gitlab_group.go
roidelapluie: This is a breaking change, I don't think we need that?

armsnyder: @roidelapluie Do you want me to revert it, or is our next release a major one anyway? The reason I was ok with this is because the provider documentation for this resource already says "private" is the default.

roidelapluie: In general I think we should have as much computed as possible, and fix the doc if it is not true?

armsnyder: I started disliking O+C attributes after reading through [this issue](https://github.com/hashicorp/terraform/issues/21278), if you're interested.

Reverting the change to this attribute and fixing up the docs makes sense. I can do a PR later.

roidelapluie: But you can remove the  O ?

armsnyder: Removing the O would make it no longer an argument, right? We still want it to be an argument that can be set.

roidelapluie: mmh I am not 100% sure. But we can keep the O, we do that elsewhere.

armsnyder: PR for this: #408

PR #363, opened 2020-07-19
Fix unhandled Jira Service error

armsnyder:
I noticed this bug while trying to investigate a [possible flaky acceptance test](https://github.com/terraform-providers/terraform-provider-gitlab/pull/356/checks?check_run_id=834966420).

The bug is that if there is an error while creating or updating the Jira Service settings, the error is ignored.


roidelapluie:
Thanks!

PR #364, opened 2020-07-19
Fix nil pointer dereferences during error handling

armsnyder:
I found a few places where there was the possibility of a panic if `response` was `nil`, which can happen if GitLab cannot be reached.

Also while I was there I removed some redundant status code checking, since the GitLab client library already treats the 204 code as a success.


roidelapluie:
Thanks!

PR #367, opened 2020-07-21
Add instance clusters to Terraform provider

sfang97:
Add support for instance level cluster management via the Gitlab Terraform provider.  Since [instance cluster API has been added to Gitlab](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/36001), instance cluster management should also be added to Terraform provider. The [API implementation is in the upstream package](https://github.com/xanzy/go-gitlab/pull/893)


sfang97:
Currently the CI checker is failing because "Unchanged files with check annotations" in `vendor/github.com/xanzy/go-gitlab/labels.go`. When I run `go mod vendor` a few files get modified including `labels.go`, and then when I run `make build` I get the `Error: gitlab/resource_gitlab_label.go:72:81: cannot use promoted field ListOptions.Page in struct literal of type gitlab.ListLabelsOptions`. After I `git restore labels.go`, make build passes.

stanhu:
@sfang97 Try this:

```diff
diff --git a/gitlab/resource_gitlab_label.go b/gitlab/resource_gitlab_label.go
index de1a3fe..f2fe533 100644
--- a/gitlab/resource_gitlab_label.go
+++ b/gitlab/resource_gitlab_label.go
@@ -69,7 +69,7 @@ func resourceGitlabLabelRead(d *schema.ResourceData, meta interface{}) error {
 	page := 1
 	labelsLen := 0
 	for page == 1 || labelsLen != 0 {
-		labels, _, err := client.Labels.ListLabels(project, &gitlab.ListLabelsOptions{Page: page})
+		labels, _, err := client.Labels.ListLabels(project, &gitlab.ListLabelsOptions{ListOptions: gitlab.ListOptions{Page: page}})
 		if err != nil {
 			return err
 		}
```

See https://github.com/golang/go/issues/9859 for more details.

stanhu:
I'm also wondering if the build weirdness is caused by this error that I got:

```
$ make
go install
go: inconsistent vendoring in /Users/stanhu/github/terraform-provider-gitlab:
	github.com/xanzy/go-gitlab@v0.32.1: is replaced in go.mod, but not marked as replaced in vendor/modules.txt

run 'go mod vendor' to sync, or use -mod=mod or -mod=readonly to ignore the vendor directory
```

sfang97:
@stanhu Thanks Stan, I tried replacing the line but no dice. The issue seems to be coming from the changes to `label.go` in the go-gitlab upstream package that aren't in the vendor file-- the last commit to the vendor version was 3 months ago, and the last commit to the go-gitlab package was 21 days ago, so I'm going to guess that something about the most recent change is making the compiler unhappy. I'll dig around in the diffs more to see if I can figure out what's going on. 

As for the weird `go mod vendor` error, the command is supposed to make a copy of the packages needed for the build/tests, so I can tell that it's trying to replace github.com/xanzy/go-gitlab v0.32.1 with github.com/sfang97/go-gitlab v0.33.1-0.20200731201522-09106effb0dc (< my unmerged change) which is correct. I'm not sure how to run that command in the CI check though?

sfang97:
@roidelapluie Hi there Julien! 👋 Could you review this PR?

gitlab/resource_gitlab_instance_cluster.go
```diff
@@ -0,0 +1,259 @@
...
+				Default:  true,
+				ForceNew: true,
+			},
+			"managed": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+				ForceNew: true,
+			},
+			"created_at": {
```
roidelapluie, line 44: That is computed

nicholasklick:
@armsnyder mind taking a pass at this?

gitlab/resource_gitlab_instance_cluster.go
```diff
@@ -0,0 +1,259 @@
...
+			},
+			"domain": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+				ForceNew: true,
```
armsnyder, line 36: Why is this force-new? It looks like the edit API lets you change `enabled`. https://docs.gitlab.com/ee/api/instance_clusters.html#edit-instance-cluster

gitlab/resource_gitlab_instance_cluster.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	log.Printf("[DEBUG] delete gitlab instance cluster %d", clusterId)
+
+	_, err = client.InstanceCluster.DeleteCluster(clusterId)
+
+	return err
+}
+
+func instanceClusterIdFromId(id string) (int, error) {
```
armsnyder, line 252: IMO just using `strconv.Atoi(id)` inline would be clearer, plus there isn't a a situation where you would encounter a string conversion error on the resource ID other than during a resource import, so the error message here doesn't really add anything.

gitlab/resource_gitlab_instance_cluster_test.go
```diff
@@ -0,0 +1,301 @@
...
+			return fmt.Errorf("got kubernetes authorization type %q; want %q", cluster.PlatformKubernetes.AuthorizationType, want.KubernetesAuthorizationType)
+		}
+
+		return nil
+	}
+}
+
+func testAccGitlabInstanceClusterConfig(rInt int, managed bool) string {
+	m := fmt.Sprintf("%t", managed)
+	if managed {
```
armsnyder, line 217: Why is this `if` block needed? Doesn't the `Sprintf` give you "true" or "false"? This might be dead code leftover from an earlier revision?

gitlab/resource_gitlab_instance_cluster_test.go
```diff
@@ -0,0 +1,301 @@
...
+  kubernetes_api_url            = "https://123.123.123"
+  kubernetes_token              = "some-token"
+  kubernetes_ca_cert            = "${trimspace(var.cert)}"
+  kubernetes_authorization_type = "abac"
+
+}
+`, instanceClusterFakeCert, rInt, m)
+}
+
+var instanceClusterFakeCert = `-----BEGIN CERTIFICATE-----
```
armsnyder, line 286: If the cert expires will it break the test? If so please add a comment showing how it is generated.

gitlab/resource_gitlab_instance_cluster_test.go
```diff
@@ -0,0 +1,301 @@
...
+}
+
+func testAccGitlabInstanceClusterConfig(rInt int, managed bool) string {
+	m := fmt.Sprintf("%t", managed)
+	if managed {
+		m = "true"
+	}
+
+	return fmt.Sprintf(`
+variable "cert" {
```
armsnyder, line 222: IMO the test would be clearer if you set the `kubernetes_ca_cert` attribute inline instead of setting it indirectly with a TF variable. Would this work?

```hcl
kubernetes_ca_cert = <<EOF
%s
EOF
```

gitlab/resource_gitlab_instance_cluster_test.go
```diff
@@ -0,0 +1,301 @@
...
+						KubernetesAuthorizationType: "abac",
+					}),
+				),
+			},
+			// create an unmanaged cluster
+			{
+				Config: testAccGitlabInstanceClusterConfig(rInt, false),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGitlabInstanceClusterExists("gitlab_instance_cluster.foo", &cluster),
+					testAccCheckGitlabInstanceClusterAttributes(&cluster, &testAccGitlabInstanceClusterExpectedAttributes{
```
armsnyder, line 42: This test step is a bit odd because your `Config` is different yet your `Check` is identical to the prior step. Can you update your `testAccGitlabInstanceClusterExpectedAttributes` helper to assert that the cluster changed from managed to unmanaged?

gitlab/resource_gitlab_instance_cluster_test.go
```diff
@@ -0,0 +1,301 @@
...
+  domain                        = "example-new.com"
+  kubernetes_api_url            = "https://124.124.124"
+  kubernetes_token              = "some-token"
+  kubernetes_ca_cert            = "${trimspace(var.cert)}"
+  kubernetes_authorization_type = "%s"
+}
+`, instanceClusterFakeCert, rInt, authType)
+}
+
+func testAccGitlabInstanceClusterManagement(rInt int, managed bool) string {
```
armsnyder, line 259: This looks like a duplicate of `testAccGitlabInstanceClusterConfig`.

armsnyder (commented):
Looks good! I left some clarifying questions and minor suggestions for ya.

armsnyder:
Looks like this MR needs a rebase as well.

gitlab/resource_gitlab_instance_cluster_test.go
sfang97: Good question Adam, I looked into this a little. `instanceClusterFakeCert` has to be a valid PEM certificate, but it doesn't need to be linked to a real cluster. For example I googled "fake pem certificate" and copied [this fake cert](https://gist.github.com/djsauble/f7b04dc331542e7c4419) into `resource_gitlab_instance_cluster_test.go` and the tests ran fine.

armsnyder: Cool thanks for checking

gitlab/resource_gitlab_instance_cluster_test.go
armsnyder: Yeah it's ok how it is. Sorry for the nitpick! 😓

gitlab/resource_gitlab_instance_cluster_test.go
```diff
@@ -0,0 +1,275 @@
...
+		}
+		if resp.StatusCode != 404 {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func testAccCheckGitlabInstanceClusterAttributes(cluster *gitlab.InstanceCluster, want *testAccGitlabInstanceClusterExpectedAttributes) resource.TestCheckFunc {
```
armsnyder, line 188: Can you add quality checks for `ManagementProjectID` and `Managed` to this function?

gitlab/resource_gitlab_instance_cluster_test.go
```diff
@@ -0,0 +1,275 @@
...
+				Config: testAccGitlabInstanceClusterConfig(rInt, true),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGitlabInstanceClusterExists("gitlab_instance_cluster.foo", &cluster),
+					testAccCheckGitlabInstanceClusterAttributes(&cluster, &testAccGitlabInstanceClusterExpectedAttributes{
+						Name:                        fmt.Sprintf("foo-cluster-%d", rInt),
+						Domain:                      "example.com",
+						EnvironmentScope:            "*",
+						KubernetesApiURL:            "https://123.123.123",
+						KubernetesCACert:            instanceClusterFakeCert,
+						KubernetesAuthorizationType: "abac",
```
armsnyder, line 34: Heads up you'll need `Managed: true` here.

gitlab/resource_gitlab_instance_cluster.go
```diff
@@ -0,0 +1,248 @@
...
+	if cluster.ManagementProject == nil {
+		d.Set("management_project_id", "")
+	} else {
+		d.Set("management_project_id", strconv.Itoa(cluster.ManagementProject.ID))
+	}
+
+	return nil
+}
+
+func resourceGitlabInstanceClusterUpdate(d *schema.ResourceData, meta interface{}) error {
```
armsnyder, line 181: Looks like you're missing setting the `enabled` and `kubernetes_namespace` attributes in `EditClusterOptions` in this function.

gitlab/resource_gitlab_instance_cluster.go
```diff
@@ -0,0 +1,248 @@
...
+			"domain": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+			"managed": {
```
armsnyder, line 37: Is it possible to change `managed` after creation? If not it should be `ForceNew: true`.

website/docs/r/instance_cluster.html.markdown
```diff
@@ -0,0 +1,64 @@
...
+
+- `domain` - (Optional, string) The base domain of the cluster.
+
+- `enabled` - (Optional, boolean) Determines if cluster is active or not. Defaults to `true`. This attribute cannot be read.
+
+- `managed` - (Optional, boolean) Determines if cluster is managed by gitlab or not. Defaults to `true`. This attribute cannot be read.
+
+- `kubernetes_api_url` - (Required, string) The URL to access the Kubernetes API.
+
+- `kubernetes_token` - (Required, string) The token to authenticate against Kubernetes.
```
armsnyder, line 46: For consistency you should also say `This attribute cannot be read.` here.

gitlab/resource_gitlab_instance_cluster.go
sfang97: Thanks! Looks like `enabled` wasn't an `EditClusterOptions` attribute, made a PR to add it: https://github.com/xanzy/go-gitlab/pull/939

gitlab/resource_gitlab_instance_cluster_test.go
sfang97: `Managed` wasn't an instance cluster attribute, adding it here: https://github.com/xanzy/go-gitlab/pull/940

gitlab/resource_gitlab_instance_cluster.go
armsnyder: Cool. In the meantime we can omit the `enabled` attribute from the schema or make it `ForceNew: true`.

gitlab/resource_gitlab_instance_cluster_test.go
armsnyder: Ok. I don't think it needs to block this PR as long as we create a new issue to follow up on it. I'm not sure how soon go-gitlab will release, as I'm not familiar with that project.

sfang97:
@armsnyder Thanks for the reviews Adam, I've applied the non-breaking changes and temporarily reverted the ones that are waiting for the go-gitlab changes to go in. Once those go in I can open a new PR in the terraform repo to reflect those changes. Take another look? :) 🏓

armsnyder (approved):
lgtm

nicholasklick:
So can we merge this?

sfang97:
@nicholasklick I think this can be merged, @armsnyder @roidelapluie what do y'all think?

armsnyder:
@sfang97 Go ahead, unless you're waiting for @roidelapluie's feedback.

armsnyder:
@sfang97 plz squash-merge lol 80 commits

armsnyder:
@sfang97 That's fine how you did it, but for future reference, I meant that when you merge this PR from the GitHub UI there is an option to squash.

sfang97:
@armsnyder Oops, understood! Thanks for the tip :)

PR #368, opened 2020-07-30
correct management_project_id argument in documentation

ahanselka:
The correct argument to add a cluster management project is in fact `management_project_id` as per [the project_cluster](https://github.com/terraform-providers/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_project_cluster.go#L94) and [group_cluster](https://github.com/terraform-providers/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_group_cluster.go#L90) resource.


roidelapluie:
thanks!

PR #370, opened 2020-08-01
Rename pc to gc

sfang97:
It seems like group_clusters.go was made by search-and-replacing the project_clusters.go file. Some instances of pc (stands for project cluster) were left in the group_clusters file, so this small PR just changes a few instances of pc into gc (stands for group cluster)


roidelapluie:
That should go to the upstream package.

sfang97:
Ah oops I needed to do this in the upstream library and not this repo. Deleting this branch and making a new one here: https://github.com/xanzy/go-gitlab/pull/896

PR #375, opened 2020-08-06
Add resource for pipelines email integration

### LINKED ISSUE #374: Bind the pipelines-email API
I would be nice to be able to set up Pipelines emails with terraform using this API:
https://docs.gitlab.com/ee/api/services.html#pipeline-emails

for instance
```tf
  name        = "example"
  description = "My awesome codebase"

  visibility_level = "public"
}

resource "gitlab_service_pipeline_emails" "example" {
  recipients = ["jeremad@jeremad.com"]
  notify_only_broken_pipelines = true
  branches_to_be_notified = "protected"
}
```

jeremad:
Closes #374

### DIFF
--- a/gitlab/resource_gitlab_service_pipelines_email.go
+++ b/gitlab/resource_gitlab_service_pipelines_email.go
@@ -1,0 +1,98 @@
+package gitlab
+
+import (
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+	gitlab "github.com/xanzy/go-gitlab"
+)
+
+func resourceGitlabServicePipelinesEmail() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGitlabServicePipelinesEmailCreate,
+		Read:   resourceGitlabServicePipelinesEmailRead,
+		Update: resourceGitlabServicePipelinesEmailCreate,
+		Delete: resourceGitlabServicePipelinesEmailDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"project": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"recipients": {
+				Type:     schema.TypeSet,
+				Required: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+			"notify_only_broken_pipelines": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+			"branches_to_be_notified": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.StringInSlice([]string{"all", "default", "protected", "default_and_protected"}, true),
+				Default:      "default",
+			},
+		},
+	}
+}
+
+func resourceGitlabServicePipelinesEmailSetToState(d *schema.ResourceData, service *gitlab.PipelinesEmailService) {
+	d.Set("recipients", strings.Split(service.Properties.Recipients, ","))
+	d.Set("notify_only_broken_pipelines", service.Properties.NotifyOnlyBrokenPipelines)
+	d.Set("branches_to_be_notified", service.Properties.BranchesToBeNotified)
+}
+
+func resourceGitlabServicePipelinesEmailCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gitlab.Client)
+	project := d.Get("project").(string)
+	d.SetId(project)
+	options := &gitlab.SetPipelinesEmailServiceOptions{
+		Recipients:                gitlab.String(strings.Join(*stringSetToStringSlice(d.Get("recipients").(*schema.Set)), ",")),
+		NotifyOnlyBrokenPipelines: gitlab.Bool(d.Get("notify_only_broken_pipelines").(bool)),
+		BranchesToBeNotified:      gitlab.String(d.Get("branches_to_be_notified").(string)),
+	}
+
+	log.Printf("[DEBUG] create gitlab pipelines emails service for project %s", project)
+
+	_, err := client.Services.SetPipelinesEmailService(project, options)
+	if err != nil {
+		return err
+	}
+
+	return resourceGitlabServicePipelinesEmailRead(d, meta)
+}
+
+func resourceGitlabServicePipelinesEmailRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gitlab.Client)
+	project := d.Id()
+
+	log.Printf("[DEBUG] read gitlab pipelines emails service for project %s", project)
+
+	service, _, err := client.Services.GetPipelinesEmailService(project)
+	if err != nil {
+		return err
+	}
+
+	d.Set("project", project)
+	resourceGitlabServicePipelinesEmailSetToState(d, service)
+	return nil
+}
+
+func resourceGitlabServicePipelinesEmailDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gitlab.Client)
+	project := d.Id()
+
+	log.Printf("[DEBUG] delete gitlab pipelines email service for project %s", project)
+
+	_, err := client.Services.DeletePipelinesEmailService(project)
+	return err
+}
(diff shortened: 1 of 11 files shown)


jeremad:
needs: https://github.com/xanzy/go-gitlab/pull/903

gitlab/resource_gitlab_service_pipelines_email.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+func resourceGitlabServicePipelinesEmail() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGitlabServicePipelinesEmailCreate,
+		Read:   resourceGitlabServicePipelinesEmailRead,
+		Update: resourceGitlabServicePipelinesEmailUpdate,
+		Delete: resourceGitlabServicePipelinesEmailDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceGitlabServicePipelinesEmailImportState,
```
armsnyder, line 19: I noticed your import and read functions are nearly identical. To save on some code, can use `State: schema.ImportStatePassthrough` here, and then change your read function to get the project from the resource id instead of the `project` attribute, and finally change your `resourceGitlabServicePipelinesEmailSetToState` function to set the `project` attribute. Since the read function is called by Terraform during import, there should be no difference.

gitlab/resource_gitlab_service_pipelines_email.go
```diff
@@ -0,0 +1,114 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
```
armsnyder, line 22: https://docs.gitlab.com/ee/api/services.html#pipeline-emails

Why omit the `add_pusher` and `pipeline_events` attributes?

gitlab/resource_gitlab_service_pipelines_email.go
```diff
@@ -0,0 +1,114 @@
...
+			"project": {
+				Type:     schema.TypeString,
+				Required: true,
```
armsnyder, line 25: I believe the `project` attribute should be `ForceNew: true`. Otherwise, if the user updates the attribute, it could add the service to the new project and leave the service enabled on the old project.

gitlab/resource_gitlab_service_pipelines_email.go
```diff
@@ -0,0 +1,114 @@
...
+			},
+			"recipients": {
+				Type:     schema.TypeString,
```
armsnyder, line 28: This would be more user friendly if the type were `list(string)` and you took care of comma-joining them before the API call.

CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
 ## 2.12.0 (Unreleased)
+
+FEATURES:
+* **New Resource:** `gitlab_service_pipeline_emails`
```
armsnyder, line 4: Usually the maintainer takes care of the changelog, but if you're going to update it yourself, be sure to use the correct format to reference the issue. Example: https://github.com/terraform-providers/terraform-provider-gitlab/commit/f05eed843ae5a6e1725f9e2523e6cf849371d846

gitlab/resource_gitlab_service_pipelines_email.go
jeremad: It does say it in the doc, bu I suspect these fields come from an older version, and are now mapped to new settings. Where trying to create a a pipeline service via the API, these are never returned.
Also, I only mapped what was avaible in the UI.

gitlab/resource_gitlab_service_pipelines_email.go
jeremad: Yeah absoletely, I got lazy

CHANGELOG.md
jeremad: I copied from:
```
FEATURES:
* **New Resource:** `gitlab_service_github`
  ([#311](https://github.com/terraform-providers/terraform-provider-gitlab/issues/311))
```
in the diff you linked. I'd be happy to fix my changelog, but I don't really what differs from usual?

gitlab/resource_gitlab_service_pipelines_email.go
jeremad: Well, a bit trickier than I thought. First the ID of the resource and the project ID are different. I could use the project ID though, since you can only have one resource per project

jeremad: I can probably remove the code of the import function. I ended up with that while debugging, but actually I fixed my issue in the go-gitlab package

CHANGELOG.md
armsnyder: The tf-release-bot expects the form `[GH-375]`, where 375 is the PR or issue number. During a release it replaces those references with the markdown format you saw.

https://www.terraform.io/docs/extend/best-practices/versioning.html#entry-format

Bot commit be like https://github.com/terraform-providers/terraform-provider-gitlab/commit/d48f13a9bfb54a161b2bd25fa7de92340f65f3ea

gitlab/resource_gitlab_service_pipelines_email.go
armsnyder: I'm not sure I see where the issue is. (Maybe you already fixed it in a rebase?) It looks like you can use `ImportStatePassthrough`. Here's the relevant section of the docs in case it helps. 😄 
https://www.terraform.io/docs/extend/resources/import.html#importer-state-function

CHANGELOG.md
jeremad: OK sure, I'll change what I did so that a bot can do the exact same thing later :D 
Thank you very much for the kid review. I was a little light on reading the documentation.

jeremad: I'll just remove my changelog changes as recommended by Hashicorp. BTW, the Changelog of this Repo does no follow the specification!

armsnyder:
In case you didn't see my reply about the importer. (It was on a resolved thread.) https://github.com/terraform-providers/terraform-provider-gitlab/pull/375#discussion_r467501760

gitlab/resource_gitlab_service_pipelines_email.go
jeremad: The issue is that the read fonction does not have access to the project ID. I use the project ID in omport function for ease of use since services are unique per project and the service ID is pretty hard to retrieve.
So I don't think I can only use the `ImportStatePassthrough` since I need to write the project in the state.

I am not sure though. Does that make sense?

jeremad: That makes me think there is probably a bug in my code

gitlab/resource_gitlab_service_pipelines_email.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		return err
+	}
+
+	return resourceGitlabServicePipelinesEmailRead(d, meta)
+}
+
+func resourceGitlabServicePipelinesEmailRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gitlab.Client)
+	project := d.Get("project").(string)
```
armsnyder, line 77: Can you change this to

```suggestion
	project := d.Id()
```

so you don't need the importer?

Idea from here - https://www.terraform.io/docs/extend/resources/import.html#importer-state-function
> Sometimes it is possible to adjust the resource Read function to replace d.Get() use with d.Id() if they exactly match

gitlab/resource_gitlab_service_pipelines_email.go
```diff
@@ -0,0 +1,103 @@
...
+				Optional:     true,
+				ValidateFunc: validation.StringInSlice([]string{"all", "default", "protected", "default_and_protected"}, true),
+				Default:      "default",
+			},
+		},
+	}
+}
+
+func resourceGitlabServicePipelinesEmailSetToState(d *schema.ResourceData, service *gitlab.PipelinesEmailService) {
+	d.SetId(fmt.Sprintf("%d", service.ID))
```
armsnyder, line 50: There would be a bug here, if you're using the project ID as the resource ID

jeremad: Let's get it straight because I am not sure we understand each other:

1) The project ID is different from the resource ID (== service ID of the GitLab API in my current design)
2) I cannot read the service from the GitLab API with the service ID (I need the project ID)
3) I don't think it is a good idea to use the project ID as the resource ID, but may be it is usual to do that?

armsnyder: Ah, I thought you had already settled on using the project ID as the resource ID. My bad.

I would definitely suggest using the project ID as the resource ID, because it is the ID that you use to import, and more importantly, the ID you use to lookup the service in GitLab.

jeremad: OK let's fo that than. Thanks

jeremad:
Thank *you* for helping me with this!

jeremad:
Any more reviewer? :)

jeremad:
@nicholasklick ? :angel:

armsnyder:

PR #378, opened 2020-08-07
Bump go-gitlab to v.0.34.1

sfang97:
Bump go-gitlab to v.0.34.1 

Needed to fix nested struct format in `gitlab/resource_gitlab_label.go`

Was failing with this error: 
gitlab/resource_gitlab_label.go:72:81: cannot use promoted field ListOptions.Page in struct literal of type gitlab.ListLabelsOptions
gitlab/resource_gitlab_label_test.go:106:80: cannot use promoted field ListOptions.PerPage in struct literal of type gitlab.ListLabelsOptions

because we weren't accessing the nested struct correctly.


mattkasa:
Looks good to me @sfang97, thanks!

PR #379, opened 2020-08-08
Fix isRunningInEE helper

sfang97:
Several tests are only meant to run in Gitlab CE, so we were using a helper function `isRunningInEE` to detect the provider version and skip functions that only run in core and not in EE. We found that when running tests that use the `isRunningInEE` function, like `gitlab/resource_gitlab_service_github_test.go` and `gitlab/resource_gitlab_project_push_rules_test.go` , the test would fail because "Provider not initialized, unable to get GitLab connection". This is because when running single one-off tests, the provider is indeed not initialized, so `isRunningInEE` would not be able to find the provider nor the version in its metadata. Running single tests that relied on this function would fail, but running all the tests with `make testacc` would pass because the provider is initialized when running a whole test suite and not single tests. 

So we change the `isRunningInEE` function to look for the Gitlab license file instead, which is generated by the acceptance-ee test in `.github/workflows/acceptance-tests.yml` and is not generated for CE tests. This should pass local tests that use the function, no matter what order the tests are run in.

### DIFF
--- a/gitlab/provider_test.go
+++ b/gitlab/provider_test.go
@@ -11,10 +11,8 @@ import (
 var testAccProviders map[string]terraform.ResourceProvider
 var testAccProvider *schema.Provider
 
-const TestEnvVar = "TF_ACC"
-
 func init() {
-	if os.Getenv(TestEnvVar) != "" {
+	if os.Getenv("TF_ACC") != "" {
 		testAccProvider = Provider().(*schema.Provider)
 		if err := testAccProvider.Configure(&terraform.ResourceConfig{}); err != nil {
 			panic(err)
--- a/gitlab/provider_test.go
+++ b/gitlab/provider_test.go
@@ -12,12 +12,14 @@ var testAccProviders map[string]terraform.ResourceProvider
 var testAccProvider *schema.Provider
 
 func init() {
-	testAccProvider = Provider().(*schema.Provider)
-	if err := testAccProvider.Configure(&terraform.ResourceConfig{}); err != nil {
-		panic(err)
-	}
-	testAccProviders = map[string]terraform.ResourceProvider{
-		"gitlab": testAccProvider,
+	if os.Getenv(resource.TestEnvVar) != "" {
+		testAccProvider = Provider().(*schema.Provider)
+		if err := testAccProvider.Configure(&terraform.ResourceConfig{}); err != nil {
+			panic(err)
+		}
+		testAccProviders = map[string]terraform.ResourceProvider{
+			"gitlab": testAccProvider,
+		}
 	}
 }
 
--- a/gitlab/provider_test.go
+++ b/gitlab/provider_test.go
@@ -10,6 +10,7 @@ import (
 
 var testAccProviders map[string]terraform.ResourceProvider
 var testAccProvider *schema.Provider
+
 const TestEnvVar = "TF_ACC"
 
 func init() {
--- a/gitlab/provider_test.go
+++ b/gitlab/provider_test.go
@@ -13,6 +13,9 @@ var testAccProvider *schema.Provider
 
 func init() {
 	testAccProvider = Provider().(*schema.Provider)
+	if err := testAccProvider.Configure(&terraform.ResourceConfig{}); err != nil {
+		panic(err)
+	}
 	testAccProviders = map[string]terraform.ResourceProvider{
 		"gitlab": testAccProvider,
 	}
--- a/gitlab/provider_test.go
+++ b/gitlab/provider_test.go
@@ -12,7 +12,7 @@ var testAccProviders map[string]terraform.ResourceProvider
 var testAccProvider *schema.Provider
 
 func init() {
-	if os.Getenv(resource.TestEnvVar) != "" {
+	if os.Getenv(TestEnvVar) != "" {
 		testAccProvider = Provider().(*schema.Provider)
 		if err := testAccProvider.Configure(&terraform.ResourceConfig{}); err != nil {
 			panic(err)
--- a/gitlab/provider_test.go
+++ b/gitlab/provider_test.go
@@ -10,6 +10,7 @@ import (
 
 var testAccProviders map[string]terraform.ResourceProvider
 var testAccProvider *schema.Provider
+const TestEnvVar = "TF_ACC"
 
 func init() {
 	if os.Getenv(TestEnvVar) != "" {
--- a/gitlab/provider_test.go
+++ b/gitlab/provider_test.go
@@ -1,18 +1,18 @@
 package gitlab
 
 import (
-	"os"
-	"testing"
-
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"os"
+	"testing"
 )
 
 var testAccProviders map[string]terraform.ResourceProvider
 var testAccProvider *schema.Provider
 
 func init() {
-	if os.Getenv("TF_ACC") != "" {
+	if os.Getenv(resource.TestEnvVar) != "" {
 		testAccProvider = Provider().(*schema.Provider)
 		if err := testAccProvider.Configure(&terraform.ResourceConfig{}); err != nil {
 			panic(err)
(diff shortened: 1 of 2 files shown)


roidelapluie:
Note that I think the CI license has expired last week

roidelapluie:
I do not agree with this fix. The acceptant test should run locally as well, on only on github actions. WDYT?

sfang97:
@roidelapluie Thanks for the insight Julien. Could you elaborate a little on what you mean? With the changes, the acceptance tests run locally (single test and full suite) as well as with github actions. Without the changes, the acceptance tests will not pass single tests locally, and are flaky on github due to test execution order I think.

roidelapluie:
The acceptance tests can run on a regular gitlab instance, not only the instances started with start-gitlab.sh

jeremad:
Jumping on the conversation here. I find this issue very annoying as well. May be an environment variable can be used to tell the test suite whether the GitLab instance is EE or CE?

armsnyder:
While we're spitballing, what about making the same API call, but using a new client instead of the one in the provider metadata? It can be initialized using the same environment variables.

armsnyder:
Oh I see now it's not an API call. Then, initialize the client once using `sync.Once` and read the version from it.

roidelapluie:
The provider should be initialized all the time. We initialize it in init().

sfang97:
```
func isRunningInEE() (bool, error) {
	if conn, ok := testAccProvider.Meta().(*gitlab.Client); ok {
		version, _, err := conn.Version.GetVersion()
		if err != nil {
			return false, err
		}
		if strings.Contains(version.String(), "-ee") {
			return true, nil
		}
	} else {
		return false, errors.New("Provider not initialized, unable to get GitLab connection")
	}
	return false, nil
}
```
With the version of `isRunningInEE` currently in master, `gitlab/resource_gitlab_service_github_test.go` and `gitlab/resource_gitlab_project_push_rules_test.go` fail when run individually with the error "Provider not initialized, unable to get GitLab connection". So the `if conn` conditional was not hit and that the provider has not been initialized. We can further verify this by putting a print statement in `provider.go:117` and we can see that `func providerConfigure` never gets called.

armsnyder:
Here's how I can reproduce the problem. It seems like the provider is not fully initialized before the first call to a `SkipFunc`.

1. Run GitLab CE
   ```
   make testacc-prepareenv
   ```
2. Running one test that uses `SkipFunc`, fails:
   ```
   GITLAB_BASE_URL=http://localhost:8080/api/v4/ GITLAB_TOKEN=ACCTEST TESTARGS="-run TestAccGitlabProjectPushRules_basic" make testacc
   ```
   ```
   TF_ACC=1 go test -v ./gitlab -run TestAccGitlabProjectPushRules_basic -timeout 40m
   === RUN   TestAccGitlabProjectPushRules_basic
   TestAccGitlabProjectPushRules_basic: testing.go:636: Provider not initialized, unable to get GitLab connection
   --- FAIL: TestAccGitlabProjectPushRules_basic (0.00s)
   FAIL
   FAIL    github.com/terraform-providers/terraform-provider-gitlab/gitlab 0.325s
   FAIL
   make: *** [testacc] Error 1
   ```
3. Running the same test after first running a test that does not use `SkipFunc`, succeeds:
   ```
   $ GITLAB_BASE_URL=http://localhost:8080/api/v4/ GITLAB_TOKEN=ACCTEST TESTARGS="-run \"(TestAccGitlabDeployToken_basic|TestAccGitlabProjectPushRules_basic)\"" make testacc
   ```
   ```
   TF_ACC=1 go test -v ./gitlab -run "(TestAccGitlabDeployToken_basic|TestAccGitlabProjectPushRules_basic)" -timeout 40m
   === RUN   TestAccGitlabDeployToken_basic
   --- PASS: TestAccGitlabDeployToken_basic (7.79s)
   === RUN   TestAccGitlabProjectPushRules_basic
   --- PASS: TestAccGitlabProjectPushRules_basic (0.11s)
   PASS
   ok      github.com/terraform-providers/terraform-provider-gitlab/gitlab 8.226s
   ```

The fact that the acceptance tests pass at all seems to be a coincidence, that the first test that is run does not use `SkipFunc`.

armsnyder:
One fix would be to call `Configure` in the `init` function.

```go
func init() {
	testAccProvider = Provider().(*schema.Provider)
	if err := testAccProvider.Configure(&terraform.ResourceConfig{}); err != nil {
		panic(err)
	}
	testAccProviders = map[string]terraform.ResourceProvider{
		"gitlab": testAccProvider,
	}
}
```

No more errors.

```
GITLAB_BASE_URL=http://localhost:8080/api/v4/ GITLAB_TOKEN=ACCTEST TESTARGS="-run TestAccGitlabProjectPushRules_basic" make testacc
TF_ACC=1 go test -v ./gitlab -run TestAccGitlabProjectPushRules_basic -timeout 40m
=== RUN   TestAccGitlabProjectPushRules_basic
--- PASS: TestAccGitlabProjectPushRules_basic (0.06s)
PASS
ok      github.com/terraform-providers/terraform-provider-gitlab/gitlab 0.464s
```

sfang97:
@armsnyder Thanks for the suggestion. I tried calling `Configure` in `provider_test.go#init` and it's passing the previously failing tests with no issue. However, when it tries running the whole suite in tests in GH, it hits the panic and fails with this error: `panic: GET https://gitlab.com/api/v4/user: 401 {message: 401 Unauthorized}`. I'm not sure what's going on, but to me this smells like the error with `SkipFunc` is a symptom of how the client is configured.

armsnyder:
Oh, the unit tests are failing. We should probably only be initializing the provider in acceptance tests. You could get this by wrapping the `init` code in

```go
if os.Getenv(resource.TestEnvVar) != ""
```

gitlab/provider_test.go
```diff
@@ -11,10 +11,17 @@ import (
 var testAccProviders map[string]terraform.ResourceProvider
 var testAccProvider *schema.Provider
 
+const TestEnvVar = "TF_ACC"
```
armsnyder, line 14: The const is already defined in the terraform-provider-sdk `resource` package.

sfang97: When I don't re define it in provider_test.go, it throws an undefined variable error. Should I just use `os.Getenv("TF_ACC)` instead?

sfang97: Hmm now just the MacOS unit tests are failing

armsnyder: Did you make sure to import `"github.com/hashicorp/terraform-plugin-sdk/helper/resource"` in `provider_test.go`?

I'm not sure about the MacOS unit tests. 😢 It could be unrelated to your PR.

sfang97:
@roidelapluie What do you think about this solution? We now explicitly initialize the provider in during acceptance tests, so the original implementation of `isRunningInEE` will be able to correctly read the provider metadata and detect license type.

gitlab/provider_test.go
sfang97: Seems like a flaky test, rerunning passed it 🤷

gitlab/provider_test.go
```diff
@@ -12,9 +12,14 @@ var testAccProviders map[string]terraform.ResourceProvider
 var testAccProvider *schema.Provider
 
 func init() {
-	testAccProvider = Provider().(*schema.Provider)
-	testAccProviders = map[string]terraform.ResourceProvider{
-		"gitlab": testAccProvider,
+	if os.Getenv("TF_ACC") != "" {
```
armsnyder, line 15: Did you see my last comment about adding the import and using the const? It would be good to refer to the same const that the Terraform test engine uses.

sfang97: Thanks for the suggestion-- I tried importing the resource package but for some reason it's not using the variable correctly: 
```
gitlab/provider_test.go:4:2: imported and not used: "github.com/hashicorp/terraform-plugin-sdk/helper/resource"
gitlab/provider_test.go:15:15: undefined: TestEnvVar
```

jeremad: you need to use `resource.TestEnvVar` 👍

armsnyder: This builds for me. Is this the same thing you tried? https://github.com/sfang97/terraform-provider-gitlab/pull/1

sfang97: Ah, I'd left out the `resource.` part, thanks y'all!

PR #380, opened 2020-08-08
gitlab_project_variable: friendlier error when masked variable is invalid

### LINKED ISSUE #371: gitlab_project_variable: Add `masked` checks
Hi, I was recently trying to add a few variables using `gitlab_project_variable` but was running into some issues with the `masked` option.

It turns out that the cause of my issue is just that my assigned variables didn't meet the masked variables requirements, which are described [here](https://gitlab.com/help/ci/variables/README#mask-a-custom-variable).

It would be great if these checks could be done either on validate or even just return a more informative error message at apply time.

For my case I was assigning a variable to the value so I doubt it would be possible to do the checks before apply.

### Terraform Version

Terraform v0.12.29
+ provider.digitalocean v1.22.0
+ provider.gitlab v2.11.0
+ provider.helm v1.2.4
+ provider.kubernetes v1.12.0

### Affected Resource(s)

- gitlab_project_variable

```hcl
provider "gitlab" {
  token = var.gitlab_token
}

data "gitlab_project" "my-project" {
  id = 123
}

resource "gitlab_project_variable" "container-auth" {
  project   = data.gitlab_project.ci-node-example.id
  key       = "CONTAINER_REGISTRY"
  value     = {"auths":{"registry.digitalocean.com":{"auth":"XXXXX"}}} # <--- Cannot handle "{"
  protected = true
  environment_scope = "*"
  masked = true # <--- error
}

# Same as above but the value is derived from a 
resource "gitlab_project_variable" "container-auth-2" {
  project   = data.gitlab_project.ci-node-example.id
  key       = "CONTAINER_REGISTRY_FROM_VAR"
  value     = var.container_registry_credentials # <-- imported from another module
  protected = true
  environment_scope = "*"
  masked = true # <--- error
}
```

### Debug Output

No crash per se. Here is the attempt to create a plan:

https://gist.github.com/michaelwooley/a3ff9c1d7085f6855839c1dda7a9e4cf

### Panic Output

N/A

### Expected Behavior

Either:

- A failed apply with a message along the lines of "invalid mask variable received, see the requirements here https://gitlab.com/help/ci/variables/README#masked-variable-requirements"
- ()

### Actual Behavior

Opaque error seen above. This caused me to wonder about a bug here first. Probably not the best thought process but the fact that it worked with masked = false suggested it. The relevant docs on this are buried pretty deep unless you try to go in through the UI.

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

N/A

### References

N/A

armsnyder:
Closes #371 

Rather than adding validation I intercepted the error.

Old error:
```
Error: POST https://gitlab.com/api/v4/projects/12345678/variables: 400 {message: {value: [is invalid]}}
```

New error:
```
Error: Invalid value for a masked variable. Check the masked variable requirements: https://docs.gitlab.com/ee/ci/variables/#masked-variable-requirements
```

### DIFF
--- a/gitlab/resource_gitlab_project_variable.go
+++ b/gitlab/resource_gitlab_project_variable.go
@@ -1,7 +1,10 @@
 package gitlab
 
 import (
+	"errors"
 	"log"
+	"net/http"
+	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	gitlab "github.com/xanzy/go-gitlab"
@@ -82,7 +85,7 @@ func resourceGitlabProjectVariableCreate(d *schema.ResourceData, meta interface{
 
 	_, _, err := client.ProjectVariables.CreateVariable(project, &options)
 	if err != nil {
-		return err
+		return augmentProjectVariableClientError(d, err)
 	}
 
 	d.SetId(buildTwoPartID(&project, &key))
@@ -102,7 +105,7 @@ func resourceGitlabProjectVariableRead(d *schema.ResourceData, meta interface{})
 
 	v, _, err := client.ProjectVariables.GetVariable(project, key)
 	if err != nil {
-		return err
+		return augmentProjectVariableClientError(d, err)
 	}
 
 	d.Set("key", v.Key)
@@ -140,7 +143,7 @@ func resourceGitlabProjectVariableUpdate(d *schema.ResourceData, meta interface{
 
 	_, _, err := client.ProjectVariables.UpdateVariable(project, key, options)
 	if err != nil {
-		return err
+		return augmentProjectVariableClientError(d, err)
 	}
 
 	return resourceGitlabProjectVariableRead(d, meta)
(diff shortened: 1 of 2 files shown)


gitlab/resource_gitlab_project_variable.go
```diff
@@ -155,3 +158,22 @@ func resourceGitlabProjectVariableDelete(d *schema.ResourceData, meta interface{
 	_, err := client.ProjectVariables.RemoveVariable(project, key)
 	return err
 }
+
+func resourceGitlabProjectVariableAugmentClientError(d *schema.ResourceData, err error) error {
```
jeremad, line 162: Do we need to prefix all of the function names like that? 🤔

armsnyder: Nope haha. Renamed in latest push.

michaelwooley:
Can't speak to the wider code quality, etc. However, the message/change looks like it should save some people a few headaches! Thanks for the turnaround.

armsnyder:

PR #382, opened 2020-08-09
Fix Deprecated SDK Function Usage

armsnyder:
Hi,

Now that Terraform SDK v2 is out, I scanned the codebase for deprecated v1 functions and found a few:
* `d.Partial`
* `d.SetPartial`
* `validation.ValidateRFC3339TimeString`

Removal notices for these functions here:
* https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#removal-of-helper-schema-resourcedata-setpartial
* https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#removal-of-deprecated-validation-functions

The removal of `Partial` and `SetPartial` from `gitlab_project` may look scary, but they're actually not adding anything to the resource behavior. Terraform performs a read before every operation to detect drift anyway. If for example in the `resourceGitlabProjectCreate` function project creation succeeds, but sharing it with groups fails, Terraform reports the error to the user, and on the next `terraform plan` will automatically have the project-creation attributes filled in by calling `resourceGitlabProjectRead`, and proceed with group sharing from that point, using the logic in `resourceGitlabProjectUpdate`. It all works without the need for explicit partial state management.

More info about this deprecation and removal in Terraform v2 here:
* https://github.com/hashicorp/terraform-plugin-sdk/issues/312

While I was looking at where `gitlab_project` might be trying to do too much with explicit state management, I saw there was redundant logic in `archiveProject` and `unarchiveProject` as well, where Terraform will detect this drift automatically, so I removed those too.

### DIFF
--- a/gitlab/resource_gitlab_deploy_token.go
+++ b/gitlab/resource_gitlab_deploy_token.go
@@ -45,7 +45,7 @@ func resourceGitlabDeployToken() *schema.Resource {
 			"expires_at": {
 				Type:         schema.TypeString,
 				Optional:     true,
-				ValidateFunc: validation.ValidateRFC3339TimeString,
+				ValidateFunc: validation.IsRFC3339Time,
 				ForceNew:     true,
 			},
 			"scopes": {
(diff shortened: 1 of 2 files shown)


armsnyder:

PR #391, opened 2020-08-14
Documentation: Add group_ldap_link to index

sirlatrom:
It was missing from the index in the left side, and I almost started implementing it myself before finding it in the source code.


nicholasklick:
LGTM

PR #356, opened 2020-07-03
Add Project-level MR approvals

husunal:
This MR adds a [Project-level MR approvals](https://docs.gitlab.com/ee/api/merge_request_approvals.html#project-level-mr-approvals) resource.

PR #401, opened 2020-09-10
Fix incorrect type assertion in gitlab_project_level_mr_approvals

armsnyder:
My apologies, I should have run the workflow for #356 in a branch to cover the new EE acceptance test, before merging. This should fix it. For CHANGELOG purposes we can consider this part of #356 and not a bugfix.


armsnyder:
I'm going to merge this now since the master workflow is broken, and this fix is clear.

PR #402, opened 2020-09-10
Fix project_id not being set in gitlab_project_level_mr_approvals

armsnyder:
EE acceptance tests are still failing in master. This should be the last fix needed. This is to fix a bug in #356 where the `project_id` attribute was not set in `gitlab_project_level_mr_approvals` during read, because it was the wrong type.

As with #401 this does not need an entry in the Changelog since it is for a new resource.


armsnyder:
Merging to fix master workflow

PR #406, opened 2020-09-10
Fix access level docs for branch protection.

yeldiRium:


roidelapluie:
Can't you set e.g. reporter?

yeldiRium:
Nope. The GitLab API only allows the three I mentioned for push access.

roidelapluie:
I think that this change is not correct as we do not prevent you to put another level (and we should not: https://www.terraform.io/docs/extend/hashicorp-provider-design-principles.html#resource-and-attribute-schema-should-closely-match-the-underlying-api ).

armsnyder:
The [API docs for branch protection](https://docs.gitlab.com/ee/api/protected_branches.html#protect-repository-branches) list `no one`, `developer`, `maintainer` and `admin` as the possible values for this particular API.

The version of go-gitlab that we use does not recognize `admin` as an `AccessLevelValue` however.

There seems to be precedence in this provider for performing string validation on attributes. Even though the upstream API call does this validation for us, I think there is value in putting it in the provider too since the error message will be clearer for the user.

If we do go ahead with this change it would be good to also modify the [schema validation](https://github.com/gitlabhq/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_branch_protection.go#L12) to only accept these values.

yeldiRium:
I actually don't care about string validation. If wrong values do not work and an error is reported that's fine by me. I just wanted to fix a documentation that contained wrong information.

PR #408, opened 2020-09-11
gitlab_group: Revert recent change to visibility_level attribute

armsnyder:
Reverts an [unreleased breaking change](https://github.com/gitlabhq/terraform-provider-gitlab/commit/01b429b0cbd42be21cad59167f8b2460f4282582#diff-ea50bce9f20ae93ae983c1f538a2d66bR64) as requested by @roidelapluie [here](https://github.com/gitlabhq/terraform-provider-gitlab/pull/362#discussion_r487260162). I also clarified the default behavior in the docs.

### DIFF
--- a/website/docs/r/group.html.markdown
+++ b/website/docs/r/group.html.markdown
@@ -44,7 +44,7 @@ support for projects in this group.
 * `request_access_enabled` - (Optional) Boolean, defaults to false.  Whether to
 enable users to request access to the group.
 
-* `visibility_level` - (Optional) The visibility of the group. Can be `private`, `internal`, or `public`.
+* `visibility_level` - (Optional) The group's visibility. Can be `private`, `internal`, or `public`.
 
 * `share_with_group_lock` - (Optional) Boolean, defaults to false.  Prevent sharing
 a project with another group within this group.
--- a/website/docs/r/group.html.markdown
+++ b/website/docs/r/group.html.markdown
@@ -44,7 +44,7 @@ support for projects in this group.
 * `request_access_enabled` - (Optional) Boolean, defaults to false.  Whether to
 enable users to request access to the group.
 
-* `visibility_level` - (Optional) The group’s visibility. Can be `private`, `internal`, or `public`.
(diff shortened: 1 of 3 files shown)


website/docs/r/group.html.markdown
```diff
@@ -44,9 +44,8 @@ support for projects in this group.
 * `request_access_enabled` - (Optional) Boolean, defaults to false.  Whether to
 enable users to request access to the group.
 
-* `visibility_level` - (Optional) Set to `public` to create a public group.
-  Valid values are `private`, `internal`, `public`.
-  Groups are created as private by default.
+* `visibility_level` - (Optional) The group’s visibility. Can be `private`, `internal`, or `public`.
+The default will be the default value configured in GitLab, usually `private`.
```
roidelapluie, line 48: Do we need this sentence ? it makes it feel that we query gitlab to get the default.

armsnyder: That is basically what happens with O+C arguments though, based on my understanding. If the value is not set then Terraform reads it and saves it ruing the next read. How would you rephrase it?

roidelapluie: Yes that's what happens, but I would not write the sentence.

armsnyder: Cool I'll remove it

roidelapluie:
I am not english native but I think `'s` is limited for people. It should be `the visibility of the group` here.

armsnyder:
> I am not english native but I think `'s` is limited for people. It should be `the visibility of the group` here.

Happy to make that change for inclusivity. I copied it from the [GitLab docs](https://docs.gitlab.com/ee/api/groups.html#new-group).

roidelapluie:
Ooh then it was fine :)

armsnyder:
> Ooh then it was fine :)

Haha no worries. Maybe @nicholasklick can take this as general feedback for GitLab documentation. 😄

roidelapluie (approved):

Issue #411, opened 2020-09-15
Publish 3.0.0 release


nicholasklick:
@roidelapluie @armsnyder 

Are we ready to publish a new major release?

What has been the process previously?

1. Create a new release
1. Notify Hashicorp via slack?

Thanks.

roidelapluie:
Previously hashicorp did all the work. I don't know what the new procedure is.

roidelapluie:
@armsnyder do you know if CHANGELOG is up to date?

armsnyder:
@nicholasklick I also don't know how to do a release. Maybe when you finish the Terraform Registry setup you'll get some instructions from Hashicorp?

@roidelapluie I just pushed an update for the last couple merges.

attachmentgenie:
publishing a release to the registry requires finishing the terraform registry setup (that person needs admin permissions on this repository), Setting a gpg key (and sharing the public part with hashicorp) as a github secret, and running the workflow from #412.

armsnyder:
@attachmentgenie Yup, sounds right.

@nicholasklick Assuming you're admin, you'll probably find some info in the publish page here, after logging in: https://registry.terraform.io/publish/provider. When I visit that page it doesn't let me select the gitlabhq org, so I'm not sure what lies beyond.

liuyangc3:
👍 Glad to see the new release in terraform registry, we are waiting for it so long

nicholasklick:
Looks like the `release` action worked!
https://github.com/gitlabhq/terraform-provider-gitlab/releases/tag/v3.0.0-pre

armsnyder:
@nicholasklick Awesome! Did you write the release notes yourself or were they generated by the releaser tool? Also, just making sure you saw our conversation [here](https://github.com/gitlabhq/terraform-provider-gitlab/pull/416#issuecomment-694003637) about possibly removing the `gitlab_push_rule` and/or `gitlab_project_share_group` resources. Can we come to a consensus about that and do it before releasing 3.0.0?

nicholasklick:
@armsnyder the release notes were generated.

nicholasklick:
<img width="1144" alt="Screen Shot 2020-09-18 at 1 06 37 AM" src="https://user-images.githubusercontent.com/24757/93558366-4bee6700-f94b-11ea-9029-37984d509f34.png">
 Looks like publishing to the registry is working also!

nicholasklick:
[...]

PR #412, opened 2020-09-15
import of release workflow from provider scafolding

attachmentgenie:
As taken from [terraform-provider-scaffolding](https://github.com/hashicorp/terraform-provider-scaffolding)

This workflow creates a draft release with all the required binary artifacts that the hashicorp registry requires. It requires an admin to set a github secret called GPG_PRIVATE_KEY which is the gpg  key with which the binaries will be signed

This workflow is a required step for achieving #411

### DIFF
--- a/.github/.goreleaser.yml
+++ b/.github/.goreleaser.yml
@@ -1,0 +1,54 @@
+# Visit https://goreleaser.com for documentation on how to customize this
+# behavior.
+before:
+    hooks:
+      # this is just an example and not a requirement for provider building/publishing
+      - go mod tidy
+  builds:
+  - env:
+      # goreleaser does not work with CGO, it could also complicate
+      # usage by users in CI/CD systems like Terraform Cloud where
+      # they are unable to install libraries.
+      - CGO_ENABLED=0
+    mod_timestamp: '{{ .CommitTimestamp }}'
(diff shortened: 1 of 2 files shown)


armsnyder:
@attachmentgenie Do you also need the [.goreleaser.yml](https://github.com/hashicorp/terraform-provider-scaffolding/blob/master/.goreleaser.yml) file?

attachmentgenie:
my apologies, yes  you do!

nicholasklick:
I have added two secrets to support this PR.

nicholasklick:
Shall we merge and test it? @roidelapluie @armsnyder

armsnyder:
Yeah we can merge and make a 3.0.0-rc1

roidelapluie:
what do we do for the push rules?

armsnyder:
Probably we should remove push rules and project group share before doing the real 3.0.0 release, but this rc1 would just be to test the workflow.

roidelapluie (approved):

nicholasklick:
also FYI Hashicorp "can add your provider to a tool we have that resigns and posts to releases.hashicorp.com so they are available to pre-0.13"

They will create builds support pre-0.13 also.

PR #366, opened 2020-07-21
feat(gitlab-application): Add resource for gitlab application, data source for current user

vymarkov:
Fixes #321 

This PR adds the ability to create [an application](https://docs.gitlab.com/ee/api/applications.html), data source for [gitlab current user](https://docs.gitlab.com/ce/api/users.html#list-current-user-for-normal-users) to don't fail in case if gitlab isn't self-hosted, see below the terraform configuration for this case.

```hcl
data "gitlab_current_user" "me" {
}

// https://github.com/terraform-providers/terraform-provider-gitlab/issues/321
resource "gitlab_application" "grafana" {
  // create the application if user has admin role, 
  // otherwise Gitlab API rejects requests with 403 http status code
  count = data.gitlab_current_user.me.is_admin ? 1 : 0

  name = "grafana"
  redirect_uri = [
    "http://localhost:8080/auth/gitlab/callback",
    "http://localhost:8080/auth/gitlab/callback2"
  ]
  scopes = ["openid", "email", "profile"]
}

output "me" {
  value = data.gitlab_current_user.me
}

output "grafana_app_secret" {
  value = {
    app_id     = gitlab_application.grafana[0].application_id
    app_secret = gitlab_application.grafana[0].secret
  }
  sensitive = true
}

```

Gitlab API has limitations that don't allow to update the existing application, thus any property changes will force deleting and creating new one.

PR #417, opened 2020-09-16
Add gitlab_application resource

raphink:


armsnyder:
closes #417 
duplicate of #366

gitlab/resource_gitlab_application.go
```diff
@@ -0,0 +1,137 @@
...
+	options := &gitlab.CreateApplicationOptions{
+		Name:         gitlab.String(d.Get("name").(string)),
+		RedirectURI:  gitlab.String(d.Get("redirect_uri").(string)),
+		Scopes:       gitlab.String(d.Get("scopes").(string)),
+		Confidential: gitlab.Bool(d.Get("confidential").(bool)),
+	}
+
+	log.Printf("[DEBUG] Create GitLab application %s", *options.Name)
+
+	application, _, err = client.Applications.CreateApplication(options)
```
armsnyder, line 71: nitpick: you could use `:=` here and remove the `var` declarations earlier, to keep the vars closer to where you use them.

The same comment applies to your create and update functions too when you call API functions.

gitlab/resource_gitlab_application.go
```diff
@@ -0,0 +1,137 @@
...
+		return err
+	}
+
+	d.SetId(fmt.Sprintf("%d", application.ID))
+
+	d.Set("application_id", application.ApplicationID)
+	// Secret is only available on creation
+	d.Set("secret", application.Secret)
+
+	return nil
```
armsnyder, line 83: Best practice is to return the read function at the end of your create, update, and delete functions. This makes sure that all attributes are set, like `redirect_uri` and `confidential` too. ([Example from the Hashicorp docs](https://learn.hashicorp.com/tutorials/terraform/provider-update#implement-update).)

When you do this you can also remove the `d.Set("application_id", application.ApplicationID)` line above.

gitlab/resource_gitlab_application.go
```diff
@@ -0,0 +1,137 @@
...
+	for _, app := range applications {
+		if app.ID == applicationID {
+			d.Set("name", app.ApplicationName)
+			d.Set("application_id", app.ApplicationID)
+			d.Set("redirect_uri", app.CallbackURL)
+			d.Set("confidential", app.Confidential)
+		}
+	}
+
+	return nil
```
armsnyder, line 111: The standard practice is to set the ID to "" if the resource is not found. This tells Terraform that the resource doesn't exist anymore, so if it gets removed out-of-band, the user won't have to manually deal with the drift.

```suggestion
			d.Set("confidential", app.Confidential)
			return nil
		}
	}

	d.SetId("")
	return nil
```

gitlab/resource_gitlab_application.go
```diff
@@ -0,0 +1,137 @@
...
+	log.Printf("[DEBUG] Delete GitLab application %d", applicationID)
+	response, err = client.Applications.DeleteApplication(applicationID)
+
+	if err != nil {
+		return err
+	}
+
+	// StatusNoContent = 204
+	// Success with no body
+	if response.StatusCode != http.StatusNoContent {
```
armsnyder, line 132: Do you need this check? Doesn't `DeleteApplication` return an error if the delete failed?

gitlab/resource_gitlab_application.go
```diff
@@ -0,0 +1,137 @@
...
+	client := meta.(*gitlab.Client)
+	applicationID, err := strconv.Atoi(d.Id())
+	if err != nil {
+		return err
+	}
+
+	var applications []*gitlab.Application
+
+	log.Printf("[DEBUG] Read GitLab application %d", applicationID)
+	applications, _, err = client.Applications.ListApplications(nil)
```
armsnyder, line 96: This needs to handle pagination. See an example [here](https://github.com/gitlabhq/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_project_variable.go#L227).

gitlab/resource_gitlab_application_test.go
```diff
@@ -0,0 +1,133 @@
...
+func TestAccGitlabApplication_basic(t *testing.T) {
+	var application gitlab.Application
+	rInt := acctest.RandInt()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckGitlabApplicationDestroy,
+		Steps: []resource.TestStep{
+			{
```
armsnyder, line 24: Can you add another TestStep with different config, to test what happens when the attributes are changed?

armsnyder (changes requested):
Thanks for the contribution!

gitlab/resource_gitlab_application.go
```diff
@@ -0,0 +1,137 @@
...
+		Read:   resourceGitlabApplicationRead,
+		Delete: resourceGitlabApplicationDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"redirect_uri": {
```
armsnyder, line 25: What do you think about changing the `Type` of `redirect_uri` and `scopes` to `schema.TypeList`, and handling the comma-joining for the user with `strings.Join(...`?

gitlab/resource_gitlab_application.go
```diff
@@ -0,0 +1,137 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"scopes": {
+				Type:     schema.TypeString,
+				Optional: true,
```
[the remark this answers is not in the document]
raphink: The example request in the API docs sends an empty `scopes` parameter, so I assumed it worked:

https://docs.gitlab.com/ee/api/applications.html#create-an-application

```
curl --request POST --header "PRIVATE-TOKEN: <your_access_token>" --data "name=MyApplication&redirect_uri=http://redirect.uri&scopes=" "https://gitlab.example.com/api/v4/applications"
```

gitlab/resource_gitlab_application.go
```diff
@@ -0,0 +1,137 @@
...
+	var application *gitlab.Application
+
+	options := &gitlab.CreateApplicationOptions{
+		Name:         gitlab.String(d.Get("name").(string)),
+		RedirectURI:  gitlab.String(d.Get("redirect_uri").(string)),
+		Scopes:       gitlab.String(d.Get("scopes").(string)),
+		Confidential: gitlab.Bool(d.Get("confidential").(bool)),
+	}
+
+	log.Printf("[DEBUG] Create GitLab application %s", *options.Name)
```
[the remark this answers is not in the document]
raphink: I don't mind changing this, but I copied this from the `deploy_token` resource type to keep the same style within the provider code.

gitlab/resource_gitlab_application.go
raphink: Style copied from `deploy_token` resource

gitlab/resource_gitlab_application.go
raphink: Style copied from `deploy_token` resource

gitlab/resource_gitlab_application.go
raphink: Style copied from `deploy_token` resource 😉

gitlab/resource_gitlab_application.go
raphink: Good point, copied this from `deploy_token` resource ;-)

gitlab/resource_gitlab_application.go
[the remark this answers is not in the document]
armsnyder: Generally in this provider I see more %q than %s but I haven't counted.

gitlab/resource_gitlab_application.go
armsnyder: I think `deploy_token` is the outlier

gitlab/resource_gitlab_application.go
armsnyder: This is the design we're moving this provider toward, however. See #318 for more info about it.

gitlab/resource_gitlab_application.go
armsnyder: Most resources in this provider do it the way I described though.

gitlab/resource_gitlab_application.go
raphink: OK then, I'll fix this one.

I'll also try and get some ideas from #366. Some of the implementations there look better than mine (e.g. lists for scopes and redirect_uri).

gitlab/resource_gitlab_application.go
raphink: Yep, I'm in favor of doing this. I'll take it from #366 as mentioned above.

armsnyder:
PR feedback is still awaiting some changes from @raphink if they are still able to work on it. 😄

PR #416, opened 2020-09-16
Update the push rule if it exists already

ziporah:
This patch fixes #352 
if the Push rule creation throws 422 => error already exists on 
```
POST /api/v4/projects/21163411/push_rule HTTP/1.1  
...
---[ RESPONSE ]--------------------------------------
HTTP/2.0 422 Unprocessable Entity
...
{
 "error": "Project push rule exists"
}
...
```
We do a put update
```
PUT /api/v4/projects/21163411/push_rule HTTP/1.1 
...
---[ RESPONSE ]--------------------------------------
HTTP/2.0 200 OK
...
```
Note: I believe the test would have to be adjusted to detect this issue, but I don't have a play environment available to validate this.

PR #407, opened 2020-09-10
Remove non existing parameter from push rules docs

yeldiRium:
The parameter `commit_committer_check` does not exist in the current version of the project push rules resource: https://github.com/gitlabhq/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_project_push_rules.go

PR #422, opened 2020-09-19
Move gitlab_project_push_rules to gitlab_project.push_rules

### LINKED ISSUE #352: 422 {error: Project push rule exists}
### Terraform Version
Terraform v0.12.27
+ provider.gitlab v2.10.0

### Affected Resource(s)
- gitlab_project_push_rules

```hcl
resource "gitlab_project_push_rules" "project_push_rules" {
  project                       = gitlab_project.project.id
  commit_message_regex          = var.PROJECT_PUSH_RULES_COMMIT_MESSAGE_REGEX
  branch_name_regex             = var.PROJECT_PUSH_RULES_BRANCH_NAME_REGEX
  author_email_regex            = var.PROJECT_PUSH_RULES_AUTHOR_EMAIL_REGEX
  file_name_regex               = var.PROJECT_PUSH_RULES_FILE_NAME_REGEX
  deny_delete_tag               = var.PROJECT_PUSH_RULES_DENY_DELETE_TAG
  member_check                  = var.PROJECT_PUSH_RULES_MEMBER_CHECK
  prevent_secrets               = var.PROJECT_PUSH_RULES_PREVENT_SECRETS
  max_file_size                 = var.PROJECT_PUSH_RULES_MAX_FILE_SIZE
  # commit_committer_check        = var.PROJECT_PUSH_RULES_COMMIT_COMMITTER_CHECK
  # commit_message_negative_regex = var.PROJECT_PUSH_RULES_COMMIT_MESSAGE_NEGATIVE_REGEX
}

output project_push_rules {
  value = gitlab_project_push_rules.project_push_rules
}

variable "PROJECT_PUSH_RULES_COMMIT_MESSAGE_REGEX" {
  description = "(Optional, string) All commit messages must match this regex"
  type        = string
  default     = null
}

variable "PROJECT_PUSH_RULES_COMMIT_MESSAGE_NEGATIVE_REGEX" {
  type    = string
  default = null
}

variable "PROJECT_PUSH_RULES_BRANCH_NAME_REGEX" {
  description = "(Optional, string) All branch names must match this regex"
  type        = string
  default     = null
}

variable "PROJECT_PUSH_RULES_AUTHOR_EMAIL_REGEX" {
  description = "(Optional, string) All commit author emails must match this regex"
  type        = string
  default     = null
}

variable "PROJECT_PUSH_RULES_FILE_NAME_REGEX" {
  description = "(Optional, string) All commited filenames must not match this regex"
  type        = string
  default     = null
}

variable "PROJECT_PUSH_RULES_DENY_DELETE_TAG" {
  description = "(Optional, bool) Deny deleting a tag"
  type        = bool
  default     = false
}

variable "PROJECT_PUSH_RULES_MEMBER_CHECK" {
  description = "(Optional, bool) Restrict commits by author (email) to existing GitLab users"
  type        = bool
  default     = false
}

variable "PROJECT_PUSH_RULES_PREVENT_SECRETS" {
  description = "(Optional, bool) GitLab will reject any files that are likely to contain secrets"
  type        = bool
  default     = true
}

variable "PROJECT_PUSH_RULES_MAX_FILE_SIZE" {
  description = "(Optional, int) Maximum file size (MB)"
  type        = number
  default     = "0"
}

variable "PROJECT_PUSH_RULES_COMMIT_COMMITTER_CHECK" {
  description = "(Optional, bool) Users can only push commits to this repository that were committed with one of their own verified emails"
  type        = bool
  default     = false
}
```

### Debug Output
```hcl
  # gitlab_project_push_rules.project_push_rules will be created
  + resource "gitlab_project_push_rules" "project_push_rules" {
      + deny_delete_tag = false
      + id              = (known after apply)
      + max_file_size   = 0
      + member_check    = false
      + prevent_secrets = true
      + project         = "REDACTED"
    }

Error: POST https://gitlab.com/api/v4/projects/REDACTED/push_rule: 422 {error: Project push rule exists}

  on main.tf line 62, in resource "gitlab_project_push_rules" "project_push_rules":
  62: resource "gitlab_project_push_rules" "project_push_rules" {

```
### Expected Behavior
for a rule to be created

### Actual Behavior
error: Project push rule exists

### Steps to Reproduce
1. `terraform apply`

not related, but i'm seeing
```hcl
An argument named "commit_committer_check" is not expected here.
```
while that element exists in the documentation

armsnyder:
Supersedes #416 #407 #269 #246 
Fixes #352 

See the discussion in #416 for some background.

The `gitlab_project_push_rules` resource has caused a lot of confusion because in GitLab EE, push rules are created implicitly when projects are created. This created a situation where if you wanted to use `gitlab_project_push_rules` you would always have to import it, and you could not create a project with push rules in the same `terraform apply`.

The solution in the PR is to remove the `gitlab_project_push_rules` resource, and add a `push_rules` attribute to the `gitlab_project` resource and data source.

I took care to add tests for adding, updating, and removing push rules in both EE and CE.

Note that `push_rules` is Optional+Computed, meaning that if the user sets one of the arguments, then removes the block entirely, it will not be removed from the upstream GitLab.


docs/data-sources/project.md
```diff
@@ -71,4 +71,30 @@ The following attributes are exported:
 
 * `remove_source_branch_after_merge` - Enable `Delete source branch` option by default for all new merge requests
 
-* `packages_enabled` - Enable packages repository for the project.
\ No newline at end of file
+* `packages_enabled` - Enable packages repository for the project.
+
+* `push_rules` Project push rules (GitLab Starter/Bronze/EE only). Push Rules documented below.
+
+Push Rules (`push_rules`) is exported with the following attributes:
```
roidelapluie, line 78: Can we push that to a dedicated section?

docs/data-sources/project.md
```diff
@@ -71,4 +71,30 @@ The following attributes are exported:
 
 * `remove_source_branch_after_merge` - Enable `Delete source branch` option by default for all new merge requests
 
-* `packages_enabled` - Enable packages repository for the project.
\ No newline at end of file
+* `packages_enabled` - Enable packages repository for the project.
+
+* `push_rules` Project push rules (GitLab Starter/Bronze/EE only). Push Rules documented below.
```
roidelapluie, line 76: Can we just say EE maybe? We don't give that level of details elsewhere IIRC.

Also can we link to upstream doc?

docs/resources/project.md
```diff
@@ -80,6 +98,34 @@ The following arguments are supported:
 
 * `packages_enabled` - (Optional) Enable packages repository for the project.
 
+* `push_rules` (Optional) Manage push rules for your GitLab project.
```
roidelapluie, line 101: This is duplicated?

gitlab/resource_gitlab_project.go
```diff
@@ -11,6 +13,8 @@ import (
 	gitlab "github.com/xanzy/go-gitlab"
 )
 
+var errProjectPushRulesUnsupported = errors.New("Project push rules are not supported in your version of GitLab")
```
roidelapluie, line 16:
```suggestion
var errProjectPushRulesUnsupported = errors.New("project push rules are not supported in your version of GitLab")
```

gitlab/resource_gitlab_project.go
```diff
@@ -191,6 +195,83 @@ var resourceGitLabProjectSchema = map[string]*schema.Schema{
...
+				},
+				"commit_message_negative_regex": {
+					Type:         schema.TypeString,
+					Optional:     true,
+					ValidateFunc: validation.StringIsValidRegExp,
+				},
+				"file_name_regex": {
+					Type:         schema.TypeString,
+					Optional:     true,
+					ValidateFunc: validation.StringIsValidRegExp,
```
roidelapluie, line 227: That validation method: are the regex of Go exactly overlapping with Ruby regexes??

docs/data-sources/project.md
armsnyder: I changed this line into a header. Is is better?

docs/data-sources/project.md
armsnyder: I searched our docs for "EE" and the only other instance of it I found was in [service_github](https://github.com/gitlabhq/terraform-provider-gitlab/blob/master/docs/resources/service_github.md)
> requires either EE (self-hosted) or Silver and above (GitLab.com).

So in that case it's equally specific. I'm not sure if there are other EE features in the provider that we simply do not mention are EE only? In that case I can remove the mention of EE in the doc.

@nicholasklick You mentioned people were asking for EE features in this provider. Do you think it makes sense for us to document which fields are EE-only, or can we leave that out for simplicity?

docs/resources/project.md
armsnyder: Nope, this page is the resource. The earlier one was the data source.

gitlab/resource_gitlab_project.go
armsnyder: In other providers I have seen them use capitalized errors for errors that are returned from the CRUD functions, because they are user-facing. I moved this error back to inline to ease the confusion. Let me know if you still want me to change it.

roidelapluie:
an we document like this:

https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/api_service

with "Nested blocks"

gitlab/resource_gitlab_project.go
armsnyder: I thought only Javascript was different? Do you want me to remove validation here?

docs/data-sources/project.md
armsnyder: Updated to use the "Nested Blocks" format.

docs/data-sources/project.md
armsnyder: In my latest push I removed mention of "EE" in the docs change.

armsnyder:
@roidelapluie So you're aware, I changed `push_rules` to be Optional+Computed. This allowed me to remove the `DiffSuppressFunc`. It also fixed a minor warning in the TF debug logs when using `push_rules`:

```
2020/09/19 12:23:26 [WARN] Provider "registry.terraform.io/-/gitlab" produced an unexpected new value for gitlab_project.this, but we are tolerating it because it is using the legacy plugin SDK.
    The following problems may be the cause of any confusing errors from downstream operations:
      - .push_rules: block count changed from 0 to 1
```

The side effect of O+C is if a user removes their `push_rules` block it will not remove the push rules from upstream. However I found multiple examples of O+C on nested blocks in the aws-terraform-provider.

docs/data-sources/project.md
armsnyder: I also added the doc link to both the data source and the resource. 😉

gitlab/resource_gitlab_project.go
roidelapluie: no, golang is opiniated.

e.g. they do not support backreference, whereas ruby does.

armsnyder: Good to know. I've removed the validation.

roidelapluie (approved):
Awesome!

Issue #427, opened 2020-09-21
Opportunities for improving code reviews


armsnyder:
I spent some time cataloguing PR feedback and bug fixes just from 2020, looking for trends that we could address either with custom analyzers ([like terraform-provider-aws has done](https://github.com/terraform-providers/terraform-provider-aws/tree/master/awsproviderlint)), PR template checklists, or contributing guidelines. The goal would be to increase code velocity and confidence. I imagine some of these may be _too_ opinionated, so am interested in feedback. One theme is incorrect documentation. Another theme is failing to match the upstream GitLab API, and I wonder if some kind of code generation or at least stubbing would be possible. Anyway, the list:

### Code

Things that could probably be caught with [go/analysis](https://godoc.org/golang.org/x/tools/go/analysis) if we want to invest there.

**High value and simple:**

- [001] [Design/Debt] Read functions should only depend on `d.Id()` for calling go-gitlab functions. [\[Example\]](https://github.com/gitlabhq/terraform-provider-gitlab/pull/350#discussion_r487443866) [\[Example\]](https://github.com/gitlabhq/terraform-provider-gitlab/pull/375#discussion_r467358368) [\[Example\]](https://github.com/gitlabhq/terraform-provider-gitlab/pull/375#discussion_r467549781) [\[Example\]](https://github.com/gitlabhq/terraform-provider-gitlab/pull/358#issuecomment-667779118)
- [002] [Bug] Use pagination when using the `List*` go-gitlab functions. [\[Example\]](https://github.com/gitlabhq/terraform-provider-gitlab/pull/417#discussion_r490055657) [\[Example\]](https://github.com/gitlabhq/terraform-provider-gitlab/pull/367#issuecomment-667617388) [\[Example\]](https://github.com/gitlabhq/terraform-provider-gitlab/pull/229)
- [003] [Design] Use either the `Exists` schema function or have `d.SetId("")` in the read function to handle resources deleted upstream. [\[Example\]](https://github.com/gitlabhq/terraform-provider-gitlab/pull/417#discussion_r490052287) [\[Example\]](https://github.com/gitlabhq/terraform-provider-gitlab/pull/358#discussion_r456755037) [Example\]](https://github.com/gitlabhq/terraform-provider-gitlab/pull/240)
[...]

PR #432, opened 2020-09-23
New gitlabproviderlint tool with first docs check

armsnyder:
### Overview

It would be neat if we could automate the most common feedback given in code reviews. In #427 I identified incorrect documentation as one of the big areas, so I thought I'd start there.

This PR adds a new tool called `gitlabproviderlint`. It is similar in design to [awsproviderlint](https://github.com/terraform-providers/terraform-provider-aws/tree/master/awsproviderlint) in that it uses the [go/analysis](https://godoc.org/golang.org/x/tools/go/analysis) standard framework for writing static checkers, and it extends the wonderful set of checkers in [tfproviderlint](https://github.com/bflad/tfproviderlint).

In addition to the `tfproviderlint` checkers, the first custom check added in this PR is `GD001: check for resources without a docs page`.

I opened #431 as an example where you can see it in action.

Screen Shot 2020-09-22 at 4 50 52 PM

### Notes

* Excuse the big vendor diff. 😄
* `gitlabproviderlint/GD001.go` is the main code for the custom analyzer.
* `gitlabproviderlint/parsedocs.go` and `gitlabproviderlint/providerindex.go` are analyzers that don't report any issues but take care of parsing the docs markdown and enumerating the resources in the provider. `GD001` depends on the output of these. Parsing markdown is a premature change for this PR, but it lays the groundwork for future docs-related analyzers.
* In the `gitlabproviderlint` make target I had to disable several of checkers from `tfproviderlint`, because they were picking up issues in the existing code. I'll open new issues to get them addressed.


timofurrer:
@armsnyder is this still a thing? 👀

armsnyder:
Hah, no, this is not needed with SDK V2 because we can not automatically generate documentation. I'm working on that as my next PR for #543.

Issue #441, opened 2020-09-23
gitlab_branch_protection should ignore individual users


jeremad:
### Terraform Version
```
Terraform v0.12.29
+ provider.gitlab v2.11.0
```
### Affected Resource(s)
Please list the resources as a list, for example:
- gitlab_branch_protection

### Expected Behavior

Giving individual push access to protected branch is a premium feature of GitLab, the provider does not implement that yet.
If an individual user has push_access, the GitLab API returns:
```
"push_access_levels":[{
  "access_level": 40,
  "access_level_description": "Tanker Sysadmin",
  "user_id": 1706972,
  "group_id": null
},{
  "access_level": 0,
  "access_level_description": "No one",
  "user_id": null,
  "group_id": null
}]
```
And this is bugged: https://github.com/gitlabhq/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_branch_protection.go#L120

We should ignore push_access_levels and merge_access_level for individual users at least while the API is not implemented in the provider

mattkasa:
I think this has been fixed by #556, but please feel free to reopen if I am mistaken 👍

jeremad:
thx

PR #442, opened 2020-09-23
fix: do not automatically bind the first thing you see

jeremad:
We should ignore merge_access and push_access of individual users and
groups. This feature is not yet implemented in the provider.

Closes: #441

### DIFF
--- a/gitlab/resource_gitlab_branch_protection.go
+++ b/gitlab/resource_gitlab_branch_protection.go
@@ -116,8 +116,16 @@ func resourceGitlabBranchProtectionRead(d *schema.ResourceData, meta interface{}
 
 	d.Set("project", project)
 	d.Set("branch", pb.Name)
-	d.Set("merge_access_level", accessLevel[pb.MergeAccessLevels[0].AccessLevel])
-	d.Set("push_access_level", accessLevel[pb.PushAccessLevels[0].AccessLevel])
+	for _, mergeAccessLevel := range pb.MergeAccessLevels {
+		if &mergeAccessLevel.UserID == nil && &mergeAccessLevel.GroupID == nil {
+			d.Set("merge_access_level", accessLevel[mergeAccessLevel.AccessLevel])
+		}
+	}
+	for _, pushAccessLevel := range pb.PushAccessLevels {
+		if &pushAccessLevel.UserID == nil && &pushAccessLevel.GroupID == nil {
+			d.Set("merge_access_level", accessLevel[pushAccessLevel.AccessLevel])
+		}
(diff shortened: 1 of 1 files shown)


jeremad:
BTW I cannot really add tests since this is a bug triggered by using a feature that is not implemented in the provider

roidelapluie:
Thanks for the contribution.

I disagree with the approach. I would prefer the user to use  lifecycle { ignore_changes } if they use unsupported features, and get proper support for it in the provider.

roidelapluie:
```
lifecycle {
  lifecycle {
    ignore_changes = [
merge_access_level,
push_access_level,
    ]
}

File: gitlab/resource_gitlab_branch_protection.go [hunk 1]
```diff
@@ -116,8 +116,16 @@ func resourceGitlabBranchProtectionRead(d *schema.ResourceData, meta interface{}
...
 	d.Set("project", project)
 	d.Set("branch", pb.Name)
-	d.Set("merge_access_level", accessLevel[pb.MergeAccessLevels[0].AccessLevel])
-	d.Set("push_access_level", accessLevel[pb.PushAccessLevels[0].AccessLevel])
+	for _, mergeAccessLevel := range pb.MergeAccessLevels {
+		if &mergeAccessLevel.UserID == nil && &mergeAccessLevel.GroupID == nil {
+			d.Set("merge_access_level", accessLevel[mergeAccessLevel.AccessLevel])
+		}
+	}
+	for _, pushAccessLevel := range pb.MergeAccessLevels {
```
roidelapluie, line 124: be careful if you run a fork of this this internally, this should be PushAccessLevels

jeremad:
fair enough.
I was going to implement the feature. But it is a bit tricky to bind the API terraformly: https://github.com/xanzy/go-gitlab/blob/master/protected_branches.go#L49

gitlab/resource_gitlab_branch_protection.go
```diff
@@ -116,8 +116,16 @@ func resourceGitlabBranchProtectionRead(d *schema.ResourceData, meta interface{}
...
+		if &pushAccessLevel.UserID == nil && &pushAccessLevel.GroupID == nil {
+			d.Set("merge_access_level", accessLevel[pushAccessLevel.AccessLevel])
```
roidelapluie, line 126: Same here, push_access_level

jeremad: oopsy

jeremad:
I am closing this one then

PR #449, opened 2020-09-29
Project mirroring

cnuss:
Hi there!

I'd like to be able to use Terraform to set up a Pull mirror on GitLab, and I found that the API had the functionality but this provider did not:
https://gitlab.com/gitlab-org/gitlab/-/issues/230816#note_398470032

Specifically: `mirror` and `mirror_trigger_builds`.

A reasonable configuration with these options is: 

```
data "gitlab_group" "group" {
  full_path = "mygroup"
}

resource "gitlab_project" "github_mirror" {
  name         = "myproject"
  namespace_id = data.gitlab_group.group.id

  import_url            = "https://github.com/mygroup/myproject.git"
  mirror                = true
  mirror_trigger_builds = true
}
```

This adds those two options to this provider.  I mainly followed the pattern of other top-level booleans in the `gitlab_project` resource is managed.

Please let me know if there's more to do here!

### DIFF
--- a/gitlab/resource_gitlab_project.go
+++ b/gitlab/resource_gitlab_project.go
@@ -268,6 +268,14 @@ var resourceGitLabProjectSchema = map[string]*schema.Schema{
 		Type:     schema.TypeInt,
 		Optional: true,
 	},
+	"mirror": {
+		Type:     schema.TypeBool,
+		Optional: true,
+	},
+	"mirror_trigger_builds": {
+		Type:     schema.TypeBool,
+		Optional: true,
+	},
 }
 
 func resourceGitlabProject() *schema.Resource {
@@ -313,6 +321,8 @@ func resourceGitlabProjectSetToState(d *schema.ResourceData, project *gitlab.Pro
 	d.Set("archived", project.Archived)
 	d.Set("remove_source_branch_after_merge", project.RemoveSourceBranchAfterMerge)
 	d.Set("packages_enabled", project.PackagesEnabled)
+	d.Set("mirror", project.Mirror)
+	d.Set("mirror_trigger_builds", project.MirrorTriggerBuilds)
 }
 
 func resourceGitlabProjectCreate(d *schema.ResourceData, meta interface{}) error {
@@ -336,6 +346,8 @@ func resourceGitlabProjectCreate(d *schema.ResourceData, meta interface{}) error
 		SharedRunnersEnabled:                      gitlab.Bool(d.Get("shared_runners_enabled").(bool)),
 		RemoveSourceBranchAfterMerge:              gitlab.Bool(d.Get("remove_source_branch_after_merge").(bool)),
 		PackagesEnabled:                           gitlab.Bool(d.Get("packages_enabled").(bool)),
+		Mirror:                                    gitlab.Bool(d.Get("mirror").(bool)),
+		MirrorTriggerBuilds:                       gitlab.Bool(d.Get("mirror_trigger_builds").(bool)),
 	}
 
 	if v, ok := d.GetOk("path"); ok {
@@ -556,6 +568,14 @@ func resourceGitlabProjectUpdate(d *schema.ResourceData, meta interface{}) error
 
 	if d.HasChange("packages_enabled") {
 		options.PackagesEnabled = gitlab.Bool(d.Get("packages_enabled").(bool))
+	}
+
+	if d.HasChange("mirror") {
+		options.Mirror = gitlab.Bool(d.Get("mirror").(bool))
+	}
+
+	if d.HasChange("mirror_trigger_builds") {
+		options.MirrorTriggerBuilds = gitlab.Bool(d.Get("mirror_trigger_builds").(bool))
 	}
 
 	if *options != (gitlab.EditProjectOptions{}) {
--- a/gitlab/resource_gitlab_project.go
+++ b/gitlab/resource_gitlab_project.go
@@ -589,10 +589,16 @@ func resourceGitlabProjectUpdate(d *schema.ResourceData, meta interface{}) error
 	}
 
 	if d.HasChange("mirror") {
+		// It appears that GitLab API requires that import_url is also set when `mirror` is updated/changed
+		// Ref: https://github.com/gitlabhq/terraform-provider-gitlab/pull/449#discussion_r549729230
+		options.ImportURL = gitlab.String(d.Get("import_url").(string))
 		options.Mirror = gitlab.Bool(d.Get("mirror").(bool))
 	}
 
 	if d.HasChange("mirror_trigger_builds") {
+		// It appears that GitLab API requires that import_url is also set when `mirror_trigger_builds` is updated/changed
+		// Ref: https://github.com/gitlabhq/terraform-provider-gitlab/pull/449#discussion_r549729230
+		options.ImportURL = gitlab.String(d.Get("import_url").(string))
 		options.MirrorTriggerBuilds = gitlab.Bool(d.Get("mirror_trigger_builds").(bool))
 	}
 
--- a/gitlab/resource_gitlab_project.go
+++ b/gitlab/resource_gitlab_project.go
@@ -329,6 +329,8 @@ func resourceGitlabProjectSetToState(d *schema.ResourceData, project *gitlab.Pro
 	d.Set("remove_source_branch_after_merge", project.RemoveSourceBranchAfterMerge)
 	d.Set("packages_enabled", project.PackagesEnabled)
 	d.Set("pages_access_level", string(project.PagesAccessLevel))
+	d.Set("mirror", project.Mirror)
+	d.Set("mirror_trigger_builds", project.MirrorTriggerBuilds)
 }
 
 func resourceGitlabProjectCreate(d *schema.ResourceData, meta interface{}) error {
@@ -582,6 +584,14 @@ func resourceGitlabProjectUpdate(d *schema.ResourceData, meta interface{}) error
 
 	if d.HasChange("pages_access_level") {
 		options.PagesAccessLevel = stringToAccessControlValue(d.Get("pages_access_level").(string))
+	}
+
+	if d.HasChange("mirror") {
+		options.Mirror = gitlab.Bool(d.Get("mirror").(bool))
+	}
+
+	if d.HasChange("mirror_trigger_builds") {
+		options.MirrorTriggerBuilds = gitlab.Bool(d.Get("mirror_trigger_builds").(bool))
 	}
 
 	if *options != (gitlab.EditProjectOptions{}) {
--- a/gitlab/resource_gitlab_project.go
+++ b/gitlab/resource_gitlab_project.go
@@ -278,12 +278,12 @@ var resourceGitLabProjectSchema = map[string]*schema.Schema{
 	"mirror": {
 		Type:     schema.TypeBool,
 		Optional: true,
-		Default: false,
+		Default:  false,
 	},
 	"mirror_trigger_builds": {
 		Type:     schema.TypeBool,
 		Optional: true,
-		Default: false,
+		Default:  false,
 	},
 }
 
--- a/gitlab/resource_gitlab_project.go
+++ b/gitlab/resource_gitlab_project.go
@@ -278,10 +278,12 @@ var resourceGitLabProjectSchema = map[string]*schema.Schema{
 	"mirror": {
 		Type:     schema.TypeBool,
 		Optional: true,
+		Default: false,
 	},
 	"mirror_trigger_builds": {
 		Type:     schema.TypeBool,
 		Optional: true,
+		Default: false,
 	},
 }
 
--- a/gitlab/resource_gitlab_project_test.go
+++ b/gitlab/resource_gitlab_project_test.go
@@ -566,6 +566,7 @@ func TestAccGitlabProject_importURLMirrored(t *testing.T) {
 			{
 				// First, import, as mirrored
 				Config: testAccGitlabProjectConfigImportURLMirror(rInt, baseProject.HTTPURLToRepo),
+				SkipFunc: isRunningInCE,
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckGitlabProjectExists("gitlab_project.imported", &mirror),
 					resource.TestCheckResourceAttr("gitlab_project.imported", "import_url", baseProject.HTTPURLToRepo),
@@ -589,6 +590,7 @@ func TestAccGitlabProject_importURLMirrored(t *testing.T) {
 			{
 				// Second, disable mirroring, using the original ImportURL acceptance test
 				Config: testAccGitlabProjectConfigImportURLMirrorDisabled(rInt, baseProject.HTTPURLToRepo),
+				SkipFunc: isRunningInCE,
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckGitlabProjectExists("gitlab_project.imported", &mirror),
 					resource.TestCheckResourceAttr("gitlab_project.imported", "import_url", baseProject.HTTPURLToRepo),
--- a/gitlab/resource_gitlab_project_test.go
+++ b/gitlab/resource_gitlab_project_test.go
@@ -545,7 +545,7 @@ func TestAccGitlabProject_importURLMirrored(t *testing.T) {
 		CheckDestroy: testAccCheckGitlabProjectDestroy,
 		Steps: []resource.TestStep{
 			{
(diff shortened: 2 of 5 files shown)


cnuss:
hi @nicholasklick my first PR here! any other steps I should do on this PR for y'all?

cnuss:
hey @armsnyder pinging you as well! could you help me with any necessary steps to get this merged and released?

also do you have an insight into to what i did wrong to make the acceptance tests fail?

cnuss:
@nicholasklick @armsnyder hey there! any feedback? am I barking up the wrong tree?

cnuss:
@roidelapluie hiya! i've seen you comment on some previous PRs, can you help me push this one through?

svenstaro:
Wouldn't it make sense for this PR to also do something with the existing `project_mirror` resource? See https://registry.terraform.io/providers/gitlabhq/gitlab/latest/docs/resources/project_mirror

Having that support a `direction` would be neat and in scope for this PR I think.

svenstaro:
Could you also add `only_mirror_protected_branches` and `mirror_overwrites_diverged_branches`?

cnuss:
> Could you also add `only_mirror_protected_branches` and `mirror_overwrites_diverged_branches`?

@svenstaro  Yep! Just took another look at the API and I think I should add `mirror_user_id` as well? On another note, could you perhaps give me some guidance on why the Acceptance Tests are failing?

svenstaro:
> > Could you also add `only_mirror_protected_branches` and `mirror_overwrites_diverged_branches`?
> 
> @svenstaro Yep! Just took another look at the API and I think I should add `mirror_user_id` as well? On another note, could you perhaps give me some guidance on why the Acceptance Tests are failing?

Oh yeah, I forgot about `mirror_user_id`. And no idea about the acceptance test. A maintainer would need to look into it.

svenstaro:
However, perhaps it's something on your side after all: https://github.com/gitlabhq/terraform-provider-gitlab/pull/449/checks?check_run_id=1184195324#step:5:180

cnuss:
> However, perhaps it's something on your side after all: https://github.com/gitlabhq/terraform-provider-gitlab/pull/449/checks?check_run_id=1184195324#step:5:180

Yea i saw that too! However it wasn't clear how to make it "expect" certain properties on the resource 🤔 ... how do I find who the maintainers are here for help? (I've mentioned a few people but haven't heard anything back), pinging @richardc as the original author!

cnuss:
@svenstaro Since you asked about adding only_mirror_protected_branches and mirror_overwrites_diverged_branches, I noticed something interesting…

The Project API only has those settings on the [Edit Project API](https://docs.gitlab.com/ee/api/projects.html#edit-project), but not the [Create Project API](https://docs.gitlab.com/ee/api/projects.html#edit-project). 

I've confirmed that's also reflected the same way on the [go-gitlab](https://github.com/xanzy/go-gitlab/blob/master/projects.go#L465) project, so it doesn't appear to be a gap in the documentation.  

What's the pattern to follow here if those properties are set when creating a new project but it's not available on the original POST? Is there a subsequent PUT to immediately update the project after creation? Do you know if there's an example of this behavior I can follow to make sure its right?

svenstaro:
Sorry @cnuss, I'm really just an interested third party and have absolutely no idea about anything else concerning this provider or gitlab API. :P 

Maybe @armsnyder can help? I'm sure people are currently just busy and will take a look at this once there's time. I'd just really like to see this stuff go in. :)

armsnyder:
@cnuss

> Does there need to be a subsequent PUT to immediately update the project after creation?

Yes, the correct pattern would be to make the two API calls you mentioned, in the Create function. I also tried setting `only_mirror_protected_branches` and `mirror_overwrites_diverged_branches` fields in a POST call, and they were ignored.

The PR looks good! Could you add a test for creating a project with mirror=true? It can be mostly a copy of the `TestAccGitlabProject_importURL` test. The test should also have a step where mirror is changed from true to false.

cnuss:
thanks for the feedback @armsnyder !

sorry for the delay in my update, I'll get that updated ASAP! I'm a n00b to writing tests in Go

nicholasklick:
@cnuss Anything else you needed to help move this PR along?

svenstaro:
This needs a rebase.

cnuss:
Let me see if i can get this cleaned up and pushed over the finish line today! Sorry for the delay!

cnuss:
@nicholasklick PR updated, Acceptance Tests Passing. How we lookin' here? Does this look good?

cc: @armsnyder @svenstaro

svenstaro:
Awesome! I'd love to see this get merged.

cnuss:
> Awesome! I'd love to see this get merged.

Lol yea I created and placed the binary locally to get around the initial delays in getting this reviewed. FYI @armsnyder Would be helpful to have Contributing guidelines so I know the right process! Had to dig thru other PRs to see what was the pattern others followed! Also it appears I was unable to set a reviewer when creating the PR so its just dangling here with nobody on the hook.

svenstaro:
I think you should add some docs to the "other" mirroring which makes people aware of the differences between push and pull mirroring. For instance, if you go [here](https://registry.terraform.io/providers/gitlabhq/gitlab/latest/docs/resources/project_mirror), the text implies that it's push but it should be made very explicit and your pull mirroring feature should be mentioned.

brijshah:
My company actually needs this feature, hoping this will get merged soon! 👍

cnuss:
@brijshah updated docs! did I capture your concerns properly?

https://github.com/gitlabhq/terraform-provider-gitlab/pull/449/commits/fbceedfd3b38939672376d795b7ffd9e846d62f8

cnuss:
> I think you should add some docs to the "other" mirroring which makes people aware of the differences between push and pull mirroring. For instance, if you go [here](https://registry.terraform.io/providers/gitlabhq/gitlab/latest/docs/resources/project_mirror), the text implies that it's push but it should be made very explicit and your pull mirroring feature should be mentioned.

@svenstaro sorry mentioned the wrong person in my previous comment, I made doc updates for pull mirroring vs push mirroring. Let me know if I captured your concerns properly!

svenstaro:
Yeah, the difference is now obvious.

brijshah:
whats left to get this merged?

nicholasklick:
@cnuss just a final blessing from @armsnyder and then I would like to cut a new release for this.

gitlab/resource_gitlab_project_test.go
```diff
@@ -509,6 +509,64 @@ func TestAccGitlabProject_importURL(t *testing.T) {
...
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckGitlabProjectDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGitlabProjectConfigImportURLMirrored(rInt, baseProject.HTTPURLToRepo),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("gitlab_project.imported", "import_url", baseProject.HTTPURLToRepo),
+					resource.TestCheckResourceAttr("gitlab_project.mirrored", "mirrored", "true"),
+					resource.TestCheckResourceAttr("gitlab_project.mirror_target_builds", "mirror_target_builds", "true"),
+
```
armsnyder, line 553: The new test looks fishy; it has some assertions on resources that are not being created by the test config. Unfortunately we can't run the acceptance tests in the GitHub workflow until #495 and #494 are resolved.

```suggestion
```

cnuss: @armsnyder I'm not sure I follow...

I thought these test the values of the properties adding in this PR.

Is there a different way to check these are set on the resource properly or am I missing something?

The properties added are here:
https://github.com/cnuss/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_project_test.go#L904-L906

armsnyder: No worries. I was referring to the resource name, which is `gitlab_project.imported` (see: https://github.com/cnuss/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_project_test.go#L901). However in your `resource.TestCheckResourceAttr` call you're using a resource name of `gitlab_project.mirrored`. My suggestion to remove these three lines is because the Terraform test framework will fail automatically if the resulting attributes do not match what was configured.

cnuss: Ahh... then this would also be correct if I wanted to do an explicit check? Sounds like you have a preference for the implicit check?

```
					resource.TestCheckResourceAttr("gitlab_project.imported", "import_url", baseProject.HTTPURLToRepo),
					resource.TestCheckResourceAttr("gitlab_project.imported", "mirrored", "true"),
					resource.TestCheckResourceAttr("gitlab_project.imported", "mirror_target_builds", "true"),
```

armsnyder: Yup, either way is fine. I was just giving you the lazier option. 😄

cnuss: Pushed the non-lazy option ;) Awaiting #494 !

nicholasklick: @cnuss #494 was closed. Is this ready for merge?

nicholasklick: Can we resolve this conversation?

cnuss: Yup fixed, resolving convo

cnuss:
Syncing this branch with master real quick for one last preflight check

cnuss:
Debugging this now: 

```
=== RUN   TestAccGitlabProject_importURLMirrored
Error:     testing.go:684: Step 0 error: config is invalid: 2 problems:
        
        - Unsupported argument: An argument named "mirrored" is not expected here. Did you mean "mirror"?
        - Unsupported argument: An argument named "mirror_target_builds" is not expected here.
```

cnuss:
@armsnyder @nicholasklick Green checks across the board! Ready for merge

armsnyder (changes requested):
Looking at the [API docs](https://docs.gitlab.com/ee/api/projects.html#edit-project), these two attributes can be modified. Can you also update the `resourceGitlabProjectUpdate` and `resourceGitlabProjectSetToState` functions to update and set these new attributes?

cnuss:
> Looking at the [API docs](https://docs.gitlab.com/ee/api/projects.html#edit-project), these two attributes can be modified. Can you also update the `resourceGitlabProjectUpdate` and `resourceGitlabProjectSetToState` functions to update and set these new attributes?

Definitely! one moment...

gitlab/resource_gitlab_project_test.go
```diff
@@ -509,6 +509,86 @@ func TestAccGitlabProject_importURL(t *testing.T) {
...
+						return nil
+					},
+				),
+			},
+			{
+				// Second, disable mirroring, using the original ImportURL acceptance test
+				Config: testAccGitlabProjectConfigImportURL(rInt, baseProject.HTTPURLToRepo),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("gitlab_project.imported", "import_url", baseProject.HTTPURLToRepo),
+					resource.TestCheckResourceAttr("gitlab_project.imported", "mirror", "false"),
```
cnuss, line 572: @armsnyder is my explicit property check causing the test to fail? When it says `Check 2/4 error`, which check is actually failing?

```
=== RUN   TestAccGitlabProject_importURLMirrored
Error:     testing.go:684: Step 0 error: Check failed: Check 2/4 error: gitlab_project.imported: Attribute 'mirror' expected "true", got "false"
```

cnuss: Trying something, but shooting in the dark: https://github.com/gitlabhq/terraform-provider-gitlab/pull/449/commits/116e25f10739e1ea7f945706eb30fd4542ab4e0a

cnuss: @armsnyder i'm at a loss and i've burnt my whole day on this... the likely solution is to remove the explicit checks as you mentioned earlier, but I'm not seeing in the test cases how the properties would be checked implicitly.

Could you reference where/how implicit checks are made and I'll try to see if that makes the test cases happy?

cnuss:
@armsnyder @nicholasklick can either of you offer guidance on where i'm going wrong with the acceptance tests?

nicholasklick:
Maybe @btyy77c could assist here?

gitlab/resource_gitlab_project_test.go
sodre: @cnuss,
Looking at the [API documentation](https://docs.gitlab.com/ce/api/projects.html#configure-pull-mirroring-for-a-project), the pull-mirroring is only available in the Starter/Bronze configuration. 

Are you able to make your tests only run if in the Enterprise Edition pipeline?

btyy77c: > @cnuss,
> Looking at the [API documentation](https://docs.gitlab.com/ce/api/projects.html#configure-pull-mirroring-for-a-project), the pull-mirroring is only available in the Starter/Bronze configuration.
> 
> Are you able to make your tests only run if in the Enterprise Edition pipeline?

You can skip CE tests by adding `SkipFunc: isRunningInCE` to your test steps.  

Example:
```
{
  // First, import, as mirrored
  SkipFunc: isRunningInCE,
  Config:   testAccGitlabProjectConfigImportURLMirror(rInt, baseProject.HTTPURLToRepo),
  ...
},
```
This means the tests will only run for the EE acceptance tests.

gitlab/resource_gitlab_project.go
```diff
@@ -574,6 +588,14 @@ func resourceGitlabProjectUpdate(d *schema.ResourceData, meta interface{}) error
 		options.PagesAccessLevel = stringToAccessControlValue(d.Get("pages_access_level").(string))
 	}
 
+	if d.HasChange("mirror") {
+		options.Mirror = gitlab.Bool(d.Get("mirror").(bool))
+	}
+
+	if d.HasChange("mirror_trigger_builds") {
+		options.MirrorTriggerBuilds = gitlab.Bool(d.Get("mirror_trigger_builds").(bool))
+	}
```
btyy77c, line 597: When I run the tests locally, the create test is passing but the update test is failing. 

I've done some debugging and I don't think this is an issue with the Terraform Provider. When I curl my GitLab api with the params `{ mirror: false, mirror_trigger_builds: false }`, the api will also error out. 

The GitLab [api tests](https://gitlab.com/gitlab-org/gitlab/-/blob/master/ee/spec/requests/api/projects_spec.rb#L798) seem to always include the `import_url` field with their params.  I'll have to do more research to see if this is a bug, or if this was intentional. 

If your GitLab api is also failing, you may need to change this to something like:

```
if d.HasChange("mirror") {
  options.Mirror = gitlab.Bool(d.Get("mirror").(bool))
  options.ImportURL = gitlab.String(d.Get("import_url").(string))
}

if d.HasChange("mirror_trigger_builds") {
  options.MirrorTriggerBuilds = gitlab.Bool(d.Get("mirror_trigger_builds").(bool))
  options.ImportURL = gitlab.String(d.Get("import_url").(string))
}
```

cnuss: @btyy77c thank you for your awesome investigative work, i'll try your suggestion and report back!

gitlab/resource_gitlab_project_test.go
cnuss: @nicholasklick I pushed https://github.com/gitlabhq/terraform-provider-gitlab/pull/449/commits/dee4ce6f2a3425413b2242f3f494ab815604c7ef to include import_url on Project Updates but the build is still failing:

```
Error:     testing.go:684: Step 0 error: Check failed: Check 3/4 error: got mirror false; want true
Ref: https://github.com/gitlabhq/terraform-provider-gitlab/pull/449/checks?check_run_id=1622366823
```

So I assume I have to disable running mirroring in CE, but it appears that EE tests only run for PRs/branches on `gitlabhq/terraform-provider-gitlab`, not a forked repo since the EE token is stored in GH Secrets on the root repo.

I'd like to set `SkipFunc`, could you make a branch for me off of master, and I'll change this PR to merge into that branch instead, then we can merge into a branch owned by `gitlabhq` for EE tests to run? Or, alternatively/additionally can you grant me write access to that branch as well?

btyy77c: @cnuss I've setup a branch on `gitlabhq` to run EE tests: https://github.com/gitlabhq/terraform-provider-gitlab/pull/512. Thanks!

cnuss:
UPDATE: This PR has been superseded by #512 in order to have a branch on the project so EE tests can run/verify mirroring.

Issue #450, opened 2020-09-29
Enable custom UserAgent for metrics purposes


nicholasklick:
GitLab would like to enable a custom useragent for http client requests from the provider to GitLab instances.

Something along the lines of: `gitlab-terraform-provider`

roidelapluie:
Currently we use 2 user agents:

go-gitlab
Go-http-client/1.1

We could indeed have terraform-provider-gitlab/3.0.0

roidelapluie:
it would also be nice to have the version of terraform in the user agent.

nicholasklick:
@roidelapluie yes agreed that would make sense to include the version

PR #451, opened 2020-10-01
Enable custom UserAgent

bishtawi:
Resolves https://github.com/gitlabhq/terraform-provider-gitlab/issues/450

### DIFF
--- a/gitlab/provider.go
+++ b/gitlab/provider.go
@@ -129,13 +129,18 @@ func providerConfigure(p *schema.Provider, d *schema.ResourceData) (interface{},
 		Insecure:   d.Get("insecure").(bool),
 		ClientCert: d.Get("client_cert").(string),
 		ClientKey:  d.Get("client_key").(string),
-
-		// NOTE: httpclient.TerraformUserAgent is deprecated and removed in Terraform SDK v2
-		// After upgrading the SDK to v2 replace with p.UserAgent()
-		TerraformUserAgent: httpclient.TerraformUserAgent(p.TerraformVersion),
 	}
 
-	return config.Client()
+	client, err := config.Client()
+	if err != nil {
+		return nil, err
+	}
+
+	// NOTE: httpclient.TerraformUserAgent is deprecated and removed in Terraform SDK v2
+	// After upgrading the SDK to v2 replace with p.UserAgent("terraform-provider-gitlab")
+	client.UserAgent = httpclient.TerraformUserAgent(p.TerraformVersion) + " terraform-provider-gitlab"
+
+	return client, err
 }
 
 func validateApiURLVersion(value interface{}, key string) (ws []string, es []error) {
--- a/gitlab/provider.go
+++ b/gitlab/provider.go
@@ -118,12 +118,13 @@ func init() {
 
 func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	config := Config{
-		Token:      d.Get("token").(string),
-		BaseURL:    d.Get("base_url").(string),
-		CACertFile: d.Get("cacert_file").(string),
-		Insecure:   d.Get("insecure").(bool),
-		ClientCert: d.Get("client_cert").(string),
-		ClientKey:  d.Get("client_key").(string),
+		Token:            d.Get("token").(string),
+		BaseURL:          d.Get("base_url").(string),
+		CACertFile:       d.Get("cacert_file").(string),
+		Insecure:         d.Get("insecure").(bool),
+		ClientCert:       d.Get("client_cert").(string),
+		ClientKey:        d.Get("client_key").(string),
+		TerraformVersion: "TODO", // TODO: How to get Terraform version?
 	}
 
 	return config.Client()
@@ -132,7 +133,7 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 func validateApiURLVersion(value interface{}, key string) (ws []string, es []error) {
 	v := value.(string)
 	if strings.HasSuffix(v, "/api/v3") || strings.HasSuffix(v, "/api/v3/") {
-		es = append(es, fmt.Errorf("terraform-gitlab-provider does not support v3 api; please upgrade to /api/v4 in %s", v))
+		es = append(es, fmt.Errorf("terraform-provider-gitlab does not support v3 api; please upgrade to /api/v4 in %s", v))
 	}
 	return
 }
--- a/gitlab/provider.go
+++ b/gitlab/provider.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/httpclient"
 	"github.com/hashicorp/terraform-plugin-sdk/terraform"
 )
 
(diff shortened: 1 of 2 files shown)


gitlab/provider.go
```diff
@@ -132,7 +133,7 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 func validateApiURLVersion(value interface{}, key string) (ws []string, es []error) {
 	v := value.(string)
 	if strings.HasSuffix(v, "/api/v3") || strings.HasSuffix(v, "/api/v3/") {
-		es = append(es, fmt.Errorf("terraform-gitlab-provider does not support v3 api; please upgrade to /api/v4 in %s", v))
+		es = append(es, fmt.Errorf("terraform-provider-gitlab does not support v3 api; please upgrade to /api/v4 in %s", v))
```
bishtawi, line 136: I can revert this if you want, but I noticed that this line is the only reference of `terraform-gitlab-provider`. All other references use `terraform-provider-gitlab`.

roidelapluie:
For the user agent, can we use https://github.com/hashicorp/terraform-plugin-sdk/blob/9da7cbac0dfcd4c7096105eeda94ea69f53c1101/helper/schema/provider.go#L451 ?

Thanks!

gitlab/provider.go
```diff
@@ -118,12 +118,13 @@ func init() {
...
-		Insecure:   d.Get("insecure").(bool),
-		ClientCert: d.Get("client_cert").(string),
-		ClientKey:  d.Get("client_key").(string),
+		Token:            d.Get("token").(string),
+		BaseURL:          d.Get("base_url").(string),
+		CACertFile:       d.Get("cacert_file").(string),
+		Insecure:         d.Get("insecure").(bool),
+		ClientCert:       d.Get("client_cert").(string),
+		ClientKey:        d.Get("client_key").(string),
+		TerraformVersion: "TODO", // TODO: How to get Terraform version?
```
bishtawi, line 127: Any advice on how to get the Terraform version? 

I saw `var Version = "0.12.7"` from https://github.com/hashicorp/terraform-plugin-sdk/blob/v1.15.0/internal/version/version.go#L14 but I dont think that is what we want because its an internal module, doesnt exist in v2 of the SDK, and its the version of Terraform the provider is compiled with rather than the version of Terraform the consumer of the provider is running.

I see there is a `TerraformVersion string` attribute in the `schema.Provider struct` at https://github.com/hashicorp/terraform-plugin-sdk/blob/v1.15.0/helper/schema/provider.go#L72 which you instantiate here https://github.com/gitlabhq/terraform-provider-gitlab/blob/master/gitlab/provider.go#L15 but I cant figure out where that value is getting set.

The `State struct` from https://github.com/hashicorp/terraform-plugin-sdk/blob/v1.15.0/terraform/state.go#L87 has a `TFVersion string` attribute but that is just the version of Terraform that wrote the state file. Also, I couldnt figure out how to get the `state struct` instance from `d *schema.ResourceData`.

Is there a way of getting the Terraform version from `d *schema.ResourceData`?

bishtawi: > For the user agent, can we use https://github.com/hashicorp/terraform-plugin-sdk/blob/9da7cbac0dfcd4c7096105eeda94ea69f53c1101/helper/schema/provider.go#L451 ?
> 
> Thanks!

Just saw your comment @roidelapluie. I saw that `UserAgent`  function but that function is attached to the `Provider` struct. The functions `providerConfigure` and `config.Client()` dont have access to the `Provider` struct that gets instantiated at https://github.com/gitlabhq/terraform-provider-gitlab/blob/master/gitlab/provider.go#L15

I we have is `d *schema.ResourceData` which from what I can tell doesnt hold a reference to the `provider`.

bishtawi: OK, here is an idea of how to get the `provider` struct into the `providerConfigure` function: https://github.com/bishtawi/terraform-provider-gitlab/commit/9fcf005a0fe275e0f67765b8d71cc937017382b4

This idea would work provided that Terraform sets the `provider.TerraformVersion` variable before calling the `ConfigureFunc`. If it sets it after, then `provider.TerraformVersion` will be empty string.

bishtawi: Looked at the source of the SDK, looks like `provider.TerraformVersion` is set before `ConfigureFunc` is called.

roidelapluie:
How much work is it to go to the v2 of the sdk?

bishtawi:
Not sure but there is a lot of breaking changes between 1.x and 2.x: https://github.com/hashicorp/terraform-plugin-sdk/blob/master/CHANGELOG.md

roidelapluie:
In 1.0 we have httpclient which has the function we want

roidelapluie:
https://github.com/hashicorp/terraform-plugin-sdk/blob/v1-maint/httpclient/useragent.go#L18

bishtawi:
Nice good find!

nicholasklick:
@bishtawi thanks for this. So what is the strong generated by `c.TerraformUserAgent`? 
Does it include something unique to the provider? Like `gitlab-terraform-provider`.

Basically our goal here is to be able to identify API requests originating from the terraform gitlab provider.

gitlab/provider.go
```diff
@@ -116,14 +121,18 @@ func init() {
...
 		BaseURL:    d.Get("base_url").(string),
 		CACertFile: d.Get("cacert_file").(string),
 		Insecure:   d.Get("insecure").(bool),
 		ClientCert: d.Get("client_cert").(string),
 		ClientKey:  d.Get("client_key").(string),
+
+		// NOTE: httpclient.TerraformUserAgent is deprecated and removed in Terraform SDK v2
+		// After upgrading the SDK to v2 replace with p.UserAgent()
+		TerraformUserAgent: httpclient.TerraformUserAgent(p.TerraformVersion),
 	}
```
roidelapluie, line 137: client := provider.Client()
client.UserAgent = httpclient.TerraformUserAgent(p.TerraformVersion)
return client

gitlab/provider.go
```diff
@@ -93,9 +94,13 @@ func Provider() terraform.ResourceProvider {
 			"gitlab_project_mirror":             resourceGitlabProjectMirror(),
 			"gitlab_project_level_mr_approvals": resourceGitlabProjectLevelMRApprovals(),
 		},
+	}
 
-		ConfigureFunc: providerConfigure,
+	provider.ConfigureFunc = func(d *schema.ResourceData) (interface{}, error) {
+		return providerConfigure(provider, d)
 	}
```
roidelapluie, line 101: This is not needed

roidelapluie:
It looks like this pull request is way too complicated and it could all be done in gitlab/provider.go

bishtawi:
@roidelapluie Sure I can simplify it; I was just trying to keep with the existing code structure by having all the specific Gitlab Client customizations in the `Client()` function rather than having the provider also customize the client.

@nicholasklick No, the `httpclient.TerraformUserAgent()` function only includes Terraform specific info. We have to add Gitlab provider info ourselves. I'll add that in.

gitlab/provider.go
bishtawi: Why do you think this is not needed? The `providerConfigure` function needs the `provider` object to get the Terraform version number. Wrapping the `providerConfigure` function allows us to do this. Is there another way to accomplish it?

gitlab/provider.go
bishtawi: Done. Personally I am not a fan of the `providerConfigure` function customizing the `client` as that seems to be the responsibility of the `Client()` function but the code is indeed simpler.

bishtawi:
Useragent will look like `HashiCorp Terraform/0.13.4 (+https://www.terraform.io) Terraform Plugin SDK/1.15.0 terraform-provider-gitlab`.

nicholasklick:
@bishtawi that useragent string looks good thanks!

nicholasklick:
@roidelapluie @armsnyder if this looks good I would love to merge and ship a new release.

armsnyder (approved):
LGTM. A possible future improvement is to include the terraform-provider-gitlab provider version in the string somewhere. GoReleaser supports setting a version variable at release time, using ldflags, which could do this. https://goreleaser.com/customization/build/

PR #456, opened 2020-10-08
Remove old-style variable interpolation

svenstaro:
It's now considered bad form in Terraform to use interpolation with only variables and Terraform will raise a lint warning for it.


nicholasklick:
LGTM

PR #461, opened 2020-10-10
Move to provider v2 ; remove vendor

roidelapluie:


armsnyder:
@roidelapluie Do you have an update on this? I have another PR with a lot of vendor changes which I'd like to rebase onto this if we're not vendoring anymore.

nicholasklick:
@armsnyder is this PR now redundant?

armsnyder:
@nicholasklick Only partially. The other part of this PR was to update the SDK to v2. However, I wonder if we need to plan for this, since HashiCorp suggests that [updating to v2 should be a major provider release](https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html). (See #552 , [4.0 milestone](https://github.com/gitlabhq/terraform-provider-gitlab/milestone/2))

roidelapluie:
Yes we want that

timofurrer:
Already done with #725 .

PR #474, opened 2020-10-30
Make the runners_token on the project secret

valdisrigdon:
This is a secret value and should be a `Sensitive` value.


armsnyder (approved):
Thanks!

PR #482, opened 2020-11-15
Documentation for `expires_at`

towo:
I added a short line for the configuration of `expires_at` as per the code.


armsnyder (approved):
Thanks!

Issue #483, opened 2020-11-17
Fix deprecated set-env and add-path references in workflows


nicholasklick:
Our builds are [currently failing](https://github.com/gitlabhq/terraform-provider-gitlab/pull/250/checks) due to [deprecated commands](https://github.blog/changelog/2020-11-09-github-actions-removing-set-env-and-add-path-commands-on-november-16/). 

We need to migrate to the [new syntax](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable).

## Definition of done

- [ ] Migrate to new syntax
- [ ] Builds passing

nicholasklick:
Fixed via https://github.com/gitlabhq/terraform-provider-gitlab/pull/484

PR #484, opened 2020-11-18
Update set-env github action command. 

MoonSupport:
https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

set-env is deprecating right now. So, we should update! :)


nicholasklick:
Fixes https://github.com/gitlabhq/terraform-provider-gitlab/issues/483

PR #490, opened 2020-11-23
Fix nil pointer dereference importing gitlab_user

### LINKED ISSUE #469: segmentation violation while running import for GitLab user (SIGSEGV)
### Terraform (and provider) Version

```
$ terraform --version
Terraform v0.13.4
+ provider registry.terraform.io/gitlabhq/gitlab v3.1.0
```

```terraform
# provider.tf
terraform {
  required_providers {
    gitlab = {
      source = "gitlabhq/gitlab"
      version = "3.1.0"
    }
    secrethub = {
      source = "secrethub/secrethub"
      version = ">= 1.2.0"
    }
  }
}
provider "gitlab" {
  token = "xxxxxxxxxxxxxxxxxxxx"
}
provider "secrethub" {
  credential = file("/path/to/.secrethub/credential")
}
# users.tf
resource "gitlab_user" "foo-bar" {
  name     = "Foo Bar"
  username = "foo-bar"
  email    = "foo+bar@example.com"
  password         = "asdfg"
  is_admin         = false
  projects_limit   = 0
  can_create_group = false
  is_external      = true
  reset_password   = false
}
```

### Debug Output
(crash)

### Crash Output
[`crash.log`](https://gist.github.com/franz-josef-kaiser/eb1f9b557011b8cbc7a756fd9c787060) as GitHub Gist.

### Expected Behavior
Import user

### Actual Behavior
Crashed; segmentation violation

The crash can be repeated in the current setup multiple times in a row. 

```
gitlab_user.foo-bar: Importing from ID "12345678"...

Error: rpc error: code = Unavailable desc = transport is closing


panic: runtime error: invalid memory address or nil pointer dereference
2020-10-22T10:57:37.686+0200 [DEBUG] plugin.terraform-provider-gitlab_v3.1.0: [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1a36b15]
```

### Steps to Reproduce
```
terraform import -input=true gitlab_user.foo-bar 12345678
```
### Attempts to work around
* Removing `-input=true` from the import statement (crashes).
* Renaming `foo-bar` to `foo_bar` and `foobar`
* Removing all parameters and values from `resource "gitlab_user"` (crashes).
    ```
    resource "gitlab_user" "foo-bar" {
      # (resource arguments)
    }
    ```

### Additional Context
You need to call the GitLab API (v4) to get some users ID. As it's a simply `GET` request, use your browser:
```
https://gitlab.com/api/v4/users?username=foo-bar
```

### Important Factoids
Local only. Ran with remote backend `http`, provided via GitLab, as well.

### References
Originally opened in hashicorp/terraform#26669

armsnyder:
Fixes #469

The issue was a missing `nil` check.

An importer test was ignoring the `skip_confirmation` field, probably masking the bug.

Note I replaced the importer with `schema.ImportStatePassthrough` because the importer was duplicated code from the read function, and the read function only uses the `id` attribute to read. The three extra attribute-sets (lines 31-33) are covered already by `resourceGitlabUserSetToState`.

### DIFF
--- a/gitlab/resource_gitlab_user.go
+++ b/gitlab/resource_gitlab_user.go
@@ -16,23 +16,7 @@ func resourceGitlabUser() *schema.Resource {
 		Update: resourceGitlabUserUpdate,
 		Delete: resourceGitlabUserDelete,
 		Importer: &schema.ResourceImporter{
-			State: func(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
-				client := meta.(*gitlab.Client)
-				log.Printf("[DEBUG] read gitlab user %s", d.Id())
-
-				id, _ := strconv.Atoi(d.Id())
-
-				user, _, err := client.Users.GetUser(id)
-				if err != nil {
-					return nil, err
-				}
-
-				resourceGitlabUserSetToState(d, user)
-				d.Set("email", user.Email)
-				d.Set("is_admin", user.IsAdmin)
-				d.Set("is_external", user.External)
-				return []*schema.ResourceData{d}, nil
-			},
+			State: schema.ImportStatePassthrough,
 		},
 
 		Schema: map[string]*schema.Schema{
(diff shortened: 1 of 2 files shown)


roidelapluie:
Thanks!

PR #491, opened 2020-11-23
gitlab_user: Do not set skip_confirmation on read

armsnyder:
Follow-up on #490 @roidelapluie 

I looked at it again, and I don't think `skip_confirmation` should ever be set during reads or imports, since it is an attribute that only has effect during initial creation. ([API ref](https://docs.gitlab.com/ee/api/users.html))

This change should fix an issue where, if the resource is created with `skip_confirmation=false`, then when the user confirms their email, it would result in an unnecessary resource diff of `skip_confirmation=true`.

### DIFF
--- a/gitlab/resource_gitlab_user.go
+++ b/gitlab/resource_gitlab_user.go
@@ -55,6 +55,7 @@ func resourceGitlabUser() *schema.Resource {
 				Type:     schema.TypeBool,
 				Optional: true,
 				Default:  true,
+				ForceNew: true,
 			},
 			"projects_limit": {
 				Type:     schema.TypeInt,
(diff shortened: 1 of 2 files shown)


roidelapluie:
If we do not set it on read, it means that it will always be flapping if a user sets it to true.

We might need a suppressdiff function?

armsnyder:
Hmm, I thought if we don't `d.Set()` it, then it will stay whatever it was originally set to.

roidelapluie:
> Hmm, I thought if we don't `d.Set()` it, then it will stay whatever it was originally set to.

It will not change the user input on import, would it?

roidelapluie:
The correct behaviour would be:

1. set it to computed + force_new

armsnyder:
Updated. That seemed to work.

armsnyder:
Let's not merge this until #493 is resolved, since I did not run these tests myself locally.

timofurrer (changes requested):
Needs a rebase, but I think otherwise LGTM. The referenced blocker has also been resolved some time ago.
What do you think @armsnyder ?

timofurrer (approved):

Issue #494, opened 2020-11-24
TestAccGitLabProjectApprovalRule_basic test failure


armsnyder:
The `TestAccGitLabProjectApprovalRule_basic` acceptance test is failing. It was introduced in #250 but we didn't catch it due to #493.

Partial log:

```
2020-11-24T09:43:11.9256854Z ##[error]    testing.go:745: Error destroying resource! WARNING: Dangling resources
2020-11-24T09:43:11.9268233Z         may exist. The full state and error is shown below.
2020-11-24T09:43:11.9268798Z         
2020-11-24T09:43:11.9269286Z         Error: Check failed: User still exists
2020-11-24T09:43:11.9269753Z         
2020-11-24T09:43:11.9270368Z         State: <no state>
2020-11-24T09:43:11.9271577Z --- FAIL: TestAccGitLabProjectApprovalRule_basic (11.62s)
```

Full log: https://github.com/gitlabhq/terraform-provider-gitlab/runs/1446853398

nicholasklick:
@btyy77c mind taking a look at this?

nicholasklick:
Actually @armsnyder this appears resolved in the latest run?

```
2020-11-30T05:31:05.5426370Z === RUN   TestAccGitLabProjectApprovalRule_basic
2020-11-30T05:31:17.1621315Z --- PASS: TestAccGitLabProjectApprovalRule_basic (11.62s)
```

nicholasklick:
Also looks good here https://github.com/gitlabhq/terraform-provider-gitlab/actions/runs/390938808

I say we close this?

armsnyder:
All good then!

PR #495, opened 2020-11-24
Fix unit and acceptance tests not running

### LINKED ISSUE #493: Acceptance tests not working
I've noticed in two active PRs that there are written tests which should be failing but are not failing.

Example job: https://github.com/gitlabhq/terraform-provider-gitlab/pull/449/checks?check_run_id=1443975846

The acceptance tests in PRs are succeeding after just 43 seconds, so I don't think they're actually running.

@nicholasklick we should probably not release again until we can sort this out, since there would have been PRs merged into master while this problem existed.

armsnyder:
Fixes #493 

Unit and acceptance tests haven't been running since #484. Some misplaced quotation marks caused the variables to not be set, so the default Make target was being run in these workflows instead of the tests.

[Example](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#example-5)

Let's check that PRs created or rebased in the last week get this fix before they merge.

I opened an issue to look into the one acceptance test that is failing: #494

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -18,11 +18,11 @@ jobs:
     # https://help.github.com/en/actions/reference/workflow-commands-for-github-actions
     - name: Set build variables
       run: |
-        echo "MAKE_TARGET=testacc >> $GITHUB_ENV"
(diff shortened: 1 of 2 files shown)


nicholasklick:
This looks correct to me.

PR #505, opened 2020-12-14
Allow to configure both oauth and private token in the provider …

christophedcpm:
when both token are provided, the gitlab client instanciated will use
both authentication method to talk to the API


christophedcpm:
sorry wrong base repo for the PR.

PR #510, opened 2020-12-23
fix deploy_token expiration

Bobonium:
These changes will fix the expiry behavior of deploy tokens.

Current state:
- Leaving `expires_at` at null in terraform results in the created token automatically being expired.
- Using an unrealistically high date as a workaround results in terraform segfaults, because gitlab internally sets them to never expire and the API will not return a valid date afterwards

Changes:
- Tokens will not expire per default, which makes more sense than creating an already expired Token
- Changes to the read behavior to check for `nil` before converting the date, as to not run into the terraform segfaults


armsnyder:
Fixes #513

armsnyder (approved):
Thanks for the fix!

PR #527, opened 2021-01-25
Link badges in README to proper workflows

svenstaro:
It's nice being able to click on those and getting redirected to the corresponding workflow.


nicholasklick:
Great thanks!

PR #532, opened 2021-01-27
Allow `full_path` in addition to `id` in gitlab_project data source

### LINKED ISSUE #140: [Feature Request] use full_path to define gitlab_project data source
Hi!

While refactoring my terraform repository I have found out that it would be awesome to provide same way to create gitlab_project data source as for gitlab_group by using full_path instead of id, this woulde make a code more human readable :).

sirlatrom:
Fixes #140.

### DIFF
--- a/docs/data-sources/project.md
+++ b/docs/data-sources/project.md
@@ -10,11 +10,17 @@ data "gitlab_project" "example" {
 }
 ```
 
+```hcl
+data "gitlab_project" "example" {
+	id = "foo/bar/baz"
+}
+```
+
 ## Argument Reference
 
 The following arguments are supported:
 
-* `id` - (Required) The integer that uniquely identifies the project within the gitlab install.
+* `id` - (Required) The integer or path with namespace that uniquely identifies the project within the gitlab install.
 
 ## Attributes Reference
 
--- a/gitlab/data_source_gitlab_project.go
+++ b/gitlab/data_source_gitlab_project.go
@@ -16,7 +16,7 @@ func dataSourceGitlabProject() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"id": {
-				Type:     schema.TypeInt,
+				Type:     schema.TypeString,
 				Required: true,
 			},
 			"name": {
--- a/gitlab/data_source_gitlab_project_test.go
+++ b/gitlab/data_source_gitlab_project_test.go
@@ -17,6 +17,11 @@ func TestAccDataGitlabProject_basic(t *testing.T) {
 		Providers: testAccProviders,
 		Steps: []resource.TestStep{
 			{
+				Config: testAccDataGitlabProjectConfigByPathWithNamespace(projectname),
+				Check: testAccDataSourceGitlabProject("gitlab_project.test", "data.gitlab_project.foo",
+					[]string{"id", "name", "path", "visibility", "description"}),
+			},
+			{
 				Config: testAccDataGitlabProjectConfig(projectname),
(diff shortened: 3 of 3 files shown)


armsnyder (changes requested):
Thanks for the contribution. A better fix would be to change the type for `id` into `string` in the resource schema (this is a backward compatible change) and update the documentation to mention that id can be an int or string. I don't think `full_path` is the standard name for this field in the GitLab API, nor is it used anywhere else in this plugin.

sirlatrom:
> 
> 
> Thanks for the contribution. A better fix would be to change the type for `id` into `string` in the resource schema (this is a backward compatible change) and update the documentation to mention that id can be an int or string. I don't think `full_path` is the standard name for this field in the GitLab API, nor is it used anywhere else in this plugin.

It's used here: https://github.com/gitlabhq/terraform-provider-gitlab/blob/f821d591b6fdb7e388c3eed2b637e758900fb553/gitlab/data_source_gitlab_group.go#L23-L30 which is what I took as inspiration.

armsnyder:
> It's used here:
> 
> https://github.com/gitlabhq/terraform-provider-gitlab/blob/f821d591b6fdb7e388c3eed2b637e758900fb553/gitlab/data_source_gitlab_group.go#L23-L30
> 
> which is what I took as inspiration.

Ahh, good point. However, I would still rather stick to the official GitLab API field names, as I see this as API debt.

See the API fields names for GET Project: https://docs.gitlab.com/ee/api/projects.html#get-single-project

See the HashiCorp guidance on this subject: https://www.terraform.io/docs/extend/hashicorp-provider-design-principles.html#resource-and-attribute-schema-should-closely-match-the-underlying-api

sirlatrom:
> Ahh, good point. However, I would still rather stick to the official GitLab API field names, as I see this as API debt.
> 
> See the API fields names for GET Project: https://docs.gitlab.com/ee/api/projects.html#get-single-project
> 
> See the HashiCorp guidance on this subject: https://www.terraform.io/docs/extend/hashicorp-provider-design-principles.html#resource-and-attribute-schema-should-closely-match-the-underlying-api

OK, I'll change the type of `id`, but are you quite sure you can go from a `TypeInt` to `TypeString` without state parsing issues? I'll give it a try.

armsnyder:
> OK, I'll change the type of `id`, but are you quite sure you can go from a `TypeInt` to `TypeString` without state parsing issues? I'll give it a try.

Yep, going from int -> string is fine. Terraform will update the type in the state on the next apply, and users can even continue to set this field to a number in their config.

sirlatrom:
> Yep, going from int -> string is fine. Terraform will update the type in the state on the next apply, and users can even continue to set this field to a number in their config.

@armsnyder Done in 55a5f54.

docs/data-sources/project.md
```diff
@@ -10,11 +10,19 @@ data "gitlab_project" "example" {
...
+```
+
 ## Argument Reference
 
 The following arguments are supported:
 
-* `id` - (Required) The integer that uniquely identifies the project within the gitlab install.
+* `id` - (Required) The integer or path with namespace that uniquely identifies the project within the gitlab install.
+
+**Note**: Exactly one of `id` or `full_path` must be provided.
```
armsnyder, line 25: Can you revert this note please

armsnyder (changes requested):
Great! One final note.

docs/data-sources/project.md
sirlatrom: Yup, sorry I missed that. 091c23f should do it.

armsnyder (approved):
Thanks!

PR #538, opened 2021-01-29
gitlab_project_mirror: Fix url attribute drift when using basicauth

### LINKED ISSUE #504: gitlab_project_mirror replaces existing mirror with each apply
### Terraform Version
Terraform v0.13.5
+ provider registry.terraform.io/gitlabhq/gitlab v3.2.0

### Affected Resource(s)
- gitlab_project_mirror

```hcl
resource "gitlab_project_mirror" "test_moving_mirror" {
  project                 = module.test_moving_old.project_id
  enabled                 = true
  url                     = "https://${var.mirror_user}:${var.mirror_token}@gitlab.com/######/###/poc/test-group/test-moving.git"
  only_protected_branches = false
  keep_divergent_refs     = false
}
```

### Debug Output
https://gist.github.com/tredfield/034de1667f23266e6fa563b211f7f306

### Panic Output
NA

### Expected Behavior
Once a gitlab_project_mirror is applied performing a plan or apply it should not change

### Actual Behavior
After performing an apply, each plan or apply after states the mirror must be replaced with statement "gitlab_project_mirror.test_moving_mirror must be replaced". See example

```
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:

  # gitlab_project_mirror.test_moving_mirror must be replaced
-/+ resource "gitlab_project_mirror" "test_moving_mirror" {
        enabled                 = true
      ~ id                      = "22846482:259337" -> (known after apply)
        keep_divergent_refs     = false
      ~ mirror_id               = 259337 -> (known after apply)
        only_protected_branches = false
        project                 = "22846482"
      ~ url                     = (sensitive value)
    }

```

The result is a new mirror is created in gitlab with each apply. The old mirror is disabled, creating a list of disabled mirrors.

### Steps to Reproduce
# initial add of mirror
1. `terraform apply`
2. `terraform apply`

# with no changes made to TF
3. `terraform plan`
4. `terraform apply`

### Important Factoids
Are there anything atypical about your accounts that we should know? For example: Running in EC2 Classic? Custom version of OpenStack? Tight ACLs?
No

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
No

armsnyder:
Fixes #504 

**Note** the `testAccGitlabProjectContext` code is not new; it is moved from the `gitlab_project_variable` test into our test utils.

The fix is to conditionally set the "url" attribute only if it does not contain redacted text.

The caveat is that the "url" attribute cannot be imported if it has redacted text. I don't expect this to be a problem. The attribute is `ForceNew: true` so any upstream change to the value is not possible.

As a bonus I fixed the destroy test to actually test the delete function, by creating the Gitlab project outside of the test state. I moved `testAccGitlabProjectContext` into `helper_test.go` now that it is used more than once, but didn't modify it.

### DIFF
--- a/gitlab/resource_gitlab_project_mirror.go
+++ b/gitlab/resource_gitlab_project_mirror.go
@@ -173,5 +173,10 @@ func resourceGitlabProjectMirrorSetToState(d *schema.ResourceData, projectMirror
 	d.Set("keep_divergent_refs", projectMirror.KeepDivergentRefs)
 	d.Set("only_protected_branches", projectMirror.OnlyProtectedBranches)
 	d.Set("project", projectID)
-	d.Set("url", projectMirror.URL)
+
+	if strings.Contains(projectMirror.URL, "*****") {
+		log.Printf("[DEBUG] project mirror %q will not detect any possible change to the url because it is redacted by Gitlab", d.Id())
+	} else {
+		d.Set("url", projectMirror.URL)
+	}
 }
--- a/gitlab/helper_test.go
+++ b/gitlab/helper_test.go
@@ -3,7 +3,11 @@ package gitlab
 import (
 	"errors"
 	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"os"
 	"strings"
+	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/xanzy/go-gitlab"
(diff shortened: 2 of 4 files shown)


gitlab/resource_gitlab_project_mirror.go
```diff
@@ -173,5 +173,10 @@ func resourceGitlabProjectMirrorSetToState(d *schema.ResourceData, projectMirror
 	d.Set("keep_divergent_refs", projectMirror.KeepDivergentRefs)
 	d.Set("only_protected_branches", projectMirror.OnlyProtectedBranches)
 	d.Set("project", projectID)
-	d.Set("url", projectMirror.URL)
+
+	if strings.Contains(projectMirror.URL, "*****") {
```
roidelapluie, line 177: that is not the best approach, you should parse the url and remove password in diff suppress func

roidelapluie (changes requested):

gitlab/resource_gitlab_project_mirror.go
armsnyder: That sounds cleaner, but do you mean that we should save the redacted value in state? Wouldn't that make it impossible for the user to update the user/pass for their push mirror, since the diff would not be recognized?

timofurrer:
@armsnyder needs some rebase and conflict solving, and is this still a thing? :)

armsnyder:
Not needed since the original issue #504 was closed. I will re-open this in a new PR with the test refactor.

PR #540, opened 2021-01-29
Add protected_branch(es) to gitlab_project

### LINKED ISSUE #492: Add optional protected_branch_ids to the project_level_mr_approvals resource
### Enhancement proposal
Add  [protected_branch_ids](https://docs.gitlab.com/ee/api/merge_request_approvals.html#create-project-level-rule) to the [project_level_mr_approvals](https://registry.terraform.io/providers/gitlabhq/gitlab/latest/docs/resources/project_level_mr_approvals) resource.

As it currenty stands my organization has different approval rules for different branches in place.  I would love to be able to scope the current resource to a branch instead of only being able to specify general rules for every branch on the project

sirlatrom:
Fixes #492.

---

Still missing:

  - [ ] Docs
  - [ ] Acceptance tests

### DIFF
--- a/gitlab/data_source_gitlab_project.go
+++ b/gitlab/data_source_gitlab_project.go
@@ -152,6 +152,32 @@ func dataSourceGitlabProject() *schema.Resource {
 					},
 				},
 			},
+			"protected_branches": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:     schema.TypeInt,
+							Computed: true,
+						},
+						"name": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"push_access_levels":      schemaDataAllowedAccessLevels(),
+						"merge_access_levels":     schemaDataAllowedAccessLevels(),
+						"unprotect_access_levels": schemaDataAllowedAccessLevels(),
+						"allowed_to_push":         schemaDataAllowedTo(),
+						"allowed_to_merge":        schemaDataAllowedTo(),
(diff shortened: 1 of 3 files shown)


sirlatrom:
@armsnyder This PR is a draft because I haven't found out how to properly trigger dependent resources to update the numeric ID of the protected branches after they're re-created. A possible solution may be to split it out into a separate resource, e.g. `gitlab_project_protected_branch`, as that could be configured with `ForceNew: true` when fields like `user_id` and `group_id` are changed.

Note: A protected branch has its own numeric ID in addition to the name of the branch, and it is updated whenever the access level, user or group is changed.

armsnyder:
@sirlatrom We should have a `gitlab_project_protected_branch` resource, which is also how the GitLab API handles it.

The resource ID should be `<project-id>:<branch-name-or-wildcard>` because those are the two path variables in the GitLab API. (`GET /projects/:id/protected_branches/:name`)

sirlatrom:
> 
> 
> @sirlatrom We should have a `gitlab_project_protected_branch` resource, which is also how the GitLab API handles it.
> 
> The resource ID should be `<project-id>:<branch-name-or-wildcard>` because those are the two path variables in the GitLab API. (`GET /projects/:id/protected_branches/:name`)

I see now there is a [gitlab_branch_protection](https://registry.terraform.io/providers/gitlabhq/gitlab/latest/docs/resources/branch_protection) resource, which I suppose could have been more aptly named.

Anyway, I can move my processing of the additional supported fields to that resource. How about the data source modifications? Should they stay on the `gitlab_project` data source, or do you envision a `gitlab_project_protected_branches`/`gitlab_branch_protections` data source to enumerate them for a given project?

sirlatrom:
Closing in favor of #542.

Issue #524, opened 2021-01-25
gitlab_group_share_group without expiration date gives nil pointer reference


valvin1:
### Terraform Version
Terraform v0.14.5
### Affected Resource(s)
- gitlab_group_share_group

### Terraform Configuration Files
```hcl
resource "gitlab_group_share_group" "my_group_share_group" {
  group_id       = data.gitlab_group.my_group.id
  share_group_id = gitlab_group.my_share_group.id
  group_access      = "maintainer"
}
```

### Debug Output
```
rpc error: code = Unavailable desc = transport is closing
```

### Panic Output

```
!!!!!!!!!!!!!!!!!!!!!!!!!!! TERRAFORM CRASH !!!!!!!!!!!!!!!!!!!!!!!!!!!!
panic: runtime error: invalid memory address or nil pointer dereference
2021-01-25T08:12:20.034Z [DEBUG] plugin.terraform-provider-gitlab_v3.4.0: [signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xe10f02]
2021-01-25T08:12:20.034Z [DEBUG] plugin.terraform-provider-gitlab_v3.4.0: 
2021-01-25T08:12:20.034Z [DEBUG] plugin.terraform-provider-gitlab_v3.4.0: goroutine 45 [running]:
2021-01-25T08:12:20.034Z [DEBUG] plugin.terraform-provider-gitlab_v3.4.0: github.com/gitlabhq/terraform-provider-gitlab/gitlab.resourceGitlabGroupShareGroupRead(0xc0001aa770, 0x104d4e0, 0xc000634000, 0x2, 0x2)
2021-01-25T08:12:20.034Z [DEBUG] plugin.terraform-provider-gitlab_v3.4.0: 	github.com/gitlabhq/terraform-provider-gitlab/gitlab/resource_gitlab_group_share_group.go:108 +0x432
2021-01-25T08:12:20.034Z [DEBUG] plugin.terraform-provider-gitlab_v3.4.0: github.com/gitlabhq/terraform-provider-gitlab/gitlab.resourceGitlabGroupShareGroupCreate(0xc0001aa770, 0x104d4e0, 0xc000634000, 0x2, 0x1aebd00)
[...]

PR #555, opened 2021-02-09
Fix group_share_group nil pointer reference

btyy77c:
Related issue: https://github.com/gitlabhq/terraform-provider-gitlab/issues/524

Updates resource `gitlab_group_share_group`.  It only sets expires_at field when time is not nil.

### DIFF
--- a/gitlab/resource_gitlab_group_share_group.go
+++ b/gitlab/resource_gitlab_group_share_group.go
@@ -105,7 +105,10 @@ func resourceGitlabGroupShareGroupRead(d *schema.ResourceData, meta interface{})
 			d.Set("group_id", groupId)
 			d.Set("share_group_id", sharedGroup.GroupID)
 			d.Set("group_access", accessLevel[convertedAccessLevel])
-			d.Set("expires_at", sharedGroup.ExpiresAt.String())
+
+			if sharedGroup.ExpiresAt != nil {
+				d.Set("expires_at", sharedGroup.ExpiresAt.String())
+			}
 
 			return nil
 		}
--- a/gitlab/resource_gitlab_group_share_group.go
+++ b/gitlab/resource_gitlab_group_share_group.go
@@ -106,7 +106,9 @@ func resourceGitlabGroupShareGroupRead(d *schema.ResourceData, meta interface{})
 			d.Set("share_group_id", sharedGroup.GroupID)
 			d.Set("group_access", accessLevel[convertedAccessLevel])
 
-			if sharedGroup.ExpiresAt != nil {
+			if sharedGroup.ExpiresAt == nil {
+				d.Set("expires_at", "")
+			} else {
 				d.Set("expires_at", sharedGroup.ExpiresAt.String())
 			}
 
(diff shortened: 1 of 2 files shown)


btyy77c:
@mattkasa @armsnyder 

Could you review this PR? 

Thanks!

gitlab/resource_gitlab_group_share_group.go
```diff
@@ -105,7 +105,10 @@ func resourceGitlabGroupShareGroupRead(d *schema.ResourceData, meta interface{})
 			d.Set("group_id", groupId)
 			d.Set("share_group_id", sharedGroup.GroupID)
 			d.Set("group_access", accessLevel[convertedAccessLevel])
-			d.Set("expires_at", sharedGroup.ExpiresAt.String())
+
+			if sharedGroup.ExpiresAt != nil {
+				d.Set("expires_at", sharedGroup.ExpiresAt.String())
+			}
```
mattkasa, line 111: So the behavior will be to not pass `expires_at` to the API if it is `nil`. Makes sense, the only concern would be if `nil` had a special meaning, like removing an expiration, but since we don't have an update endpoint, and `ForceNew` is set on all attributes, I think this is fine the way you have it. Great work 👍

mattkasa (approved):

gitlab/resource_gitlab_group_share_group_test.go
```diff
@@ -46,8 +60,13 @@ func TestAccGitlabGroupShareGroup_import(t *testing.T) {
 		Steps: []resource.TestStep{
 			{
 				// create shared groups
-				Config: testAccGitlabGroupShareGroupConfig(randName, "guest", "2099-03-03"),
-				Check:  testAccCheckGitlabGroupSharedWithGroup(randName, "2099-03-03", gitlab.GuestPermissions),
```
mattkasa, line 50: Looks like this `Check` is just removed, nothing added back, was that intentional or just needs to be added back?

mattkasa (changes requested):
Just a question about the test.

gitlab/resource_gitlab_group_share_group.go
armsnyder: Hehe I have the same nitpick, but it's more of a semantic thing. Since the `expires_at` field cannot be modified upstream, so it doesn't make a difference in this case. Though, in a typical read function, you'd want to set each attribute which is returned from the GitLab API, i.e. set "expires_at" to "" in the case where ExpiresAt is nil.

armsnyder (approved):
looks great, if you can address @mattkasa 's comment on the test. approving since I have nothing else to add. 😄

gitlab/resource_gitlab_group_share_group.go
btyy77c: I've updated to 

```
if sharedGroup.ExpiresAt == nil {
				d.Set("expires_at", "")
} else {
				d.Set("expires_at", sharedGroup.ExpiresAt.String())
}
```

I thought the empty string would fail the `validateDateFunc` but it seems to be ok now 😅

gitlab/resource_gitlab_group_share_group_test.go
btyy77c: I noticed other `import` tests skipped this `check` so I wasn't sure it was needed.  I went ahead and added it back. 

Thanks!

armsnyder:
lgtm after the last couple commits 👍

gitlab/resource_gitlab_group_share_group.go
mattkasa: Looks great!

mattkasa (approved):

PR #556, opened 2021-02-10
Add most Premium features for gitlab_branch_protection

### LINKED ISSUE #165: [Feature Request] Add users and groups to gitlab_branch_protection
It looks like `gitlab_branch_protection` resource currently only supports `access_level` (referred to as Roles in Gitlab UI). [The API](https://docs.gitlab.com/ee/api/protected_branches.html#example-with-user--group-level-access-starter) also supports `user_id` and `group_id`. Would be fantastic to have those available as well.

sirlatrom:
Fixes #165.

---

~TO DO:~
  - [x] Working EE acceptance tests
  - [x] Updated docs


sirlatrom:
@armsnyder This is probably not yet fully working, but I hit a snag on the acceptance tests in gitlab/resource_gitlab_branch_protection_test.go where e.g. `TestAccGitlabBranchProtection_createWithCodeOwnerApproval` gives me a 403 when creating the `gitlab_branch_protection`. When trying to create a protected branch through the UI, it doesn't work until the project has files in the repository, which is currently not possible to do with the provider except if specifying a template project.

sirlatrom:
> `TestAccGitlabBranchProtection_createWithCodeOwnerApproval` gives me a 403 when creating the `gitlab_branch_protection`

This turned out to be that I'm not allowed to set the `unprotect_access_levels` or `allowed_to_unprotect` fields through the API. I will omit those fields in the resource for now, then.

sirlatrom:
@armsnyder I've resolved all acceptance test issues now.

Please note that while the response structure contains lists of access levels, allowed users and groups in one long list per action (push, merge), the create options are not made this way, and one has to filter through the lists to find what one is looking for. Hence the proposed argument structure, including the use of `TypeSet` instead of `TypeList`, as the input order is not guaranteed to be reflected in the response when reading.

gabrielrinaldi:
Any plans on merging this? This would be great to allow a bot to push code while restricting everyone else.

mattkasa:
@sirlatrom this looks great! thank you for implementing this!

Apologies for it taking so long to get reviewed :bow:

Your code and tests look great to me, and I agree with you about using `TypeSet` for `allowed_to_push` and `allowed_to_merge`. Especially since that leaves us room to add `access_level` inside those in future as well to match the API 👍 

I've been able to build and run your branch locally, and test it against a Premium licensed GitLab instance, but I am seeing this when running acceptance tests against that Premium instance, so I was wondering if you had any thoughts related to the `code_owners_approval_required` licensed feature check? I did verify it's enabled and available on my test instance and wondering if we might also see this failure if we merge and it runs the EE acceptance tests.

```
=== RUN   TestAccGitlabBranchProtection_basic
    testing.go:677: Step 4, expected error:
        
        Check failed: Check 3/3 error: got code_owner_approval_required true; want false
        
        To match:
        
        feature unavailable: code owner approvals
        
        
--- FAIL: TestAccGitlabBranchProtection_basic (45.56s)
=== RUN   TestAccGitlabBranchProtection_createWithCodeOwnerApproval
    testing.go:665: Step 2, no error received, but expected a match to:
        
        feature unavailable: code owner approvals
        
        
--- FAIL: TestAccGitlabBranchProtection_createWithCodeOwnerApproval (32.51s)
=== RUN   TestAccGitlabBranchProtection_createWithMultipleAccessLevels
--- PASS: TestAccGitlabBranchProtection_createWithMultipleAccessLevels (1.53s)
FAIL
FAIL	github.com/gitlabhq/terraform-provider-gitlab/gitlab	80.738s
FAIL
make: *** [testacc] Error 1
```

mattkasa:
Ah, looks like I'm getting that on `master` currently as well, so it's nothing to do with your changes, in that case I'll go ahead and merge this and we can move forward from there 👍

mattkasa (approved):

PR #565, opened 2021-02-11
Fix "Run failed" on forks

armsnyder:
The acceptance-ee job is failing on the forked repo anytime someone makes a PR, which results in annoying notifications to the contributor. The fix is to only run the job if the repo is in the gitlabhq org.

I've verified that, with this fix, the job is enabled on pushes to this repo, but is disabled on forks.


armsnyder:
Opened an issue for the flaky test #566

PR #567, opened 2021-02-11
Support Project (Protected) Environments (if license allows it)

### LINKED ISSUE #444: Environments support
From what I gathered currently there is no way for provider to create Project Environments nor there is a way to manage Protected Environments. Seems that at least normal Environments are supported by github.com/xanzy/go-gitlab

Any plans to add a resource for it?

sirlatrom:
WIP on Protected Environments.

Adds resource `gitlab_project_protected_environment`.

I've added `gitlab_project_environment` as well, with a rather minimal set of attributes.

Fixes #444.


sirlatrom:
Ping @armsnyder for opinions

sirlatrom:
Oops, forgot to wait for https://github.com/xanzy/go-gitlab/pull/1066 before marking as ready 😅

gitlab/resource_gitlab_project_environment.go
```diff
@@ -0,0 +1,163 @@
...
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringIsNotEmpty,
+			},
+			"external_url": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.StringIsNotEmpty,
+			},
+			"state": {
```
armsnyder, line 41: Can you add this attribute to the doc page

gitlab/resource_gitlab_project_environment.go
```diff
@@ -0,0 +1,163 @@
...
+
+	project := d.Get("project").(string)
+
+	log.Printf("[DEBUG] Project %s create gitlab environment %q", project, *options.Name)
+
+	client := meta.(*gitlab.Client)
+
+	environment, resp, err := client.Environments.CreateEnvironment(project, &options)
+	if err != nil {
+		if resp.StatusCode == http.StatusNotFound {
```
armsnyder, line 66: resp always needs a nil check, since the err could be anything.

gitlab/resource_gitlab_project_environment.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return fmt.Errorf("error converting environment ID to int: %v", err)
+	}
+
+	name := d.Get("name").(string)
+	options := gitlab.EditEnvironmentOptions{
+		Name: &name,
+	}
+	if externalURL, ok := d.GetOk("external_url"); ok {
+		options.ExternalURL = gitlab.String(externalURL.(string))
```
armsnyder, line 125: Usually use HasChange in the update function

```suggestion
	if d.HasChange("external_url") {
		options.ExternalURL = gitlab.String(d.Get("external_url").(string))
```

gitlab/resource_gitlab_project_environment.go
```diff
@@ -0,0 +1,163 @@
...
+	client := meta.(*gitlab.Client)
+
+	_, err = client.Environments.StopEnvironment(project, environmentIDInt)
+	if err != nil {
+		return fmt.Errorf("error stopping gitlab project %q environment %q: %v", project, environmentID, err)
+	}
+	_, err = client.Environments.DeleteEnvironment(project, environmentIDInt)
+	if err != nil {
+		return fmt.Errorf("error deleting gitlab project %q environment %q: %v", project, environmentID, err)
+	}
```
armsnyder, line 160: Is the StopEnvironment API call idempotent? For example, if the DeleteEnvironment call fails and a user retries the delete, will the StopEnvironment API call fail on the second try? (It shouldn't.)

gitlab/resource_gitlab_project_environment_test.go
```diff
@@ -0,0 +1,221 @@
...
+	conn := testAccProvider.Meta().(*gitlab.Client)
+	var project string
+	var environment int
+	var err error
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type == "gitlab_project" {
+			project = rs.Primary.ID
+			environment, err = strconv.Atoi(rs.Primary.ID)
+		} else if rs.Type == "gitlab_project_environment" {
+		}
```
armsnyder, line 171: Is this unfinished, or can we delete this empty if block?

gitlab/resource_gitlab_project_protected_environment.go
```diff
@@ -0,0 +1,211 @@
...
+							Optional:     true,
+							ValidateFunc: validation.StringIsNotEmpty,
+						},
+						"group_id": {
+							Type:         schema.TypeInt,
+							ForceNew:     true,
+							Optional:     true,
+							ValidateFunc: validation.StringIsNotEmpty,
+						},
+					},
```
armsnyder, line 63: The schema validation could be improved here.

1. `user_id` and `group_id` should probably use integer validations. Since you don't use these attributes in your acceptance test, there's a good chance the `validation.StringIsNotEmpty` validation has a bug here.

2. The API docs say that exactly one of `access_level`, `user_id`, and `group_id` must be specified. You can check for that easily using `ExactlyOneOf`.

3. Usually for enum attributes like `access_level` we validate against the list of valid access levels (see https://github.com/gitlabhq/terraform-provider-gitlab/blob/9884585ddf50efa061798e452c1e3375b8714af7/gitlab/resource_gitlab_project_membership.go#L42)

gitlab/resource_gitlab_project_protected_environment_test.go
```diff
@@ -0,0 +1,239 @@
...
+  name    = "ProjectProtectedEnvironment-%[1]d-this-suffix-matches-wildcard"
+}
+
+resource "gitlab_project_protected_environment" "ProjectProtectedEnvironment" {
+  depends_on  = [gitlab_project_environment.env]
+  project     = gitlab_project.foo.id
+  environment = "ProjectProtectedEnvironment-%[1]d%[2]s"
+  deploy_access_levels {
+    access_level = "maintainer"
+  }
```
armsnyder, line 236: Can you add test steps for:
1. Having more than once `deploy_access_levels` blocks
2. Using `group_id` and `user_id`.

armsnyder (changes requested):

gitlab/resource_gitlab_project_environment.go
sirlatrom: Stop:

> It returns 200 if the environment was successfully stopped, and 404 if the environment does not exist. 

https://docs.gitlab.com/ee/api/environments.html#stop-an-environment

Delete:

> It returns 204 if the environment was successfully deleted, and 404 if the environment does not exist.

https://docs.gitlab.com/ee/api/environments.html#delete-an-environment

gitlab/resource_gitlab_project_protected_environment.go
sirlatrom: Regarding point 2, when trying to set `ExactlyOneOf`, there is no way to achieve that if multiple blocks are expected to be supported:

>```
>ExactlyOneOf: []string{"deploy_access_levels.0.acces_level", "deploy_access_levels.0.user_id", "deploy_access_levels.0.group_id"},
>ExactlyOneOf: []string{"deploy_access_levels.0.acces_level", "deploy_access_levels.0.user_id", "deploy_access_levels.0.group_id"},
>ExactlyOneOf: []string{"deploy_access_levels.0.acces_level", "deploy_access_levels.0.user_id", "deploy_access_levels.0.group_id"},
>```

gitlab/resource_gitlab_project_environment.go
sirlatrom: Done in be501eb73c189081777f409af7cc2dd3b5ccc77a

gitlab/resource_gitlab_project_protected_environment_test.go
sirlatrom: Done in 4d9f680437b362b3a2e3071a9e0a1a362fc58672

sirlatrom:
The upstream PR has been merged, so now I'm only waiting on the next release before I can mark this as ready.

nicholasklick:
@sirlatrom looks like https://github.com/xanzy/go-gitlab/pull/1066 is merged.

Can we merge this MR now? Any blockers?

nicholasklick:
Any other concerns from @armsnyder?

gitlab/resource_gitlab_project_environment.go
```diff
@@ -0,0 +1,163 @@
...
+	environmentIDInt, err := strconv.Atoi(environmentID)
+
+	log.Printf("[DEBUG] Project %s read gitlab environment %d", project, environmentIDInt)
+
+	client := meta.(*gitlab.Client)
+
+	environment, resp, err := client.Environments.GetEnvironment(project, environmentIDInt)
+	if err != nil {
+		if resp.StatusCode == http.StatusNotFound {
```
mattkasa, line 92: Can we add a nil check for `resp` here?

```suggestion
		if resp != nil && resp.StatusCode == http.StatusNotFound {
```

gitlab/resource_gitlab_project_environment.go
```diff
@@ -0,0 +1,163 @@
...
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("error getting gitlab project %q environment %q: %v", project, environmentID, err)
+	}
+
+	d.SetId(buildTwoPartID(&project, gitlab.String(fmt.Sprintf("%v", environment.ID))))
+	d.Set("project", project)
+	d.Set("name", environment.Name)
+	d.Set("state", environment.State)
```
mattkasa, line 103: I know it's currently done this way all over the provider, but it would be good to handle `error` for `Set`:

```suggestion
	err = d.Set("project", project)
	if err != nil {
		return err
	}

	err = d.Set("name", environment.Name)
	if err != nil {
		return err
	}

	err = d.Set("state", environment.State)
	if err != nil {
		return err
	}
```

gitlab/resource_gitlab_project_protected_environment.go
```diff
@@ -0,0 +1,215 @@
...
+
+func resourceGitlabProjectProtectedEnvironmentRead(d *schema.ResourceData, meta interface{}) error {
+	log.Printf("[DEBUG] read gitlab protected environment %s", d.Id())
+
+	project, environment, err := parseTwoPartID(d.Id())
+	if err != nil {
+		return err
+	}
+	d.Set("project", project)
+	d.Set("environment", environment)
```
mattkasa, line 112: Would be good to handle `error` for `Set`:

```suggestion
	err = d.Set("project", project)
	if err != nil {
		return err
	}

	err = d.Set("environment", environment)
	if err != nil {
		return err
	}
```

gitlab/resource_gitlab_project_protected_environment.go
```diff
@@ -0,0 +1,215 @@
...
+	if err != nil {
+		if resp.StatusCode == http.StatusNotFound {
+			log.Printf("[DEBUG] Project %s gitlab protected environment %q not found", project, environment)
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("error getting gitlab project %q protected environment %q: %v", project, environment, err)
+	}
+
+	d.Set("environment", protectedEnvironment.Name)
```
mattkasa, line 128:
```suggestion
	err = d.Set("environment", protectedEnvironment.Name)
	if err != nil {
		return err
	}
```

gitlab/resource_gitlab_project_environment_test.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	env, response, err := conn.Environments.GetEnvironment(project, environment)
+	if err == nil {
+		if env != nil {
+			return fmt.Errorf("project Environment %v still exists", environment)
+		}
+	}
+	if response.StatusCode != 404 {
```
mattkasa, line 180:
```suggestion
	if response != nil && response.StatusCode != 404 {
```

gitlab/resource_gitlab_project_protected_environment.go
```diff
@@ -0,0 +1,215 @@
...
+
+	project := d.Get("project").(string)
+
+	log.Printf("[DEBUG] Project %s create gitlab protected environment %q", project, *options.Name)
+
+	client := meta.(*gitlab.Client)
+
+	protectedEnvironment, resp, err := client.ProtectedEnvironments.ProtectRepositoryEnvironments(project, &options)
+	if err != nil {
+		if resp.StatusCode == http.StatusNotFound {
```
mattkasa, line 93:
```suggestion
		if resp != nil && resp.StatusCode == http.StatusNotFound {
```

gitlab/resource_gitlab_project_protected_environment.go
```diff
@@ -0,0 +1,215 @@
...
+	d.Set("project", project)
+	d.Set("environment", environment)
+
+	log.Printf("[DEBUG] Project %s read gitlab protected environment %q", project, environment)
+
+	client := meta.(*gitlab.Client)
+
+	protectedEnvironment, resp, err := client.ProtectedEnvironments.GetProtectedEnvironment(project, environment)
+	if err != nil {
+		if resp.StatusCode == http.StatusNotFound {
```
mattkasa, line 120:
```suggestion
		if resp != nil && resp.StatusCode == http.StatusNotFound {
```

gitlab/resource_gitlab_project_protected_environment_test.go
```diff
@@ -0,0 +1,258 @@
...
+		}
+	}
+
+	pt, response, err := conn.ProtectedEnvironments.GetProtectedEnvironment(project, environment)
+	if err == nil {
+		if pt != nil {
+			return fmt.Errorf("project Protected Environment %s still exists", environment)
+		}
+	}
+	if response.StatusCode != 404 {
```
mattkasa, line 179:
```suggestion
	if response != nil && response.StatusCode != 404 {
```

mattkasa (changes requested):
@sirlatrom this looks really good! I left some suggestions, all error handling for `Set` or checking `resp != nil` before using `resp` as @armsnyder suggested, if those are ok with you I'd be happy to merge 👍

gitlab/resource_gitlab_project_environment.go
armsnyder: I'm gonna disagree here, as skipping error checking with primitive types is a regular practice in TF providers, since the SDK will catch these issues for us. Guidance for non-primitive types here: https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types

sirlatrom: @armsnyder I suppose that goes for the rest of the suggested error checks, too?

armsnyder: Yeah, but maybe wait until we come to some consensus in #651 😄

sirlatrom:
Sorry if my rebase on the master branch has broken the suggested code changes. I can see there's an ongoing discussion about the error checking of `d.Set(...)`. Until that is resolved, couldn't this PR follow the existing approach of not error checking when the argument is a primitive type, and then be handled by #651 after the fact?

gitlab/resource_gitlab_project_protected_environment.go
```diff
@@ -0,0 +1,215 @@
...
+	case "developer":
+		return gitlab.DeveloperPermissions, nil
+	case "maintainer":
+		return gitlab.MaintainerPermissions, nil
+	case "owner":
+		return gitlab.OwnerPermissions, nil
+	default:
+		return -1, fmt.Errorf("unknown access level %q", level)
+	}
+}
```
mattkasa, line 215: This seems to be duplicating `accessLevelNameToValue` (https://github.com/gitlabhq/terraform-provider-gitlab/blob/master/gitlab/util.go#L15), but with some differences (is `noone` correct, or `no one`?).

Would it make sense to fix/update `accessLevelNameToValue` instead?

mattkasa:
> couldn't this PR follow the existing approach of not error checking when the argument is a primitive type, and then be handled by #651 after the fact?

Absolutely 👍

mattkasa (commented):
This looks really good, just one question about `accessLevelStringToValue`.

gitlab/resource_gitlab_project_protected_environment.go
sirlatrom: Yes, certainly. I'll make the change, remove the above code and replace uses of it with `accessLevelNameToValue`.

docs/resources/project_protected_environment.md
```diff
@@ -0,0 +1,63 @@
...
+
+* `environment` - (Required) The name of the project protected environment, or a wildcard expression for which environments to protect.
+
+* `deploy_access_levels` - (Required) One or more `deploy_access_levels` blocks as defined below.
+
+---
+
+A `deploy_access_levels` block supports the following:
+
+* `access_level` - (Optional) The access level which is allowed to deploy to the protected environment. Valid values are `developer` or `maintainer`. Mutually exclusive with `user_id` and `group_id`.
```
mattkasa, line 51: > Valid values are `developer` or `maintainer`.

Sorry if I'm missing it, where is this being validated?

sirlatrom: Only server-side, I don't think I've added validation in the schema.

sirlatrom: I guess the note on the API docs means I can use StringInSlice for validating `developer`, `maintainer` or `admin`: https://docs.gitlab.com/ee/api/protected_environments.html#valid-access-levels

sirlatrom: Turns out I did validate it here, though it would also accept "reporter", etc.

https://github.com/gitlabhq/terraform-provider-gitlab/blob/2cc27a4e0f881ad64f590f58df7290f98c69f575/gitlab/resource_gitlab_project_protected_environment.go#L45-L50

Would you like me to change it to strictly allow `developer`, `maintainer` or `admin`? The UI only provides the first two as options.

mattkasa: @sirlatrom Yeah, I think it would be good to make the `ValidateFunc` match the documentation 👍

sirlatrom: That should be done now, and I've rebased on master, too.

campers:
@sirlatrom thanks for your work on this, was there any more work that needed to be done on this PR other than fixing the lint error?

sirlatrom:
> @sirlatrom thanks for your work on this, was there any more work that needed to be done on this PR other than fixing the lint error?

I cannot recall any other outstanding issues, but I've since moved on to another employer and will not be able to spend any more time on this.

tahir-gvs:
Any chance of this getting released soon?

PR #569, opened 2021-02-12
Fix inconsistent gitlab_project delete (fix master build failure)

### LINKED ISSUE #566: gitlab_project destroy doesn't wait for repository delete
Recent acceptance test failure: https://github.com/gitlabhq/terraform-provider-gitlab/pull/565/checks?check_run_id=1877859420

```
=== RUN   TestAccGitLabProjectApprovalRule_import
Error:     testing.go:745: Error destroying resource! WARNING: Dangling resources
        may exist. The full state and error is shown below.
        
        Error: Check failed: Repository still exists
        
        State: <no state>
--- FAIL: TestAccGitLabProjectApprovalRule_import (10.19s)
```

The test creates a gitlab_project, and the error is from the `testAccCheckGitlabProjectApprovalRuleDestroy` function, which calls `testAccCheckGitlabProjectDestroy`.

I think we need to do like #539 for gitlab_project.

armsnyder:
Fixes #566 

I ran the tests with `-count 10` to make sure this resolves the issue.

The bug is when `gitlab_project` is deleted, it can change state from being deleted to being undeleted.

The bug showed up in the `gitlab_project_import_rule` test because the test involves a lot of resources. I wasn't able to resolve the bug by messing with `gitlab_project_import_rule` alone.

Sidenote: Acceptance tests take ~5 minutes longer to run, but this can be improved by not including the `gitlab_project` resource in the test config, which should be our goal anyway in resolving #205

### DIFF
--- a/gitlab/resource_gitlab_project.go
+++ b/gitlab/resource_gitlab_project.go
@@ -686,28 +686,42 @@ func resourceGitlabProjectDelete(d *schema.ResourceData, meta interface{}) error
 
 	// Wait for the project to be deleted.
 	// Deleting a project in gitlab is async.
+	// There is also a failure mode where, after a delete, a GET project can return a 404, and then
+	// repeatedly a 200 afterward. So we must retry the DELETE and wait for a consistent response.
 	stateConf := &resource.StateChangeConf{
 		Pending: []string{"Deleting"},
 		Target:  []string{"Deleted"},
 		Refresh: func() (interface{}, string, error) {
+			// Check that the project was deleted or marked for deletion.
 			out, response, err := client.Projects.GetProject(d.Id(), nil)
+			if response != nil && response.StatusCode == http.StatusNotFound {
+				return "", "Deleted", nil
+			}
 			if err != nil {
-				if response.StatusCode == 404 {
-					return out, "Deleted", nil
-				}
-				log.Printf("[ERROR] Received error: %#v", err)
-				return out, "Error", err
+				return "", "Error", err
 			}
 			if out.MarkedForDeletionAt != nil {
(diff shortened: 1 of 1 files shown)


armsnyder:
@roidelapluie Do you think this fix is ok? It basically adds retry logic to deleting a project.

Tests are failing about 1/3 of the time on the latest Gitlab, so I'd like if we can address this soon. 😄

gitlab/resource_gitlab_project.go
```diff
@@ -686,28 +686,42 @@ func resourceGitlabProjectDelete(d *schema.ResourceData, meta interface{}) error
...
+				return "", "Error", err
 			}
 			if out.MarkedForDeletionAt != nil {
 				// Represents a Gitlab EE soft-delete
-				return out, "Deleted", nil
+				return "", "Deleted", nil
+			}
+
+			// If it is not deleted, try deleting it again.
+			response, err = client.Projects.DeleteProject(d.Id())
```
roidelapluie, line 709: Why??

roidelapluie:
It does not sound good .. this should not be needed right?

gitlab/resource_gitlab_project.go
armsnyder: Heh. I left details in the #569 issue comments. Sometimes a deleted project becomes un-deleted. That's what I observed, though I am not sure the root cause.

timofurrer:
@armsnyder @roidelapluie Is this still a thing? Does it make sense to discuss this in an issue first, to reduce stale PRs?

armsnyder:
Closing this as stale since to my knowledge it is not causing workflow errors anymore.

PR #572, opened 2021-02-16
Support the Note field in the User resource/data

ilpianista:
As per subject. ~~I had to bump the `go-gitlab` dependency version~~. Nope, you already did! :-)


armsnyder (approved):
Thanks, looks great!

PR #573, opened 2021-02-18
Mark gitlab_project.default_branch as Computed

sirlatrom:
This makes it possible for other resources, such as `gitlab_branch_protection`, to depend on knowing the default branch once created.

Example:

```tf
  namespace_id = 1234
  path         = "example"
  name         = "example"
}

resource "gitlab_branch_protection" "default" {
  project                      = gitlab_project.example.id
  branch                       = gitlab_project.example.default_branch
  merge_access_level           = "developer"
  push_access_level            = "maintainer"
}
```

If not marked as `Computed`, the `default_branch` attribute comes out as `""` instead when used in other resources, unless explicitly set by the user.

### DIFF
--- a/gitlab/resource_gitlab_project.go
+++ b/gitlab/resource_gitlab_project.go
@@ -44,6 +44,7 @@ var resourceGitLabProjectSchema = map[string]*schema.Schema{
 	"default_branch": {
 		Type:     schema.TypeString,
 		Optional: true,
+		Computed: true,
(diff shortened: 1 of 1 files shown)


sirlatrom:
I'm sorry, this was a misunderstanding. I'll close this and investigate if GitLab's API is eventually going to return a non-`null` value for `default_branch` for empty repositories.

PR #575, opened 2021-02-18
add diffSuppressFunc for 'expires_at' attribute in 'deploy_token'

### LINKED ISSUE #522: gitlab_deploy_token and expires_at
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version

```
terraform -v
Terraform v0.14.5
+ provider registry.terraform.io/gitlabhq/gitlab v3.4.0

terraform init -upgrade

Initializing the backend...

Initializing provider plugins...
- Finding gitlabhq/gitlab versions matching "3.4.0"...
- Installing gitlabhq/gitlab v3.4.0...
- Installed gitlabhq/gitlab v3.4.0 (signed by a HashiCorp partner, key ID BC097C3333027B14)

Partner and community providers are signed by their developers.
If you'd like to know more about provider signing, you can read about it here:
https://www.terraform.io/docs/plugins/signing.html
```

### Affected Resource(s)

- gitlab_deploy_token

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

```hcl
resource "gitlab_deploy_token" "PROJECT_token" {
  project    = gitlab_project.PROJECT.id
  name       = "Deploy Token for ABCD"
  username   = "deploy-token-0001"
  scopes     = ["read_repository"]
  expires_at = "2025-03-14T00:00:00.000Z"
}

```

### Expected Behaviour

Above "expires_at" is a copy-paste from your example +5 years in the future. This was ALWAYS triggering a recreation of the token as "expires_at" was changing. After lots of trial and error  found that `"2025-03-14T00:00:00Z"` (removing trailing ".000") does not trigger a recreation.

Guess that either the example should change or how date-time is handled.

RomanNess:
Fixes #522 by adding a `diffSuppressFunc` for `expires_at` attribute of resource `deploy_token` that compares the date strings as `time.Time`.

### DIFF
--- a/gitlab/resource_gitlab_deploy_token_test.go
+++ b/gitlab/resource_gitlab_deploy_token_test.go
@@ -199,10 +199,12 @@ func TestExpiresAtSuppressFunc(t *testing.T) {
 	}
 
 	for _, test := range testcases {
-		actual := expiresAtSuppressFunc("", test.old, test.new, nil)
-		if actual != test.expected {
-			t.Fatalf("FAIL %s - expiresAtSuppressFunc\n\told: %s, new: %s\n\texpected: %t\n\tactual: %t",
-				test.description, test.old, test.new, test.expected, actual)
-		}
+		t.Run(test.description, func(t *testing.T) {
+			actual := expiresAtSuppressFunc("", test.old, test.new, nil)
+			if actual != test.expected {
+				t.Fatalf("FAIL\n\told: %s, new: %s\n\texpected: %t\n\tactual: %t",
+					test.old, test.new, test.expected, actual)
+			}
+		})
 	}
 }
--- a/gitlab/resource_gitlab_deploy_token_test.go
+++ b/gitlab/resource_gitlab_deploy_token_test.go
@@ -156,7 +156,7 @@ resource "gitlab_deploy_token" "foo" {
   name     = "deployToken-%d"
   username = "my-username"
 
-  expires_at = "2021-03-14T07:20:50Z"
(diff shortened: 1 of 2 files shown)


gitlab/resource_gitlab_deploy_token_test.go
```diff
@@ -165,3 +165,44 @@ resource "gitlab_deploy_token" "foo" {
...
+			expected:    true,
+		}, {
+			description: "cannot parse new date",
+			old:         "2025-03-14T00:00:00Z",
+			new:         "invalid-date",
+			expected:    false,
+		},
+	}
+
+	for _, test := range testcases {
```
armsnyder, line 201: Small nitpick for next time you're writing a table test -- It's better to use `t.Run` to create a new subtest for each case in the table, so that all cases will be evaluated instead of stopping on the first failure. https://blog.golang.org/subtests#TOC_4.

armsnyder (approved):

gitlab/resource_gitlab_deploy_token_test.go
RomanNess: Thanks for the feedback. That's a valid nitpick. I'll update the PR later today.

RomanNess: done. :)

armsnyder:
Thanks for the contribution! :)

Issue #453, opened 2020-10-05
Ability to manage strict group/project members list


radkomateusz:
Hi, I want to submit feature request. 

I want to manage whole list of group/project members by single resource as I would like to ensure that terraform configure only given set of members. 

Expected behaviour:
Resource ensures that only given set of members are configured.
If additional members will be added manually, then TF plan/apply will remove them.
If members are removed manually, then TF plan/apply will add them again. 

### Proposed Terraform Configuration Files
```hcl
resource "gitlab_project_membership_binding" "test" {
  project_id   = "12345"
  user_ids      = [1337, 1338,1372]
  access_level = "maintainer"
}
```

I want to have similar concept here like in GCP provider (https://www.terraform.io/docs/providers/google/r/google_project_iam.html) - to be able manage: whole policy mapping, only binding for given access level, and also single membership (which is currently the only possible).

thuck:
I was just thinking on something similar; but using a slight different format for the code:
```
resource "gitlab_project_membership_binding" "test" {
  project_id       = "12345"
  member {
     user_id       = 1337
     access_level  = "maintainer"
  }
  member {
     user_id       = 1338
     access_level  = "developer"
  }
}
```
So when using this we could do something on the lines:
```
resource "gitlab_project_membership_binding" "test" {
  for_each          = local.projects
  project_id        = each.value.name
  dynamic member {
    for_each        = each.value.members
    content {
      user_id       = member.value.user_id
      access_level  = member.value.access_level
    }
  }
}
```

armsnyder:
@radkomateusz Thanks for the suggestion. However, we like to follow the [HashiCorp design principle that a resource should map directly to a resource in the upstream API](https://www.terraform.io/docs/extend/hashicorp-provider-design-principles.html#resources-should-represent-a-single-api-object). Since there isn't a GitLab API for adding multiple members, I don't think we'll have a resource for it in the provider.

Take a look at [count](https://www.terraform.io/docs/configuration/resources.html#count-multiple-resource-instances-by-count) or [for_each](https://www.terraform.io/docs/configuration/resources.html#for_each-multiple-resource-instances-defined-by-a-map-or-set-of-strings) for ways to create multiple resources dynamically based on a list of ids, as @thuck shared. 😄

radkomateusz:
@armsnyder Thanks for the answer! :) 

Although I understand mentioned principle, I still have an issue which couldn't be addressed by current resources. 

[...]

PR #593, opened 2021-03-23
Add new data source gitlab_project_membership

tommyknows:
This PR adds a new data source `gitlab_project_membership` which can get all members of a project.

Similar to the `group_membership`, but targeting projects. Also allows filtering by access level.


klausenbusk:
This sounds useful for: https://github.com/gitlabhq/terraform-provider-gitlab/issues/453#issuecomment-789781249.

@roidelapluie could you have a look?

gitlab/data_source_gitlab_project_membership.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/xanzy/go-gitlab"
+)
+
+func dataSourceGitlabProjectMembership() *schema.Resource {
+	acceptedAccessLevels := make([]string, 0, len(accessLevelID))
+	for k := range accessLevelID {
+		acceptedAccessLevels = append(acceptedAccessLevels, k)
+	}
+	return &schema.Resource{
+		Read: dataSourceGitlabProjectMembershipRead,
```
timofurrer, line 19: Please use the context-aware CRUD functions, see the [upgrade guide](https://www.terraform.io/plugin/sdkv2/guides/v2-upgrade-guide#more-support-for-context-context).

Also make sure to use the new `ctx` to pass it to the `gitlab` API functions as `gitlab.WithContext(ctx)`. It's usually the last parameter of the API functions.

gitlab/data_source_gitlab_project_membership.go
```diff
@@ -0,0 +1,159 @@
+package gitlab
+
+import (
+	"fmt"
+	"log"
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
```
timofurrer, line 9: Please upgrade to the [terraform SDK v2](https://www.terraform.io/plugin/sdkv2/sdkv2-intro) and make sure to rebase to an up-to-date `master` branch.

timofurrer (changes requested):
Please rebase to latest master. Otherwise LGTM!

timofurrer:
@tommyknows do you still want to work on this? I think it might be good candidate for the next release!

tommyknows:
Sorry for the delay. While I don't need this anymore, I'll be happy to have a look at rebasing this. Sounds like there's a bit more to do here, I'll try to tackle this until Monday!

tommyknows:
Not quite sure why the test is failing, but my assumption is that there's a user that's added by default to a project with the "maintainer" level?

(I expect that the terraform user that creates the test project is the `Owner`, so I'm not sure where that other user could be coming from; the group / user that contains this project, maybe?)

armsnyder (changes requested):
Something to note is that this new resource has an attribute `access_level`, which is not present on the upstream API. The API docs only mention a `query` attribute. 

https://docs.gitlab.com/ee/api/members.html#list-all-members-of-a-group-or-project

If this `query` parameter acts as a url-encoded query string, then I think it would be alright to have attributes on the resource that are used to form up that query parameter. However, as the code is now, it is doing the filtering after the API call. This is out-of-scope for what a provider should be responsible for.

Can you update the code to set the Query field? (https://github.com/xanzy/go-gitlab/blob/ae600d99fa047c58bc2de89f27d060f542601ed5/project_members.go#L39)

Unfortunately the API docs don't have an example of how to set this field, but either way, we shouldn't be adding custom fields to resources. (With some exceptions, like configuring terraform behavior.)

tommyknows:
> we shouldn't be adding custom fields to resources. (With some exceptions, like configuring terraform behavior.)

Just FYI: the group membership data source also supports such an attribute and does the [same filtering](https://github.com/gitlabhq/terraform-provider-gitlab/blob/main/internal/provider/data_source_gitlab_group_membership.go#L181), which is why I have added it. 

I'll be happy to remove it though.

However, I've also just seen why the test is failing:
> Due to [an issue](https://gitlab.com/gitlab-org/gitlab/-/issues/219299), projects in personal namespaces don’t show owner (50) permission for owner.

(from the docs you linked :) )

Instead, they're given / shown maintainer permissions, which explains the test failures.

armsnyder:
Thanks, I still think we should either remove the attribute or drive it using `query` instead of custom code, even though the code was copied from a similar resource.

Terraform users can always use Terraform functions to filter the output. So the filtering logic in the provider is doubly redundant. 😄

armsnyder:
For reasoning, see: https://www.terraform.io/plugin/hashicorp-provider-design-principles#resource-and-attribute-schema-should-closely-match-the-underlying-api

tommyknows:
I'm totally fine with that decision, just wanted to mention why I did it :-)

I'll remove the attribute then.

internal/provider/data_source_gitlab_project_membership.go
```diff
@@ -0,0 +1,188 @@
...
+			"> **Note**: exactly one of project_id or full_path must be provided.",
+
+		ReadContext: dataSourceGitlabProjectMembershipRead,
+		Schema: map[string]*schema.Schema{
+			"project_id": {
+				Description: "The ID of the project.",
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Optional:    true,
+				ConflictsWith: []string{
```
armsnyder, line 27: `ExactlyOneOf` would be more correct here, since it enforces that one must be set. See: https://github.com/gitlabhq/terraform-provider-gitlab/blob/b1521811a1e99391801e69b19287565a4918257d/internal/provider/resource_gitlab_deploy_token.go#L33

internal/provider/data_source_gitlab_project_membership.go
```diff
@@ -0,0 +1,188 @@
...
+			"username":     member.Username,
+			"name":         member.Name,
+			"state":        member.State,
+			"avatar_url":   member.AvatarURL,
+			"web_url":      member.WebURL,
+			"access_level": accessLevelValueToName[gitlab.AccessLevelValue(member.AccessLevel)],
+		}
+
+		if member.ExpiresAt != nil {
+			values["expires_at"] = member.ExpiresAt.String()
```
armsnyder, line 181:
```suggestion
			values["expires_at"] = member.ExpiresAt.Format(time.RFC3339)
```

(See docs of time.String() for why it should never be used outside of debugging)

armsnyder (changes requested):

internal/provider/data_source_gitlab_project_membership.go
tommyknows: Should I also update the group membership with this as well?

internal/provider/data_source_gitlab_project_membership.go
tommyknows: Should I also update the group membership with this as well?

tommyknows: Also, very interesting, I didn't know!

armsnyder: Let's keep the other resource as-is, since it might be a breaking change there. 😉

tommyknows: Actually, `members.ExpiresAt` is a `go-gitlab.ISOTime` type, which has the following `String` implementation:
```go
func (t ISOTime) String() string {
    return time.Time(t).Format(iso8601)
}
```

So it is stable, but with the wrong format :-)

armsnyder: Interesting! `gitlab.ISOTime` is a type alias of `time.Time`. Looking at the API docs - https://docs.gitlab.com/ee/api/members.html - it includes precision down to the second, but `iso8601` is just a date without time. So I would still recommend my suggestion.

tommyknows:
Alright, implemented your changes as suggested!

armsnyder (approved):
lgtm

Looking at ISOTime some more, it might be truncating time at midnight due to the custom unmarshal function.

@timofurrer Can you look and see if you agree with the use of RFC1113 here, regardless of this?

armsnyder:
We should confirm the format of the upstream `expires_at` with an API call. I'm out of time for today but can check back later.

armsnyder (changes requested):
Sorry for reversing course on this, but can you go back to calling `String()` on the time?

I checked the plain GitLab API and it does use yyyy-mm-dd format for the expiration:

```
curl -s -H 'PRIVATE-TOKEN: ACCTEST1234567890123' 'http://pc.local:8080/api/v4/projects/34/members' | jq
[
  {
    "id": 1,
    "username": "root",
    "name": "Administrator",
    "state": "active",
    "avatar_url": "https://www.gravatar.com/avatar/e64c7d89f26bd1972efa854d13d7dd61?s=80&d=identicon",
    "web_url": "http://1c405fe90a22/root",
    "access_level": 40,
    "created_at": "2022-01-25T04:44:05.084Z",
    "expires_at": null,
    "membership_state": "active"
  },
  {
    "id": 36,
    "username": "abcd",
    "name": "abcd",
    "state": "active",
    "avatar_url": "https://www.gravatar.com/avatar/9733471c0d68f18036ec45049f7646f4?s=80&d=identicon",
    "web_url": "http://1c405fe90a22/abcd",
    "access_level": 10,
    "created_at": "2022-03-08T09:45:14.762Z",
    "expires_at": "2022-03-09",
    "membership_state": "active"
  }
]
```

So we should match that format in the provider attribute, which would mean using `String()`

tommyknows:
Checked on a private instance as well, and it kinda makes sense that it's only day-specific: when creating a member with an expiry, one can also only select a date and not a specific time.

Reverted this back to `String` :-)

armsnyder (approved):

tommyknows:
Sorry for the delay, I'm always losing track of this! I updated the commit to fix the failing test, though it looks like they're still failing :( 

I'm not sure why - the pagination test uses the same userCount (`21`) as the group membership test, which seems to pass? And the projects test seems to fail as well, though I don't think my change should have influenced this?

internal/provider/data_source_gitlab_project_membership.go
```diff
@@ -0,0 +1,184 @@
...
+	d.Set("full_path", project.PathWithNamespace)
+
+	if err := d.Set("members", flattenGitlabProjectMembers(d, allPMs)); err != nil {
+		return diag.FromErr(err)
+	}
+
+	var optionsHash strings.Builder
+	optionsHash.WriteString(strconv.Itoa(project.ID))
+
+	if data, ok := d.GetOk("access_level"); ok {
```
timofurrer, line 153: `access_level` is no longer a thing in the schema, but probably `query` will become one, thus please change accordingly :)

internal/provider/data_source_gitlab_project_membership.go
```diff
@@ -0,0 +1,184 @@
...
+		optionsHash.WriteString(data.(string))
+	}
+
+	id := schema.HashString(optionsHash.String())
+	d.SetId(fmt.Sprintf("%d", id))
```
timofurrer, line 158: The id should be set first, so that if something goes wrong there isn't any dangling state and tf knows what belongs to where :)

internal/provider/data_source_gitlab_project_membership_test.go
```diff
@@ -0,0 +1,98 @@
...
+				Config: testAccDataSourceGitlabProjectMembershipPagination(project.ID),
+				Check:  resource.TestCheckResourceAttr("data.gitlab_project_membership.this", "members.#", fmt.Sprintf("%d", userCount)),
+			},
+		},
+	})
+}
+
+func testAccDataSourceGitlabProjectMembershipConfig(rInt int) string {
+	return fmt.Sprintf(`
+resource "gitlab_project" "foo" {
```
timofurrer, line 62: Let's only have what's "under test" in the test config and use the `helper_test.go` functions to create the necessary test setup.

timofurrer (changes requested):
Thanks for following up @tommyknows 🎉 

I've left some comments. I'm aware that you copied things I gave feedback for straight from the group membership data source. However, I think we should still make the improvements and then eventually refactor the group membership, too instead of adding tech debt. 

Also before continuing it would be sense that you rebase the latest `main` branch first the get the necessary changes for the pipeline etc :)

internal/provider/data_source_gitlab_project_membership.go
```diff
@@ -0,0 +1,184 @@
...
+				Computed:     true,
+				Optional:     true,
+				ExactlyOneOf: []string{"project_id", "full_path"},
+			},
+			"members": {
+				Description: "The list of project members.",
+				Type:        schema.TypeList,
+				Computed:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
```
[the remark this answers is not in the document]
tommyknows: The docs on the API say:
> The group_saml_identity attribute is only visible to a group owner for [SSO enabled groups](https://docs.gitlab.com/ee/user/group/saml_sso/index.html).

[source](https://docs.gitlab.com/ee/api/members.html#limitations)

The `membership_state` attribute is also present in the response, though as you noticed it is not in go-gitlab. 

Is there anything I need to do here?

internal/provider/data_source_gitlab_project_membership.go
```diff
@@ -0,0 +1,184 @@
...
+	// Get project memberships
+	listOptions := &gitlab.ListProjectMembersOptions{
+		ListOptions: gitlab.ListOptions{
+			PerPage: 20,
+			Page:    1,
+		},
+	}
+
+	listMembers := client.ProjectMembers.ListProjectMembers
+	if inherited, ok := d.GetOk("inherited"); ok && inherited.(bool) {
```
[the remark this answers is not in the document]
tommyknows: Oh good catch!

It's also not part of the API, technically (it's a different API call) and the `group_membership` data source doesn't have it either.

Shall I still add the field anyways?

internal/provider/data_source_gitlab_project_membership.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+	} else if fullPath, ok := d.GetOk("full_path"); ok {
+		// Get project by full path
+		project, _, err = client.Projects.GetProject(fullPath.(string), nil)
+		if err != nil {
+			return diag.FromErr(err)
+		}
+	} else {
+		return diag.Errorf("one and only one of project_id or full_path must be set")
+	}
```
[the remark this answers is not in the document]
tommyknows: Thanks, that's definitely cleaner!

internal/provider/data_source_gitlab_project_membership.go
tommyknows: changed the string to `query` :)

internal/provider/data_source_gitlab_project_membership.go
```diff
@@ -0,0 +1,184 @@
...
+			return diag.FromErr(err)
+		}
+
+		allPMs = append(allPMs, pms...)
+
+		if resp.NextPage == 0 {
+			break
+		}
+		listOptions.Page = resp.NextPage
+	}
```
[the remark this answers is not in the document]
tommyknows: That's a nice one!

Going one step further;
```go
for listOpts := &gitlab.ListProjectMembersOptions{
   ...
}; listOpts.Page != 0; listOpts.Page = resp.NextPage {
   ...
}
```

:smile:

internal/provider/data_source_gitlab_project_membership_test.go
tommyknows: Just to make sure I understand this correctly, you want me to move the function `testAccDataSourceGitlabProjectMembershipConfig` (that's a mouthful!) to the `helpers_test.go` file?

tommyknows:
Thanks for the thorough review and the nice feedback! should've caught way more things already, sorry about that :-) 

I'm awaiting your response for some of the comments, will push the updated PR afterwards.

internal/provider/data_source_gitlab_project_membership_test.go
timofurrer: Not exactly, but e.g. use the `testAccCreateProject()` function from that module to setup the project before the test steps. Something along the lines of this: 

https://github.com/gitlabhq/terraform-provider-gitlab/blob/d18aee439f62646265b5dd0b954f500449cc85a8/internal/provider/resource_gitlab_project_issue_test.go#L20

internal/provider/data_source_gitlab_project_membership.go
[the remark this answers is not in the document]
timofurrer: Yes, I think it's a nice feature :)

internal/provider/data_source_gitlab_project_membership.go
tommyknows: With "set first", do you mean right at the beginning of the function - right after we have the project id - or just before we write the other parts of the hash?

timofurrer: Just before writing the other states with `d.Set()`.

internal/provider/data_source_gitlab_project_membership.go
```diff
@@ -0,0 +1,198 @@
...
+	var pid interface{}
+	if v, ok := d.GetOk("project_id"); ok {
+		pid = v.(int)
+	} else if v, ok := d.GetOk("full_path"); ok {
+		pid = v.(string)
+	} else {
+		return diag.Errorf("one and only one of project_id or full_path must be set. This is a provider bug, please report upstream at https://github.com/gitlabhq/terraform-provider-gitlab/issues")
+	}
+
+	// Get project by full path
```
timofurrer, line 118:
```suggestion
	// Get project to have both, the `project_id` and `full_path` for setting to state
```

internal/provider/data_source_gitlab_project_membership.go
```diff
@@ -0,0 +1,198 @@
...
+
+	d.Set("project_id", project.ID)
+	d.Set("full_path", project.PathWithNamespace)
+
+	if err := d.Set("members", flattenGitlabProjectMembers(d, allPMs)); err != nil {
+		return diag.FromErr(err)
+	}
+
+	if data, ok := d.GetOk("query"); ok {
+		optionsHash.WriteString(data.(string))
```
timofurrer, line 171: You need to move this also before actually writing the id ...

tommyknows:
Looks like the tests have finally passed, so I guess this is ready for another look @timofurrer.

examples/data-sources/gitlab_project_membership/data-source.tf
```diff
@@ -0,0 +1,15 @@
...
+
+# By project's full path
+data "gitlab_project_membership" "example" {
+  full_path = "foo/bar"
+}
+
+# Only get members with access level "owner"
+data "gitlab_project_membership" "example" {
+  project_id   = 123
+  access_level = "owner"
```
armsnyder, line 14: Can you take another look at your examples? `access_level` isn't an attribute anymore.

armsnyder (changes requested):
Looks good! Just the examples need a refresher.

timofurrer:
I think this is still relevant :) I've just rebased and pushed, let's see. Since it's been a while I'll give it another review once the tests pass.

timofurrer (commented):
LGTM 🎉

timofurrer (approved):

PR #598, opened 2021-04-01
Update number of possible access level values

BanzaiMan:
https://github.com/gitlabhq/terraform-provider-gitlab/blob/620964c9a7f6705e54d915e6071424b8b8b2164d/gitlab/util.go#L205-L215

util.go indicates that there are six (seven, if you include the deprecated one)
possible values. It would be nice to have these valid values enumerated in
the documents, but that is for another day.


timofurrer:
Since this requires some major refactoring I created the issue #794 to follow up on this. @BanzaiMan you are more than welcome to tackle it again 🍰

PR #601, opened 2021-04-06
Fix links in CHANGELOG.md

floh96:
Links were returning a 404 because of the #.


nicholasklick:
@floh96 argh. Whoops. Good catch.

PR #634, opened 2021-06-10
Added gitlab_branch as a resource and data source 

eddb7:
I have moved this pr here as my branch in the previous PR was accidentally progressed.

Original message:
Hi I have been using this provider and required a way to automate and manage the creation and deletion of gitlab branches.
This allows the management of a gitlab_branch by adding it as a resource or data source block in terraform
The create, delete and read methods are using:
https://github.com/xanzy/go-gitlab/blob/master/branches.go#L188

The data structure for the resources is modelled on the response from
https://docs.gitlab.com/ee/api/branches.html#get-single-repository-branch

The properties around branch protection have been excluded for consistency, allowing this to work alongside the gitlab_branch_protection resources

It would be great to get some feedback and thoughts about wether this could potentially be integrated into the main package


gitlab/data_source_gitlab_branch.go
```diff
@@ -0,0 +1,133 @@
...
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
+			"merged": {
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
+			"commit": {
+				Type:     schema.TypeList,
+				MaxItems: 1,
```
armsnyder, line 41: https://github.com/bflad/tfproviderlint/tree/main/passes/S031

gitlab/data_source_gitlab_branch.go
```diff
@@ -0,0 +1,133 @@
...
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
```
armsnyder, line 44: What is the purpose of the optional attributes under here, if the schema is computed?

gitlab/data_source_gitlab_branch.go
```diff
@@ -0,0 +1,133 @@
...
+	client := meta.(*gitlab.Client)
+	name := d.Get("name").(string)
+	project := d.Get("project").(string)
+	log.Printf("[DEBUG] read gitlab branch %s", name)
+	branch, resp, err := client.Branches.GetBranch(project, name)
+	if err != nil {
+		log.Printf("[DEBUG] failed to read gitlab branch %s response %v", name, resp)
+		return err
+	}
+	d.SetId(fmt.Sprintf("%s-%s", project, name))
```
armsnyder, line 118: Use `buildTwoPartID`, which has `:` as a delimiter.

gitlab/resource_gitlab_branch.go
```diff
@@ -0,0 +1,181 @@
...
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+			},
+			"project": {
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+			},
+			"ref": {
```
armsnyder, line 26: This should be required, with no default, in order to match the API. https://docs.gitlab.com/ee/api/branches.html#create-repository-branch Especially since newer repositories use `main` as a default.

gitlab/resource_gitlab_branch.go
```diff
@@ -0,0 +1,181 @@
...
+				Computed: true,
+			},
+			"merged": {
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
+			"commit": {
+				Type:     schema.TypeList,
+				Computed: true,
+				MaxItems: 1,
```
armsnyder, line 51: https://github.com/bflad/tfproviderlint/tree/main/passes/S031

gitlab/resource_gitlab_branch.go
```diff
@@ -0,0 +1,181 @@
...
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
```
armsnyder, line 53: Same as with the datasource, I'm not sure why these are optional.

gitlab/resource_gitlab_branch.go
```diff
@@ -0,0 +1,181 @@
...
+	name := d.Get("name").(string)
+	project := d.Get("project").(string)
+	ref := d.Get("ref").(string)
+	log.Printf("[DEBUG] read gitlab branch %s", name)
+	branch, resp, err := client.Branches.GetBranch(project, name)
+	if err != nil {
+		log.Printf("[DEBUG] failed to read gitlab branch %s response %v", name, resp)
+		return err
+	}
+	d.SetId(fmt.Sprintf("%s-%s", project, name))
```
armsnyder, line 139: Same as data source, can you build the ID with a `:`?

gitlab/resource_gitlab_branch.go
```diff
@@ -0,0 +1,181 @@
...
+
+import (
+	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	gitlab "github.com/xanzy/go-gitlab"
+	"log"
+)
+
+func resourceGitlabBranch() *schema.Resource {
+	return &schema.Resource{
```
armsnyder, line 11: Can you add an importer?

gitlab/resource_gitlab_branch.go
```diff
@@ -0,0 +1,181 @@
...
+	return resourceGitlabBranchRead(d, meta)
+}
+
+func resourceGitlabBranchRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gitlab.Client)
+	name := d.Get("name").(string)
+	project := d.Get("project").(string)
+	ref := d.Get("ref").(string)
+	log.Printf("[DEBUG] read gitlab branch %s", name)
+	branch, resp, err := client.Branches.GetBranch(project, name)
```
armsnyder, line 134: Please handle a 404 error here, and set the ID to `""` in that case. This ensures that the branch can be safely removed inside GitLab out-of-band. See https://learn.hashicorp.com/tutorials/terraform/provider-setup#implement-read

armsnyder (changes requested):
Thanks for the contribution! I left some comments.

ric-bguichard:
Any news for this merge request ?

eddb7:
Thanks for the review I will look to implement those changes soon @armsnyder

gitlab/resource_gitlab_branch.go
eddb7: I have had to leave this as optional, it is not ideal but the reason being is there is no real way to determine the ref of the branch after creation as the api does not return this in either the branches or commit apis. This then makes it near impossible to import it while this is required with no default as it will always flag a new value and attempt to destroy the imported branch upon terraform apply. If there is a workaround please let me know but for now I have added "main" as the default value and reverted to it being an optional property.

ThePlenkov:
Hi! I was surprised finding that we have so much resources available and branch is not there. Especially having this Gitlab flow as methodology: https://docs.gitlab.com/ee/topics/gitlab_flow.html Do you know why it was not included into official delivery?  What is the proposed way to initialize a project and create branches? Thank you!

armsnyder:
@eddb7 Could you rebase your PR onto latest master please?

eddb7:
> @eddb7 Could you rebase your PR onto latest master please?

Hi yes this is done now

gitlab/data_source_gitlab_branch.go
```diff
@@ -0,0 +1,74 @@
+package gitlab
+
+import (
+	"log"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
```
timofurrer, line 6: Please upgrade to the [terraform SDK v2](https://www.terraform.io/plugin/sdkv2/sdkv2-intro) and make sure to rebase to an up-to-date `master` branch.

gitlab/data_source_gitlab_branch.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/xanzy/go-gitlab"
+)
+
+func dataSourceGitlabBranch() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceGitlabBranchRead,
```
timofurrer, line 12: Please use the context-aware CRUD functions, see the [upgrade guide](https://www.terraform.io/plugin/sdkv2/guides/v2-upgrade-guide#more-support-for-context-context).

Also make sure to use the new `ctx` to pass it to the `gitlab` API functions as `gitlab.WithContext(ctx)`. It's usually the last parameter of the API functions.

timofurrer (changes requested):
LGTM from use-case and features as others already mentioned.

Please rebase to latest `master` and have a look at the comments I've left.

timofurrer:
@eddb7 dang it, it's been a while and I completely forgot about this one here. Sorry, my bad.

I've just compared that we have here and in #863 and to me this PR here is a little bit more complete with respect to upstream API alignment, also we have the data source.

@eddb7 and @Moon1706: IMHO, it makes sense to further work on #634 and close #863, in case, @eddb7, you are still working on this here. Do you both agree?

I'm sorry for the confusion I have caused ✌️

eddb7:
> @eddb7 dang it, it's been a while and I completely forgot about this one here. Sorry, my bad.
> 
> I've just compared that we have here and in #863 and to me this PR here is a little bit more complete with respect to upstream API alignment, also we have the data source.
> 
> @eddb7 and @Moon1706: IMHO, it makes sense to further work on #634 and close #863, in case, @eddb7, you are still working on this here. Do you both agree?
> 
> I'm sorry for the confusion I have caused ✌️

No worries, i'm happy to keep working on this just didn't want to waste time, I should be able to find the time to finish those bits before monday/tuesday next week, cheers for the speedy reply

Moon1706:
> > @eddb7 dang it, it's been a while and I completely forgot about this one here. Sorry, my bad.
> > I've just compared that we have here and in #863 and to me this PR here is a little bit more complete with respect to upstream API alignment, also we have the data source.
> > @eddb7 and @Moon1706: IMHO, it makes sense to further work on #634 and close #863, in case, @eddb7, you are still working on this here. Do you both agree?
> > I'm sorry for the confusion I have caused ✌️
> 
> No worries, i'm happy to keep working on this just didn't want to waste time, I should be able to find the time to finish those bits before monday/tuesday next week, cheers for the speedy reply

In present time for me most important is ready functionality. I want use this feature. If resource will be finished in monday/tuesday next week all will cool. Thank you @eddb7 !

timofurrer:
@eddb7 I think that something went wrong during your rebase, can you please have a look?

eddb7:
Hey guys, sorry this is taking a while to finish off finding a method to get the ref consistently from the api seems near impossible still, wonder if anyone has any solutions on to how to get round this?  I was thinking to make a three part ID but that would mean the data source would also need to have ref defined which isn't ideal for it's usability. If there is a way to pass the set value for ref into the read method consistently other than using ID that would also let us get round this issue

eddb7:
I think this is done, although the solution for getting the ref for the branch on the Read method for the branch isn't ideal but seems to be all we can do from the api at the moment, let me know what you guys think, very open to other solutions for ref / if there is a way to consistently use a method like d.Get("ref").(string) that would return a value even during imports I would say that would be more secure than checking it against api in current state

timofurrer (changes requested):
I'm actually on vacation and will review in more detail beginning next week.

However, from looking quickly at the changes a few things:

* the docs are generated from the `Description`s fields. You can run `make reviewable` to generate those. Also the examples are coming from the `examples/` directory (incl. import) Please make use of it :) Check out the other resources and data sources for examples.
* please remove the obsolete `gitlab/provider.go` file
* would be nice if you could already clean / squash your commit history to max a handful of meaningful commits.

As I've said, I'll review in much more detail beginning next week 🎉

eddb7:
> I'm actually on vacation and will review in more detail beginning next week.
> 
> However, from looking quickly at the changes a few things:
> 
> * the docs are generated from the `Description`s fields. You can run `make reviewable` to generate those. Also the examples are coming from the `examples/` directory (incl. import) Please make use of it :) Check out the other resources and data sources for examples.
> * please remove the obsolete `gitlab/provider.go` file
> * would be nice if you could already clean / squash your commit history to max a handful of meaningful commits.
> 
> As I've said, I'll review in much more detail beginning next week 🎉

I've done those bits now thanks for taking time during your holiday to progress this. I haven't squashed commits from before the recent rebase as this seems a fair bit of hassle right now, preferably can just squash this branches commits on merge, if this is an issue it is probably easier for me to create a new branch but let me know. 
I look forward to your full review next week!

internal/provider/data_source_gitlab_branch.go
```diff
@@ -0,0 +1,101 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/xanzy/go-gitlab"
+)
+
+var _ = registerDataSource("gitlab_branch", func() *schema.Resource {
+	return &schema.Resource{
+		Description: "Provide details about a gitlab project branch \n\n" +
+			"> **Note**: name and project must be specified.",
```
timofurrer, line 15: I think we can avoid that comment. The `required` arguments are documented in a well-known format already :)

internal/provider/data_source_gitlab_branch.go
```diff
@@ -0,0 +1,101 @@
...
+
+		ReadContext: dataSourceGitlabBranchRead,
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Description: "The name of the branch.",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"project": {
+				Description: "The full path or id of the group.",
```
timofurrer, line 25:
```suggestion
				Description: "The full path or id of the project.",
```

internal/provider/resource_gitlab_branch.go
```diff
@@ -0,0 +1,236 @@
...
+				Elem:        commitSchema,
+			},
+		},
+	}
+})
+
+var commitSchema = &schema.Resource{
+	Schema: map[string]*schema.Schema{
+		"id": {
+			Description: "The unique id assigned to the commit by gitlab.",
```
timofurrer, line 75:
```suggestion
			Description: "The unique id assigned to the commit by GitLab.",
```

internal/provider/resource_gitlab_branch.go
```diff
@@ -0,0 +1,236 @@
...
+			Type:        schema.TypeString,
+			Computed:    true,
+		},
+		"author_name": {
+			Description: "The name of the author.",
+			Type:        schema.TypeString,
+			Computed:    true,
+		},
+		"authored_date": {
+			Description: "The date which the commit was authored",
```
timofurrer, line 90: Let's specify the format in which it is returned.

internal/provider/resource_gitlab_branch.go
```diff
@@ -0,0 +1,236 @@
...
+			Type:        schema.TypeString,
+			Computed:    true,
+		},
+		"committer_name": {
+			Description: "The name of the user that committed.",
+			Type:        schema.TypeString,
+			Computed:    true,
+		},
+		"short_id": {
+			Description: "The short id assigned to the commit by gitlab",
```
timofurrer, line 110:
```suggestion
			Description: "The short id assigned to the commit by GitLab.",
```

internal/provider/resource_gitlab_branch.go
```diff
@@ -0,0 +1,236 @@
...
+			Type:        schema.TypeString,
+			Computed:    true,
+		},
+		"title": {
+			Description: "The title of the commit",
```
timofurrer, line 115:
```suggestion
			Description: "The title of the commit.",
```

internal/provider/resource_gitlab_branch.go
```diff
@@ -0,0 +1,236 @@
...
+			Type:        schema.TypeString,
+			Computed:    true,
+		},
+		"message": {
+			Description: "The commit message",
```
timofurrer, line 120:
```suggestion
			Description: "The commit message.",
```

internal/provider/resource_gitlab_branch.go
```diff
@@ -0,0 +1,236 @@
...
+			Type:        schema.TypeString,
+			Computed:    true,
+		},
+		"parent_ids": {
+			Description: "The id of the parents of the commit",
```
timofurrer, line 125:
```suggestion
			Description: "The id of the parents of the commit.",
```

internal/provider/resource_gitlab_branch.go
```diff
@@ -0,0 +1,236 @@
...
+
+		CreateContext: resourceGitlabBranchCreate,
+		ReadContext:   resourceGitlabBranchRead,
+		DeleteContext: resourceGitlabBranchDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Description: "The name for this branch",
```
timofurrer, line 24: Please end the descriptions with a `.`

internal/provider/resource_gitlab_branch.go
```diff
@@ -0,0 +1,236 @@
...
+	return &schema.Resource{
+		Description: "This resource allows you to create and manage GitLab branches.",
+
+		CreateContext: resourceGitlabBranchCreate,
+		ReadContext:   resourceGitlabBranchRead,
+		DeleteContext: resourceGitlabBranchDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		Schema: map[string]*schema.Schema{
```
timofurrer, line 22: The schema is missing the fields `developers_can_merge` and `developers_can_push`.

internal/provider/data_source_gitlab_branch.go
```diff
@@ -0,0 +1,101 @@
...
+	d.Set("project", project)
+	d.Set("web_url", branch.WebURL)
+	d.Set("default", branch.Default)
+	d.Set("can_push", branch.CanPush)
+	d.Set("protected", branch.Protected)
+	d.Set("merged", branch.Merged)
+	d.Set("developer_can_merge", branch.DevelopersCanMerge)
+	d.Set("developer_can_push", branch.DevelopersCanPush)
+	commit := flattenCommit(branch.Commit)
+	if err := d.Set("commit", commit); err != nil {
```
timofurrer, line 97: NIT:

```suggestion
	if err := d.Set("commit", flattenCommit(branch.Commit)); err != nil {
```

internal/provider/data_source_gitlab_branch_test.go
```diff
@@ -0,0 +1,110 @@
...
+%s
+
+data "gitlab_branch" "foo" {
+  name = "${gitlab_branch.foo.name}"
+  project = "${gitlab_branch.foo.project}"
+}
+`, testAccDataGitlabBranchSetup(rInt))
+}
+
+func testAccDataGitlabBranchSetup(rInt int) string {
```
timofurrer, line 88: Let's focus the test on the data source / resource being tested.

It would be nice to use the `testGitlabClient` directly for setup code. See the helper functions in `helper_test.go` for examples.

timofurrer: I also assume that depending on the execution order the `data.gitlab_branch.foo` wouldn't find the branch ...

internal/provider/data_source_gitlab_branch_test.go
```diff
@@ -0,0 +1,110 @@
...
+		for _, attribute := range testAttributes {
+			if searchAttr[attribute] != branchAttr[attribute] {
+				return fmt.Errorf("expected branch's parameter `%s` to be: %s, but got: `%s`", attribute, branchAttr[attribute], searchAttr[attribute])
+			}
+		}
+		return nil
+	}
+}
+
+func testAccDataSourceGitlabBranchProtection(n string, exp testAccGitlabBranchProtectionAttributes) resource.TestCheckFunc {
```
timofurrer, line 61: What's the idea of having a separate function for the branch protection fields?

internal/provider/resource_gitlab_branch.go
```diff
@@ -0,0 +1,236 @@
...
+	client := meta.(*gitlab.Client)
+	project, name, err := parseTwoPartID(d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	log.Printf("[DEBUG] read gitlab branch %s", name)
+	branch, resp, err := client.Branches.GetBranch(project, name, gitlab.WithContext(ctx))
+	if err != nil {
+		if resp.StatusCode == 404 {
```
timofurrer, line 163: Please use the `is404(err)` helper function from `util.go` :)

internal/provider/resource_gitlab_branch.go
```diff
@@ -0,0 +1,236 @@
...
+	d.Set("ref", ref)
+	d.SetId(buildTwoPartID(&project, &name))
+	d.Set("name", branch.Name)
+	d.Set("project", project)
+	d.Set("web_url", branch.WebURL)
+	d.Set("default", branch.Default)
+	d.Set("can_push", branch.CanPush)
+	d.Set("merged", branch.Merged)
+	commit := flattenCommit(branch.Commit)
+	if err := d.Set("commit", commit); err != nil {
```
timofurrer, line 196: NIT:

```suggestion
	if err := d.Set("commit", flattenCommit(branch.Commit)); err != nil {
```

internal/provider/resource_gitlab_branch.go
```diff
@@ -0,0 +1,236 @@
...
+	ref := d.Get("ref").(string)
+	if ref == "" {
+		for _, br := range commitRefs {
+			if br.Name != branch.Name {
+				ref = br.Name
+				break
+			}
+		}
+	}
+	d.Set("ref", ref)
```
timofurrer, line 187: What's the idea here? 

Can't we use the `commit.id` field as `ref` ?

timofurrer (changes requested):
In general, this looks good to me :) I've added a few comments about style issues (for which we don't have linters yet, I'm sorry) and smaller things.

internal/provider/data_source_gitlab_branch_test.go
eddb7: > I also assume that depending on the execution order the `data.gitlab_branch.foo` wouldn't find the branch ...

Do you mean we should test the case where it doesnt find a branch?? It wouldnt be able to execute in any other order due to the dependencies in the terraform configuration itself ?

internal/provider/resource_gitlab_branch.go
eddb7: Nope then if the user defined a ref like "main" or "develop" it would be overwritten every time by the id from the commit and forcce replacement of the resources, one of the biggest issues with this feature is gitlab doesnt provide a consistent way to get the ref via the api but still requires this field for branch creation which I mentioned in some comments above -- also why originally my solution used ref with a default field rather than a required -- happy to revert back to this but this was also mentioned by another reviewer -- to me there is currently no perfect solution as the data source does not need the ref to find a branch otherwise I would suggest using something along the lines of a 3 part id and just using the defined ref from configuration rather than ever reading it from the api

internal/provider/data_source_gitlab_branch_test.go
timofurrer: Yes, just noticed. Nevermind ;)

Still the first comment applies nonetheless :)

internal/provider/resource_gitlab_branch.go
timofurrer: I see ... maybe we can also address this upstream.

In the meanwhile the currently implementation is okay for me. However, I think we can refactor it a little to only get `commitRefs` in case the `ref` is empty.

internal/provider/data_source_gitlab_branch_test.go
eddb7: Do you mean create the project and group through direct api creation methods using the helper rather than via terraform config in the tests, I just followed how the other tests were written generally -- personally think it's better this way as it's self documenting in how to use the data source / resource in an actual terraform block -- also potentially a better integration test than just assuming it works well with other terraform resources and going direct via api

timofurrer: I agree to these upsides, however, this test also breaks if there is a bug in let's say the `gitlab_project` resource, which really is suboptimal. I'd rather want the test itself to fail for as few reasons as possible w.r.t to production code (of course there may be issues in the test code itself, too).

internal/provider/resource_gitlab_branch.go
eddb7: Agreed, if any changes could be made on upstream that would be ideal but will implement this suggestion for now

internal/provider/data_source_gitlab_branch_test.go
eddb7: Without it I get this error:
https://github.com/gitlabhq/terraform-provider-gitlab/runs/5274065200?check_suite_focus=true#step:6:8

eddb7: It was a bug because I initially didn't have the protection fields on the resource so the lookup between datasource : resource in this test was not matching for those fields -- added this to the resource now and regenerated docs

internal/provider/data_source_gitlab_branch_test.go
eddb7: --- FAIL: TestAccDataGitlabBranch_basic (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x2d8 pc=0x180a06e]

goroutine 20 [running]:
testing.tRunner.func1.2({0x1953860, 0x22a53f0})
        /usr/local/go/src/testing/testing.go:1209 +0x24e
testing.tRunner.func1()
        /usr/local/go/src/testing/testing.go:1212 +0x218
panic({0x1953860, 0x22a53f0})
        /usr/local/go/src/runtime/panic.go:1038 +0x215
github.com/gitlabhq/terraform-provider-gitlab/internal/provider.testAccCreateProject(0xc00009b860)
        /Users/Selecta/Documents/atosGO/terraform-provider-gitlab/internal/provider/helper_test.go:103 +0x16e
github.com/gitlabhq/terraform-provider-gitlab/internal/provider.TestAccDataGitlabBranch_basic(0xc00009b860)
        /Users/Selecta/Documents/atosGO/terraform-provider-gitlab/internal/provider/data_source_gitlab_branch_test.go:14 +0x4c
testing.tRunner(0xc00009b860, 0x1b01348)
        /usr/local/go/src/testing/testing.go:1259 +0x102
created by testing.(*T).Run
        /usr/local/go/src/testing/testing.go:1306 +0x35a
FAIL    github.com/gitlabhq/terraform-provider-gitlab/internal/provider 0.293s
FAIL
make: *** [test] Error 1

internal/provider/resource_gitlab_branch_test.go
```diff
@@ -0,0 +1,186 @@
...
+	rInt := acctest.RandInt()
+	testAccCheck(t)
+	project := testAccCreateProject(t)
+	fooBranchName := fmt.Sprintf("testbranch-%d", rInt)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: providerFactories,
+		CheckDestroy:      testAccCheckGitlabBranchDestroy,
+		Steps: []resource.TestStep{
```
timofurrer, line 25: Can you add some more tests to test updating the resource?

timofurrer (changes requested):

internal/provider/resource_gitlab_branch.go
```diff
@@ -0,0 +1,254 @@
...
+				Computed:    true,
+			},
+			"developer_can_push": {
+				Description: "Bool, true if developer level access allows git push.",
+				Type:        schema.TypeBool,
+				Computed:    true,
+			},
+			"commit": {
+				Description: "The commit associated with the branch ref.",
+				Type:        schema.TypeSet,
```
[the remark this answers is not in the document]
eddb7: I get this lint error if I add MaxItems --  schema should use TypeList with MaxItems 1. I think generally the framework was pretty fussy to get to work with commits in a list if I remember right, unless that has changed as I wrote most of this 1 year ago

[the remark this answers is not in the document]
timofurrer: Ah yeah, true!

timofurrer (dismissed):
LGTM

internal/provider/resource_gitlab_branch.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+	return []map[string]interface{}{
+		{
+			"id":              commit.ID,
+			"short_id":        commit.ShortID,
+			"title":           commit.Title,
+			"author_name":     commit.AuthorName,
+			"author_email":    commit.AuthorEmail,
+			"authored_date":   commit.AuthoredDate.String(),
+			"committed_date":  commit.CommittedDate.String(),
```
armsnyder, line 247: These should use the format of the GitLab API.
```suggestion
			"authored_date":   commit.AuthoredDate.Format(time.RFC3339),
			"committed_date":  commit.CommittedDate.Format(time.RFC3339),
```

internal/provider/resource_gitlab_branch.go
```diff
@@ -0,0 +1,254 @@
...
+			return diag.FromErr(err)
+		}
+		for _, br := range commitRefs {
+			if br.Name != branch.Name {
+				ref = br.Name
+				break
+			}
+		}
+	}
+	d.Set("ref", ref)
```
armsnyder, line 202: Isn't `ref` only known during creation time, since it is the commit or branch that you use to create this branch? I think you should move `d.Set("ref", ref)` to the CREATE function, and delete this code in the READ function. Note you will also need to set `ref` as an ignored attribute with [ImportStateVerifyIgnore](https://www.terraform.io/plugin/sdkv2/testing/acceptance-tests/teststep#importstateverifyignore) in your `Test ImportState` test step.

internal/provider/resource_gitlab_branch_test.go
```diff
@@ -0,0 +1,205 @@
...
+		branch, resp, err := testGitlabClient.Branches.GetBranch(project, name)
+		if err == nil {
+			if branch != nil && branch.Name == name {
+				return fmt.Errorf("Branch still exists")
+			}
+		}
+		if resp.StatusCode != 404 {
+			return err
+		}
+		return nil
```
armsnyder, line 115: Code here can be simplified. You don't need to check the branch name, since you specified it in the GET request.

```suggestion
		if err != nil {
			if is404(err) {
				return nil
			}
			return err
		}
		return errors.New("branch still exists")
```

internal/provider/resource_gitlab_branch_test.go
```diff
@@ -0,0 +1,205 @@
...
+		_, branchName, err := parseTwoPartID(rs.Primary.ID)
+		if err != nil {
+			return errors.New("Error in splitting project and branch IDs")
+		}
+		if branch.WebURL == "" {
+			return errors.New("got empty web url")
+		}
+		if branchName != want.Name || branchName != branch.Name {
+			return fmt.Errorf("got name %s; want %s", branch.Name, want.Name)
+		}
```
armsnyder, line 136: I'm probably forgetting some context, but why parse the ID? Shouldn't we always expect the branch name to equal the `want.Name`? How about this:

```suggestion
		if branch.WebURL == "" {
			return errors.New("got empty web url")
		}
		if branch.Name != want.Name {
			return fmt.Errorf("got name %s; want %s", branch.Name, want.Name)
		}
```

armsnyder (changes requested):
Apologies if any of these comments have already been addressed in prior feedback. 😄 This looks nearly finished!

internal/provider/resource_gitlab_branch.go
eddb7: I will try out implementing this as sounds slightly more watertight than the current solution thanks

armsnyder (approved):

internal/provider/data_source_gitlab_branch_test.go
armsnyder: Looks resolved

internal/provider/resource_gitlab_branch.go
armsnyder: looks resolved

internal/provider/data_source_gitlab_branch.go
armsnyder: looks resolved to me

internal/provider/resource_gitlab_branch.go
armsnyder: This is resolved

timofurrer (approved):

Issue #487, opened 2020-11-19
Support badges


declension:
First - thanks for a great project - it's changed how we manage lots of things :+1: 

Badges are a great visual way of indicating status at a glance on larger projects. 

The [Gitlab API has supported badges](https://docs.gitlab.com/ee/api/project_badges.html) for a while, so I'd love for this to be configurable in terraform.

LipingSun:
Any plan to support this feature?

LipingSun:
@armsnyder Any update? It would be useful to have this feature supported.

victorboissiere:
We would also love to have this feature 👍 I'll take a look if I can find some time to directly make the PR.
Which way could work best?
- Add a new variable in the project resource
- Manage it as a new standalone resource

mekras:
I can't imagine what it would look like with a variable. Probably the resource will be better. The main thing is that we can not only add and change them, but also delete them.

Here is an example of badges in one of our projects:

![изображение](https://user-images.githubusercontent.com/192067/121399764-6d6f0300-c95f-11eb-9480-cb24095a31f6.png)

fandujar:
@armsnyder I think we can close this issue even tho we don't have group badges support yet

fandujar:
@declension @LipingSun @victorboissiere @mekras you can find project badge support on version 3.7.0 released hours ago.

armsnyder:
> I think we can close this issue even tho we don't have group badges support yet

Agreed. We'd of course like to have full API coverage in the provider, but until then if anyone needs group badges specifically please open a new issue to help prioritize. 😄

PR #648, opened 2021-07-14
Adds support for gitlab project badges

fandujar:
This PR adds support for gitlab project badges mentioned here https://github.com/gitlabhq/terraform-provider-gitlab/issues/487

Schemas were modeled from here https://docs.gitlab.com/ee/api/project_badges.html

### DIFF
--- a/docs/resources/project_badge.md
+++ b/docs/resources/project_badge.md
@@ -1,0 +1,39 @@
+# gitlab\_project\_badge
+
+This resource allows you to create and manage badges for your GitLab projects.
+For further information on hooks, consult the [gitlab
+documentation](https://docs.gitlab.com/ce/user/project/badges.html).
+
+
+## Example Usage
+
+```hcl
+resource "gitlab_project" "foo" {
+  name = "foo-project"
+}
+
+resource "gitlab_project_badge" "example" {
+  project   = gitlab_project.foo.id
+  link_url  = "https://example.com/badge-123"
+  image_url = "https://example.com/badge-123.svg"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `project` - (Required) The id of the project to add the badge to.
+
+* `link_url` - (Required) The url linked with the badge.
+
+* `image_url` - (Required) The image url which will be presented on project overview.
+
+
+## Attributes Reference
+
+The resource exports the following attributes:
+
+* `rendered_link_url` - The link_url argument rendered (in case of use of placeholders).
+
+* `rendered_image_url` - The image_url argument rendered (in case of use of placeholders).
--- a/docs/resources/project_badge.md
+++ b/docs/resources/project_badge.md
@@ -4,7 +4,6 @@ This resource allows you to create and manage badges for your GitLab projects.
 For further information on hooks, consult the [gitlab
 documentation](https://docs.gitlab.com/ce/user/project/badges.html).
 
-
 ## Example Usage
 
 ```hcl
@@ -28,7 +27,6 @@ The following arguments are supported:
 * `link_url` - (Required) The url linked with the badge.
 
 * `image_url` - (Required) The image url which will be presented on project overview.
-
 
 ## Attributes Reference
 
--- a/docs/resources/project_badge.md
+++ b/docs/resources/project_badge.md
@@ -35,3 +35,12 @@ The resource exports the following attributes:
 * `rendered_link_url` - The link_url argument rendered (in case of use of placeholders).
 
 * `rendered_image_url` - The image_url argument rendered (in case of use of placeholders).
+
+## Import
+
+GitLab project badges can be imported using an id made up of `{project_id}:{badge_id}`,
+ e.g.
+
+```bash
+terraform import gitlab_project_badge.foo 1:3
+```
--- a/gitlab/resource_gitlab_project_badge.go
+++ b/gitlab/resource_gitlab_project_badge.go
@@ -26,7 +26,7 @@ func resourceGitlabProjectBadge() *schema.Resource {
 				Required: true,
 			},
 			"image_url": {
-				Type:      schema.TypeString,
+				Type:     schema.TypeString,
 				Required: true,
 			},
 			"rendered_link_url": {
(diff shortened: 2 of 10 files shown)


fandujar:
@ash2k  Could you review this PR?

ash2k:
@fandujar Although I do work for GitLab, I'm not familiar enough with this project or GitLab API to properly review this, sorry. I've reviewed some other PRs here, but I was involved in preparing them, so had the context. I can have a look at the Go code itself only.

fandujar:
Thanks for letting me know @ash2k  Sorry for bother you. @nicholasklick maybe you could guide me?

gitlab/resource_gitlab_project_badge.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	log.Printf("[DEBUG] create gitlab project badge %q / %q", *options.LinkURL, *options.ImageURL)
+
+	badge, _, err := client.ProjectBadges.AddProjectBadge(project, options)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(fmt.Sprintf("%d", badge.ID))
```
armsnyder, line 59: The upstream API for badges requires a project ID as well as the badge ID. https://docs.gitlab.com/ee/api/project_badges.html#get-a-badge-of-a-project I think you need both of these IDs in your resource ID in order to avoid colliding IDs between projects.

armsnyder (changes requested):

gitlab/resource_gitlab_project_badge.go
fandujar: @armsnyder You are right. I will change it

fandujar: @armsnyder The resource will receive the project ID and every step will use it, so it won't collide.

fandujar: The same behavior of project hook: https://github.com/gitlabhq/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_project_hook.go#L133

armsnyder: I'll explain the reason below, but my ask is:

Could you change the `gitlab_project_badge` resource's ID format to be `projectid:badgeid`, then add an importer, then modify the `resourceGitlabProjectBadgeRead` function to only reference the resource ID instead of by reading the `project` attribute? See `gitlab_project_mirror` for reference:

https://github.com/gitlabhq/terraform-provider-gitlab/blob/8e2faa78634fe687a5d491e3370ca9fd206f2476/gitlab/resource_gitlab_project_mirror.go#L18-L20

https://github.com/gitlabhq/terraform-provider-gitlab/blob/8e2faa78634fe687a5d491e3370ca9fd206f2476/gitlab/resource_gitlab_project_mirror.go#L132-L137

### Explanation

Is there some GitLab documentation that says badge IDs are globally unique? Here's the scenario I want to avoid:
1. TF user creates `gitlab_project_badge.foo` for project `foo`. It gets an ID `1`.
2. TF user creates `gitlab_project_badge.bar` for project `bar`. It also gets an ID `1`.

Then there would be two of the same TF resource in the TF state with the same ID, which is not allowed by TF.

There is another reason for including the project ID in the resource ID, and that is it makes adding an importer a lot easier. **Ideally you want all of the IDs required to GET the upstream resource to be included in the TF resource ID**, so you can avoid needing to write a custom importer function, and use the `ImportStatePassthrough` helper instead: https://www.terraform.io/docs/extend/resources/import.html#importer-state-function

(We want all resources in this provider to have an importer.)

fandujar: @armsnyder Didn't think about how importer would work and it is a good enough reason. I will add the change tomorrow. Thank you for explaining

fandujar: @armsnyder Done. Could you review it again?

gitlab/resource_gitlab_project_badge_test.go
```diff
@@ -0,0 +1,178 @@
...
+	rInt := acctest.RandInt()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckGitlabProjectBadgeDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGitlabProjectBadgeConfig(rInt),
+			},
```
armsnyder, line 60: These tests can be combined. See the first bullet here: https://www.terraform.io/docs/extend/resources/import.html#recommendations-for-import

```suggestion
```

armsnyder (changes requested):
Can you add a documentation page under /docs please?

gitlab/resource_gitlab_project_badge_test.go
fandujar: @armsnyder I added the docs and combined tests in one single function. If you find something else, please let me know

docs/resources/project_badge.md
```diff
@@ -0,0 +1,39 @@
+# gitlab\_project\_badge
```
armsnyder, line 1: Can you remove the consecutive newlines (line 7 and 32), and add a terminating newline character to this file?

We very recently added the [markdown-linter](https://github.com/avto-dev/markdown-lint) to our CI, and it will complain about these after we merge to master.

armsnyder (changes requested):

docs/resources/project_badge.md
fandujar: @armsnyder Done. I tested it with the same tool and it's ok now.

armsnyder: Thanks for checking

docs/resources/project_badge.md
```diff
@@ -0,0 +1,37 @@
...
+
+* `image_url` - (Required) The image url which will be presented on project overview.
+
+## Attributes Reference
+
+The resource exports the following attributes:
+
+* `rendered_link_url` - The link_url argument rendered (in case of use of placeholders).
+
+* `rendered_image_url` - The image_url argument rendered (in case of use of placeholders).
```
armsnyder, line 37: Can you document how to import please?

See https://www.terraform.io/docs/extend/resources/import.html#resource-documentation-implementation

armsnyder (changes requested):

docs/resources/project_badge.md
fandujar: Done

armsnyder (approved):
Thanks for the new resource! 🎉

fandujar:
Thank you @armsnyder for all review.

PR #649, opened 2021-07-14
chore: add error handling to resourceGitlabProjectSetToState

mattkasa:
This adds error handling to the `resourceGitlabProjectSetToState` function which used to call `Set` repeatedly without checking for errors.


ash2k (approved):
👍

PR #650, opened 2021-07-15
chore: remove incomplete gitlab_project_repository_file resource

mattkasa:
This file was part of #556 but did not include a test, docs, or an entry in gitlab/provider.go, so we are removing it until it can be completed.


armsnyder (approved):
lgtm

PR #651, opened 2021-07-15
chore: add error checking and fix other linter errors

mattkasa:
This adds error checking to `d.Set` calls, as well as fixes other linter errors.


gitlab/resource_gitlab_project_test.go
```diff
@@ -502,7 +502,12 @@ func TestAccGitlabProject_importURL(t *testing.T) {
 		t.Fatalf("failed to create base project: %v", err)
 	}
 
-	defer client.Projects.DeleteProject(baseProject.ID)
+	defer func() {
+		_, derr := client.Projects.DeleteProject(baseProject.ID)
+		if err == nil {
+			err = derr
+		}
```
ash2k, line 509:
```suggestion
		if derr != nil {
			t.Fatalf("failed to delete base project: %v", derr)
		}
```

gitlab/resource_gitlab_project_test.go
```diff
@@ -586,7 +591,12 @@ func TestAccGitlabProject_importURLMirrored(t *testing.T) {
...
```
ash2k, line 598:
```suggestion
		if derr != nil {
			t.Fatalf("failed to delete base project: %v", derr)
		}
```

ash2k (changes requested):
Everything is good apart from those defers :)

armsnyder:
Thanks for the contribution! I would prefer that we do not accept this change, because it is not necessary. The Terraform provider documentation states that using `d.Set` with primitive types do not require error checking. See: https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types

ash2k:
@armsnyder Currently the acceptance tests are failing. My understanding is that the error checking added here have actually caught real bugs - either in tests or in production code (clearly both are bad). Even if error checking is unnecessary when the API is used correctly, these failures clearly show that this change is valuable as it helps to catch cases where API is misused.

p.s. I wish Terraform SDK provided a separate API for primitive types. If misused, it should panic. It's just poor API design to ask library user to ignore errors in some cases.

armsnyder:
@ash2k My comment was really only about `d.Set`. The other changes are great, and I really love the addition of a linter!

I agree it is poor SDK design but would still defer to the documented best practices here, to make our code more accessible to contributors from other providers. There is no value added in error checking these statements, and because error checking is verbose in Golang it only complicates the code. Luckily linting missing error checks from non-primitive values is supported by tfproviderlint, which I've opened a PR to add: #653

ash2k:
@armsnyder

> There is no value added in error checking these statements

I'm sorry, but I do not agree on this one :) Returned error is a 100% correct signal, linters usually have false positives/negatives. Just checking the error is a simple, reliable and self-sufficient solution to the problem 🤷 I agree the linter should be added too to catch issues without actually executing the code (i.e. earlier in the dev cycle and/or where there is no test coverage), but in my opinion it's an additive measure rather than a replacement for error checking.

Having said the above, I'll defer the decision to you as you are a long time maintainer of the project and I have no plans to interfere or take over. Thanks for all the work you have put in into this project!

armsnyder:
@ash2k I think I understand where you're coming from; checking all errors would be a foolproof way of eliminating issues that stem from missing error checks. 😄 I can see how my assertion that there is "no value" was a bit inflammatory/hyperbolic. Sorry about that!

However I am also trying to consider code readability and quality of life for contributors. These factors matter as well.

The acceptance test framework that Terraform provides us verifies `d.Set` issues with primitives for us automatically, since they are caught at runtime. In fact, I think it will do this regardless of whether a test is specifically written for the new resource.

armsnyder:
I also recognize and appreciate the amount of work put into this PR, so I don't say any of this lightly.

ash2k:
> However I am also trying to consider code readability and quality of life for contributors. These factors matter as well.

I was just discussing this with someone yesterday. In my opinion correctness (and hence security) is number one priority. Then comes developer productivity and then everything else, including software performance.

> The acceptance test framework that Terraform provides us verifies d.Set issues with primitives for us automatically, since they are caught at runtime. In fact, I think it will do this regardless of whether a test is specifically written for the new resource.

If it does that then how is it possible that the error handling have caught something? It's totally possible I'm misunderstanding something...

mattkasa:
@armsnyder I think if it was the case that the test framework Terraform provides would verify `d.Set` primitives that I would not be finding so many real bugs via adding these error checks, for instance, this one I am fixing right now:

```
=== RUN   TestAccGitlabServiceJira_basic
    testing.go:705: Step 0 error: errors during apply:
        
        Error: Invalid address to set: []string{"tag_push_events"}
        
          on /var/folders/nv/n1lf4tdx0fs9hkl3r88vt0_c0000gn/T/tf-test470943420/main.tf line 10:
          (source code not available)
        
        
    testing.go:766: Error destroying resource! WARNING: Dangling resources
        may exist. The full state and error is shown below.
        
        Error: errors during refresh: Invalid address to set: []string{"note_events"}
        
        State: <nil>
--- FAIL: TestAccGitlabServiceJira_basic (3.17s)
```

This is because in `resourceGitlabServiceJiraRead` we are trying to set `tag_push_events`, `note_events`, `pipeline_events`, and `job_events`. None of which exist in `go-gitlab` (see https://github.com/xanzy/go-gitlab/blob/4eecb48f2fe28ec0f5a22b2c4828d4252654c5aa/services.go#L728).

This is just one of dozens of such previously uncaught issues, that are being revealed by adding these error checks. So I would suggest that we may need to find a balance of readability and code correctness that allows us to catch and fix bugs like this. WDYT?

armsnyder:
> In my opinion correctness (and hence security) is number one priority. Then comes developer productivity and then everything else, including software performance.

Hehe we're getting a bit philosophical. I guess I would disagree slightly, as I don't think you can hold one value above others unconditionally, since the factors are almost always different from project to project. I also believe readability is a factor in security. In the case of Terraform providers I'd align toward best practices since they are well documented and (in my opinion) very reasonable (with the gripe that there were some strange API design choices). In the case of an open source like this one where there has not historically been a significant amount of participation, I'd value quality of life higher than if this were an internal project.

However I know that GitLab has a lot of openly documented internal best practices, so feel free to cite them, as I am not myself very familiar.

> If it does that then how is it possible that the error handling have caught something?

>  I think if it was the case that the test framework Terraform provides would verify d.Set primitives that I would not be finding so many real bugs via adding these error checks

It only catches misuse of primitives. The bugs in this provider were all having to do with complex types, as far as I can tell. These should all be caught by [XR004](https://github.com/bflad/tfproviderlint/tree/main/xpasses/XR004).

@mattkasa In fairness, the `service_jira` bugs were non impacting to the usability of the resource, right?

It's getting late where I am, so I won't be able to continue the discussion tonight. Sorry to bail! I suggest deferring to patterns in the HashiCorp best practices other popular TF providers. Best practices aren't automatically right, but taken as a whole, I still think the right balance would be to enforce correctness using passive linters where possible until we have reason to believe it is not sufficient.

mattkasa:
It's quite late here for me as well, but I am pushing up what I have so you can both see what issues have been revealed, and my best guesses at fixes for them.

The remaining 4 CE acceptance test failures are related to the error checking on line 479 of `resource_gitlab_project.go` and appear to be non-trivial, it looks like we are ignoring a 400 from gitlab saying the transfer of a project to a new namespace has failed.

But it's late, so bears further investigation tomorrow 😄  It feels like this has been useful, at least in terms of discovering that a few resources are likely not working the way we thought they were. Looking forward to working on making the provider better with you both!

mattkasa:
I'm going to go ahead and mark this PR as a draft, since merging #652 and #653 will cause conflicts in this branch, and it will have to be redone anyway.

I am thinking it is probably better to merge #652 and #653 first, then go through systematically and resolve each of these issues one by one in smaller, easier to review PRs, and then remove any remaining linter exclusions and fix any remaining issues at that point in a followup later on.

What do you both think about that plan, @armsnyder and @ash2k?

armsnyder:
@mattkasa Sounds good to me. Once all of the golangci-lint and markdown-lint linter issues are resolved, we should remember to enable the linters on pull requests. (Right now they're only running on pushes to master.) I still owe you a review on the bugs fixed here to better understand whether/how we can/should enforce error handling, and breaking it into multiple PRs would help a ton!

armsnyder:
@mattkasa After we merge your bug fixes, we should start running the test suite with `TF_SCHEMA_PANIC_ON_ERROR=1` (the default setting in the v2 SDK).

```
TF_SCHEMA_PANIC_ON_ERROR=1 make testacc
<snip>
=== RUN   TestAccDataSourceGitlabUser_basic
panic: created_at: '' expected type 'string', got unconvertible type 'time.Time'

goroutine 8807 [running]:
github.com/hashicorp/terraform-plugin-sdk/helper/schema.(*ResourceData).Set(0xc0002cc310, 0x207a5ac, 0xa, 0x2062cc0, 0xc001196810, 0x0, 0x0)
	/Users/asnyder/go/pkg/mod/github.com/hashicorp/terraform-plugin-sdk@v1.16.1/helper/schema/resource_data.go:198 +0x2e6
<snip>
FAIL	github.com/gitlabhq/terraform-provider-gitlab/gitlab	69.294s
FAIL
make: *** [testacc] Error 1

timofurrer:
I've separately pushed the lint script with #793 for now. I think this PR needs quite some rebasing - especially with the completed upgrade to SDK v2 the situation is quite different. I also suggest to tackle a lot of the things here individually. 

I'm taking the liberty to close this for now. (@armsnyder feel free to re-open if you disagree with my judgement :))

PR #652, opened 2021-07-15
golangci-lint: exclude ResourceData.Set

armsnyder:
The common case when using this method is to ignore the error return. This is because the Terraform test framework checks these errors for primitive types for us.

See https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,3 @@
+linters-settings:
+  errcheck:
+    exclude: errcheck_excludes.txt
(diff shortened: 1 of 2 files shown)


mattkasa (approved):
LGTM, I think we should merge this, and then #653, and after resolving the issues you marked with `TODO` in that PR, we can then later revisit `errcheck` and resolve anything still outstanding at that time 👍

PR #657, opened 2021-07-15
CI: remove extraneous errcheck script; gofmt with golangci-lint

armsnyder:
* golangci-lint already runs errcheck, so I removed our custom script for that.
* I enabled gofmt in the golangci-lint settings so it will check formatting, to consolidate our linters some more and remove another custom script.
* golangci-lint already runs `go vet`, so I removed that step from the CI workflow. I set it to enabled in the config to be explicit.


mattkasa (approved):
LGTM 👍

PR #653, opened 2021-07-15
Add tfproviderlint linter to CI

armsnyder:
Adds the [tfproviderlint](https://github.com/bflad/tfproviderlint) linter to the CI workflow. This linter is used by the [aws provider](https://github.com/hashicorp/terraform-provider-aws/tree/main/awsproviderlint) and others to catch common mistakes when writing Terraform providers, and will help automate code reviews. Reported issues conveniently show up as code annotations in pull requests.

I littered the code with `// lintignore` comments instead of fixing all of the preexisting issues, since there were so many. The intent is for newly submitted code to adhere to the linter checks, and we can resolve these past issues over time.

The full list of checks is [documented here](https://github.com/bflad/tfproviderlint). I used the extended ruleset, but with some exceptions:
* [XAT001](https://github.com/bflad/tfproviderlint/tree/main/xpasses/XAT001) - Our SDK version does not support this yet
* [XR003](https://github.com/bflad/tfproviderlint/tree/main/xpasses/XR003) - I was unsure whether custom timeouts make sense for our provider
* [XR005](https://github.com/bflad/tfproviderlint/tree/main/xpasses/XR005) - This can wait until we have auto-generated documentation set up
* [XS001](https://github.com/bflad/tfproviderlint/tree/main/xpasses/XS001) - This can wait until we have auto-generated documentation set up
* [XS002](https://github.com/bflad/tfproviderlint/tree/main/xpasses/XS002) - Personally I find it more valuable to let contributors order attributes in order of importance

PR #658, opened 2021-07-15
Annotate golangci-lint failing checks

mattkasa:
chore: run golangci-lint for pull requests and annotate failing checks so we can fix them individually, similarly to the tfproviderlint checks annotated in #653


armsnyder (approved):
Thanks!

PR #662, opened 2021-07-18
Add CustomCIPath to resource `gitlab_project`

kitos9112:
This Merge Request aims to add a very simple change to the GitLab resource Project resource so that the CustomCIPath can be passed as a parameter to the Terraform resource

 - Amend documentation to reflect the change
 - Add a standalone acceptance test

### DIFF
--- a/gitlab/resource_gitlab_project_test.go
+++ b/gitlab/resource_gitlab_project_test.go
@@ -48,6 +48,7 @@ func TestAccGitlabProject_basic(t *testing.T) {
 		PackagesEnabled:    true,
 		PagesAccessLevel:   gitlab.PublicAccessControl,
 		BuildCoverageRegex: "foo",
+		CIConfigPath:       ".gitlab-ci.yml@mynamespace/myproject",
 	}
 
 	defaultsMainBranch = defaults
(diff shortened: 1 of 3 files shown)


gitlab/resource_gitlab_project_test.go
```diff
@@ -581,6 +581,36 @@ resource "gitlab_project" "foo" {
 	})
 }
 
+func TestAccGitlabProject_initializeWithCustomCIPath(t *testing.T) {
```
armsnyder, line 584: Usually we try to only create new test functions if they require some unique setup. It looks like this could be added by adding the new attribute in the first two TestSteps in the existing TestAccGitlabProject_basic test. Other than that nitpick it looks good!

armsnyder (changes requested):

gitlab/resource_gitlab_project_test.go
kitos9112: Yay, my initial uncommitted changes were made to the first test function based on that simple assumption, however I ended up creating a new test case function for it as I have never contributed to the project before. I'll take a look tomorrow morning and correct as per your comments

kitos9112: @armsnyder I've made the appropriate changes and ran the Acc. tests locally. Seems Ok

armsnyder: Thanks, I appreciate it.

armsnyder (approved):
Looks great!

PR #673, opened 2021-07-24
add support for gitlab group badges

zbindenren:
This adds support for group badges.

### DIFF
--- a/gitlab/resource_gitlab_group_badge_test.go
+++ b/gitlab/resource_gitlab_group_badge_test.go
@@ -1,0 +1,167 @@
+package gitlab
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+	"testing"
+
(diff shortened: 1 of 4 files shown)


gitlab/resource_gitlab_group_badge_test.go
```diff
@@ -0,0 +1,167 @@
...
+  description = "Terraform acceptance tests"
+
+  # So that acceptance tests can be run in a gitlab organization
+  # with no billing
+  visibility_level = "public"
+}
+
+resource "gitlab_group_badge" "foo" {
+  group     = "${gitlab_group.foo.id}"
+  link_url  = "http://example.com/badge-%d"
```
mkjmdski, line 142: I am not sure if some point http:// connections will not be disabled to prevent mixed content I propose that you set `https://` here and make `testAccGitlabGroupBadgeUpdateConfig` function change the site url instead of protocol

zbindenren:
Thx

PR #677, opened 2021-07-26
Clarify supported Terraform versions

nicholasklick:


mattkasa (approved):
LGTM, I think this makes sense to clarify what we currently say 👍

Issue #60, opened 2018-07-19
Support OAuth Tokens


Lasering:
The [official terraform documentation](https://www.terraform.io/docs/providers/gitlab/index.html) for the Gitlab provider says that `token` should be a personal access token. Which is correct since [config.go](https://github.com/terraform-providers/terraform-provider-gitlab/blob/master/gitlab/config.go#L49) invokes `client := gitlab.NewClient(httpClient, c.Token)` and [that](https://github.com/xanzy/go-gitlab/blob/master/gitlab.go#L340) is setting the `authType` to be `privateToken`.

However, [provider.go](https://github.com/terraform-providers/terraform-provider-gitlab/blob/master/gitlab/provider.go#L67) has for the `token` description `The OAuth token used to connect to GitLab.` which hints that somewhere is the past `token` could be an OAuth token. Since `go-gitlab` supports OAuth tokens with the [`NewOAuthClient` function](https://github.com/xanzy/go-gitlab/blob/master/gitlab.go#L384) this provider could also support it. I suggest adding a new argument called `oauth_token` which would be optional, and if it was set then the `NewOAuthClient` would be used.

Lasering:
I can make a pull request if that would make it to production faster.

armsnyder:
What would you expect the DELETE operation to do? Should it create a [revert](https://docs.gitlab.com/ee/api/commits.html#revert-a-commit)?

armsnyder:
The documentation should be fixed, but I'm curious, what is the use case for using an OAuth token in Terraform, where a personal access token isn't preferred?

Lasering:
I wanted to have a Gitlab system OAuth application that invoked terraform recipes. In a OAuth application you get an access token which is not a personal access token. Meanwhile I evolved my setup and no longer use a terraform recipe. But seems other people might have a similar use case.

wuperry563:
Hey there @armsnyder , I've actually ran into this problem within my organization as well. I'd like to take some time to contribute to this project. 

My use case is I have a UI App written for Gitlab users to generate a OAuth token, and have users create repos and files they have access rights to. 

Just wanted to make sure no one else is working on this, and that I can pick this up as an assignee.

seboudry:
Hi there!
Is there a plan to make a release with this feature?

PR #678, opened 2021-07-26
add oauth token support

wuperry563:
Related to: https://github.com/gitlabhq/terraform-provider-gitlab/issues/60


wuperry563:
@armsnyder Ready for review. This is my first contribution, so please feel free to let me know if somethings missing. 

Thanks!

roidelapluie:
Thanks! I think it should be a dedicated option, we should not try to guess which kind of token we are dealing with.

brandong1:
@nicholasklick Tagging you as requested.

nicholasklick:
@wuperry563 please note that @roidelapluie is a maintainer for this project, so please kindly follow any guidance or feedback in order to move this towards merge. Thanks!

wuperry563:
> Thanks! I think it should be a dedicated option, we should not try to guess which kind of token we are dealing with.

Sounds good. I'll look into this!

armsnyder:
@roidelapluie I was thinking the same, but noticed something in the docs.

The [API docs](https://docs.gitlab.com/ee/api/#personalproject-access-tokens) say you can use a PAT the same as an OAuth token.

> You can also use personal or project access tokens with OAuth-compliant headers.

I propose changing the constructor from `gitlab.NewClient()` to `gitlab.NewOAuthClient()` for both PATs and OAuth tokens.

I tried it in a branch here and tests are passing: https://github.com/gitlabhq/terraform-provider-gitlab/compare/master...oauth-token

The only catch is that the the validation is slightly more strict, so I increased the length of our ACCTEST token. (Real PATs are the correct length, and I've verified that they work.)

wuperry563:
@armsnyder It sounds like you already got it working and can submit a PR. Is there anything that I can help with as a first-time contributor? 

Should I close my PR?

roidelapluie:
Oooh so it seems then that we are not "cheating" by using NewOAuthClient instead of NewClient. I guess we can transparently make the change in the code and slightly adapt our documentation.

armsnyder:
Great! That being the case I've opened a new PR with my diff #686.

Thanks much to @wuperry563 for the initial work and getting us all looking into this!

PR #686, opened 2021-08-03
Support running the provider using an OAuth token

### LINKED ISSUE #60: Support OAuth Tokens

armsnyder:
Supersedes #678
Closes #60 

Thanks to @wuperry563 for the initial PR which this builds on.

As discussed in that PR, GitLab supports the same auth header for both PAT and OAuth. I've tested this change using a real PAT and it still works.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -1,6 +1,6 @@
 TEST?=./gitlab
 SERVICE?=gitlab-ce
-GITLAB_TOKEN?=ACCTEST
+GITLAB_TOKEN?=ACCTEST1234567890123
 GITLAB_BASE_URL?=http://127.0.0.1:8080/api/v4
 GOFMT_FILES?=$$(find . -name '*.go' |grep -v vendor)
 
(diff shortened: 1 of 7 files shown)


wuperry563:
Thanks for looking into the issue! My org has a real need for this!

roidelapluie (approved):

roidelapluie:
Thanks!

Issue #667, opened 2021-07-20
Feature Request: Change the default branch protection of a group


mick1627:
### Terraform Version
1

### Affected Resource(s)
gitlab_group

### Expected Behavior
Be able to set the default branch protection at group level as describe here  https://docs.gitlab.com/ee/user/group/index.html#change-the-default-branch-protection-of-a-group

### Actual Behavior
not possible actually
By default, default branch has branch protection, and then it not possible to use gitlab_branch_protection. Terraform will raise the following error => Error: protected branch "master" on project "1881" already exists
If I can disable default branch protection at group level, I will be able to manage branch protection using gitlab_branch_protection.

armsnyder:
API reference: https://docs.gitlab.com/ee/api/groups.html#new-group

The ask is to add a `default_branch_protection` attribute to the `gitlab_group` resource.

cmd-werner:
Hi everyone, in addition to this request I'd like to add an issue that is raised with the default configuration.

By default, gitlab will set the branch protection to be fully protected at the group level. If you (like me) manage the branch protection at the project level, then for new projects the execution will fail with an error similar to:

```
│ Error: protected branch "master" on project "12345678" already exists: {ID:987654321 Name:master PushAccessLevels:[0xc0000adb00] MergeAccessLevels:[0xc0000adb30] UnprotectAccessLevels:[] AllowForcePush:false CodeOwnerApprovalRequired:false}
│ 
│   on resource "gitlab_branch_protection" "main" {
│ 
```

The ability to have this group level feature will allow us to avoid manual configuration on groups.


Thanks

mgriffin13:
The two issues are definitely related, but I think they should both be solved. Adding the `default branch protection` setting for a `gitlab_group` resource, but they should also allow `gitlab_branch_protection` to manage an already existing branch protection object in Gitlab.

switchboardOp:
https://github.com/gitlabhq/terraform-provider-gitlab/pull/706 implements this and the [upstream go library](https://github.com/xanzy/go-gitlab) has released a version (`0.50.4`) that supports it.

callum-stakater:
[...]

PR #706, opened 2021-08-26
support default_branch_protection on group

domcyrus:
Support using default_branch_protection on group. This is implementing one part of #667 

It will need a newer dependency of the "github.com/xanzy/go-gitlab" package. Currently only using github.com/xanzy/go-gitlab@v0.50.4-0.20210826102709-de0117b95832 (master) as the change of the needed default_branch_protection on group is only in master.


domcyrus:
I've updated the PR to use go-gitlab 0.50.4.

domcyrus:
Updated PR all tests run successfully.

switchboardOp:
I'd love to see this merged now that the upstream dependency is released. Default branch protection is causing some issues for me. I was halfway through my own implementation when I found this PR. looks good @domcyrus!

willianpaixao:
@domcyrus would you mind updating the documentation of the underlying resources as well?

domcyrus:
@willianpaixao I added the missing documentation and also added unit tests.

domcyrus:
Is there anything else needed?

willianpaixao:
@domcyrus thanks! LGTM.

switchboardOp:
fixes #667

mkjmdski:
Hi @domcyrus. The newest go-gitlab is already in master you should take it's go mod files and this branch should be good to go.

domcyrus:
@mkjmdski updated with upstream master.

mkjmdski:
@roidelapluie could you please enable pipeline in here?

roidelapluie:
Thanks!

QuingKhaos:
As a note: it could have supported string values like `visibility_level` which are translated internally to the int access levels to be consistent..

roidelapluie:
```
terraform-provider-gitlab $ ack PrivateVisibility
gitlab/util_test.go
46:                     Level:  gitlab.PrivateVisibility,

gitlab/util.go
80:             "private":  gitlab.PrivateVisibility,

vendor/github.com/xanzy/go-gitlab/types.go
488:    PrivateVisibility  VisibilityValue = "private"

```

roidelapluie:
I think I now understand what you mean. In terraform, the best practice is to stick to the remote API as close as possible. If the gitlab API asks for an integer, I think we should not invent our own values.

QuingKhaos:
> In terraform, the best practice is to stick to the remote API as close as possible.
I may agree.

> If the gitlab API asks for an integer, I think we should not invent our own values.

It's already done with the access control level values (`maintainer` instead of `40`) - sorry, confused them with the visibility level. Maybe not directly in the provider, but through the Go SDK. In this case I would prefer consistency with the other values and have

Issue #292, opened 2020-04-02
[Feature Request] Support for option "Squash commits when merge request is accepted." for projects


MarcinNowak-codes:
Currently field for option "Squash commits when merge request is accepted." doesn't exist in GitLab's API for project [Get single project](https://docs.gitlab.com/ee/api/projects.html#get-single-project). I create this issue to cross track GitLab's API progress.

### References
[Set default value for merge request options](https://gitlab.com/gitlab-org/gitlab-foss/-/issues/27956)
[Add project setting to force usage of squash merge](https://gitlab.com/gitlab-org/gitlab/-/issues/2012)

MarcinNowak-codes:
Another issue in Gitlab to cross track [Configurable defaults for "Squash commits" option](https://gitlab.com/gitlab-org/gitlab/-/issues/17613) since Mar 26, 2017. This time it is open.

MarcinNowak-codes:
I've created issue in go-gitlab [Add support for option "Squash commits when merge request is accepted." for projects](https://github.com/xanzy/go-gitlab/issues/838).  Feature is scheduled for GitLab 13.1 and MR is in WIP status.

armsnyder:
This is the issue in GitLab to add the API for this. Currently there is no milestone. https://gitlab.com/gitlab-org/gitlab/-/issues/238531

dglynn:
Hi we are looking to maintain this option via terraform for all our projects. I see the corresponding API ticket is open in your Gitlab issue tracker: https://gitlab.com/gitlab-org/gitlab/-/issues/261458

Can we get some updates as to when this is going to land in the API options itself and how soon after will it be in the provider now you maintain this in house. Thanks

aston-r:
API is available on GitLab side. Issue is closed: https://gitlab.com/gitlab-org/gitlab/-/issues/261458

tringuyen-yw:
Can you please make sure that the merge option "Squash commits when merging" has the 4 possible values as in Gitlab Settings/General:

- Do not allow
- Allow
- Encourage
- Require

ahaynssen:
Changes on the GitLab API are now merged and available via the docs. https://gitlab.com/gitlab-org/gitlab/-/issues/261458#note_423281500 

Paging @sirlatrom :) I saw you in the comments mention you might have code ready to update for this! 

Thanks!

sirlatrom:
[...]

PR #718, opened 2021-09-23
Bump github.com/xanzy/go-gitlab from 0.50.0 to 0.51.1

mkjmdski:
Fixes to pull request from dependabot at https://github.com/gitlabhq/terraform-provider-gitlab/pull/713/files

PR #702, opened 2021-08-21
feat: add squash_option to project resource

ckoliber:
In this feature `squash_option` added to project resource

PR #719, opened 2021-09-23
Add squash option with data source

mkjmdski:
Thanks for the initial pull request to @ckoliber (https://github.com/gitlabhq/terraform-provider-gitlab/pull/702)! Closes: https://github.com/gitlabhq/terraform-provider-gitlab/issues/292 also depends on https://github.com/gitlabhq/terraform-provider-gitlab/pull/718

### DIFF
--- a/gitlab/resource_gitlab_project.go
+++ b/gitlab/resource_gitlab_project.go
@@ -158,9 +158,9 @@ var resourceGitLabProjectSchema = map[string]*schema.Schema{
 		Optional: true,
 	},
 	"squash_option": {
-		Type:          schema.TypeString,
-		Optional:      true,
-		Default:  	   "default_off",
+		Type:     schema.TypeString,
+		Optional: true,
+		Default:  "default_off",
 	},
 	"remove_source_branch_after_merge": {
 		Type:     schema.TypeBool,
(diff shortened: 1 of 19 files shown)


mkjmdski:
@roidelapluie ready to merge :)

gitlab/resource_gitlab_project.go
```diff
@@ -162,6 +162,11 @@ var resourceGitLabProjectSchema = map[string]*schema.Schema{
 		Type:     schema.TypeBool,
 		Optional: true,
 	},
+	"squash_option": {
+		Type:     schema.TypeString,
```
roidelapluie, line 166: Plase use validation.StringInSlice.

gitlab/util.go
```diff
@@ -143,6 +143,21 @@ func stringToMergeMethod(s string) *gitlab.MergeMethodValue {
...
 
+func stringToSquashOptionValue(s string) *gitlab.SquashOptionValue {
+	lookup := map[string]gitlab.SquashOptionValue{
+		"never":       gitlab.SquashOptionNever,
+		"always":      gitlab.SquashOptionAlways,
+		"default_on":  gitlab.SquashOptionDefaultOn,
+		"default_off": gitlab.SquashOptionDefaultOff,
+	}
+
+	value, ok := lookup[s]
```
roidelapluie, line 154: We should error here.

gitlab/resource_gitlab_project.go
roidelapluie: Mmh validation with stringToSquashOptionValue would be even better.

gitlab/util.go
mkjmdski: there is error handling on line 155, or do you mean something else?

gitlab/resource_gitlab_project.go
mkjmdski: It's kind of hard to pass the type of stringToSquashOptionValue function into ValidateFunc :/. So I stayed with validation.StringInSlice like in other examples.

gitlab/util.go
mkjmdski: But this causes places like this:

```golang
		SquashOption:                              stringToSquashOptionValue(d.Get("squash_option").(string)),
```

to have issues with number of returned agruments, it exactly expects one of them, I am not sure how should I handle that.

mkjmdski: I'd be happy to follow some example on how to implement stringMapping function as validation but it seems like that this one does not differ from the others :)

Issue #671, opened 2021-07-21
gitlab_branch_protection - Doesn't work if default branch protection matches the desired protection level


kemra102:
### Terraform Version
```sh
$ terraform version
Terraform v1.0.1
on linux_amd64
```

### Affected Resource(s)
- `gitlab_branch_protection`

### Terraform Configuration Files
```hcl
resource "gitlab_project" "project" {
  name               = var.name
  path                 = var.path
  description       = var.description
  namespace_id = var.group_id
  default_branch = "main"
  ...
}

resource "gitlab_branch_protection" "main" {
  branch                                        = "main"
  merge_access_level                  = "maintainer"
  project                                        = gitlab_project.project.id
  push_access_level                     = "no one"
  code_owner_approval_required = true
}
```

### Expected Behavior
With the project default set such that the default branch protection is enabled we should be able to create the repo and set the additional options present in the `gitlab_branch_protection` resource such as `code_owner_approval_required`.

### Actual Behavior
An error stating that the branch protection already exists:
```sh
Error: protected branch "main" on project "1234" already exists: {ID: 1234 Name:main PushAccesslevels;[0x0002828a0] MergeAccessLevels:[0xc0002828d0] UnprotectedAccessLevels[] AllowForcePush:false CodeOwnerApprovalRequired:false}
```
For now we are working around this by importing the `gitlab_branch_protection` then running a `terraofrm` apply again which re-creates (deletes/creates) the resource correctly.

### Important Factoids
The issue (this is an educated guess at best) seems to come down to one important fact, if the project default "default branch" is set to protected and that is in fact the setting we want then we are locked out of configuring other settings on that branch. Setting the project default to NOT have a protected default branch is NOT an option for us as we are subject to very strict security requirements.

### References
- GH-667

mgriffin13:
I am experiencing the same behavior. We are using Terraform Enterprise, so importing the resource becomes much more challenging for us.

Would it be possible for TF to just delete/create the gitlab_branch_protection resource initially if it detects it already exists?

mgriffin13:
[...]

PR #721, opened 2021-09-23
No need to protect branch at creation, there is a resource to handle that

mgsegner:
Hello,

I encountered issues with project creation and default branch protection. Let me detail my case.

**Issue**
When a repository is created the default branch is protected "implicilty" by the provider, but we can not manage the protection with resource gitlab_branch_protection because it ends with an error saying the branch is already protected.
So we are unable to define resource gitlab_branch_protection on the default branch to ensure it will stay protected over time.
I guess the current behavior aims to copy the GUI one, but it could be annoying with the terraform usage.

**Commit**
This commit will remove the "implicit" branch protection in the provider

**New behavior**
At creation time default branch will not be protected
Ensure the default branch is protected in terraform definition by using gitlab_branch_protection resource

This commit will change current behavior of project creation and an explicit definition of gitlab_branch_protection will be necessary after that.
Do you think this commit and the new behavior are right ?

Regards,

### DIFF
--- a/docs/resources/project.md
+++ b/docs/resources/project.md
@@ -41,6 +41,7 @@ The following arguments are supported:
 * `tags` - (Optional) Tags (topics) of the project.
 
 * `default_branch` - (Optional) The default branch for the project.
+Default branch is not protected at project creation, to do so manage default branch with [`gitlab_branch_protection`](branch_protection.html) resource.
 
 * `import_url` - (Optional) Git URL to a repository to be imported.
 
--- a/gitlab/resource_gitlab_project.go
+++ b/gitlab/resource_gitlab_project.go
@@ -501,14 +501,6 @@ func resourceGitlabProjectCreate(d *schema.ResourceData, meta interface{}) error
 			return fmt.Errorf("Failed to set default branch to %q for project %q: %w", newDefaultBranch, d.Id(), err)
 		}
 
-		log.Printf("[DEBUG] protect new default branch %q for project %q", newDefaultBranch, d.Id())
-		_, _, err = client.ProtectedBranches.ProtectRepositoryBranches(project.ID, &gitlab.ProtectRepositoryBranchesOptions{
-			Name: gitlab.String(newDefaultBranch),
-		})
-		if err != nil {
-			return fmt.Errorf("Failed to protect default branch %q for project %q: %w", newDefaultBranch, d.Id(), err)
-		}
-
 		log.Printf("[DEBUG] unprotect old default branch %q for project %q", oldDefaultBranch, d.Id())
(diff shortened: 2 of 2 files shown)


mkjmdski:
This is great pull request! I had the issue with it that I couldn't change the default branch protection rules without importing it after creation what required manual step during repository provisioning. Although this behaviour will differ from default behaviour when user creates a repository from the UI so I'd suggest that you write explicitly in the documentation of the `gitlab_project` that default branch created by the resource will not be protected.

As work around before merging this pull request I suggest to anybody that encountred same issue as I did that you do not define default_branch in the repo and just set branch protection rule on not existing branch and then default branch will set automatically within the first commit to the repo.

roidelapluie:
This is a breaking change that would need a major release of the provider.

mgsegner:
Thanks for your feedback, I will have a look to documentation and update it as suggested

mgsegner:
Documentation has been updated accordingly

mkjmdski:
@mgsegner I suggest that you write this explicitly on line 45 regarding `default_branch` property. Otherwise people could miss it because most of us don't read entire documentation :D

Stromweld:
I would check out #671 default branch protection resource doesn’t work on initial project creation. Work around is to disable branch protection on initial branch creation at group level. Then branch protection resource will work. out of the box defaults will throw an error that the branch is already protected even if you have different settings.

mgsegner:
@Stromweld, you are right this commit will not cover every situations.
I thought of your proposal before this merge, but I thought too it was not really clean because a terraform create action would have to either create, update or do nothing and it is a weird behavior for me.
Unfortunately, it seems it is the most appropriate to cover every use cases

If everyone agrees, this PR can be closed and I will work on a new one to handle this situation in the gitlab_branch_protection if no one is already on it.

roidelapluie:
The branch protection should not be done by the provider. As per terraform philosophy, we shoud stick to the original API behaviour, and not do a second call.

armsnyder:
So the result of this change would be that a project initialized with a README would always be created with no protected branch, correct? That would be a different behavior than you would get using the GitLab API directly, where it would respect the branch protection settings configured for the instance or group.

I still wonder if maybe this PR is our best way of solving the issue of GitLab effectively creating a branch protection out-of-band.

This relates to #792 and more specifically #749.

haarchri:
any progress here ? i wonder if we have a blocker to merge

armsnyder:
This PR is currently blocked because we are lacking a consensus on how to approach the problem of the GitLab API creating resources out-of-band that may need to be also specified as Terraform resources.

I would like to see more discussion in #792 before we merge.

Issue #688, opened 2021-08-05
Feature request: resource to create files


spothound:
Hi there,

I am using the terraform provider 3.7.0 and I cannot find a way to create a project with certain files inside.

I would like to ask if there is a way to do this with the Terraform module and, it is not possible, make a feature request for this.

I know that there are [API calls to create files](https://docs.gitlab.com/ee/api/repository_files.html#create-new-file-in-repository) so that should be possible with Terraform as well.

Best regards.

duyleekun:
The team removed that feature due to lacking testing
 https://github.com/gitlabhq/terraform-provider-gitlab/pull/650

spothound:
Thanks, I will check it!

djtecha:
This is a bummer as it would be useful to incorporate for FLUX

timofurrer:
We've lately merged #724 which provides this in the `gitlab_repository_files` resource 🎉

PR #350, opened 2020-06-25
feature: add environment and repository file

BrendanThompson:
This PR contains the code to implement `gitlab_resource_file` and `gitlab_environment`, as per the below checklist a few items are still required before this is good to be merged.

One thing that I would like some assistance with however is the fact that sometimes when using the `gitlab_resource_file` resource type the GitLab API returns a `400 Bad Request` when deleting files as well as creating. This is why I have added a sleep into both create and delete functions. It seems to only occur when a few files are being added or deleted. Any advice here would be greatly appreciated.

# ToDo
- [x] Add tests for `gitlab_resource_file`
- [x] Add tests for `gitlab_environment`
- [x] Update website data for the two new resources

PR #724, opened 2021-10-02
feature: add gitlab_repository_file resource. Closes #720, #688, #350

### LINKED ISSUE #720: [Feature Request] Add ability to push files
Hi!

I could be missing something really obvious here but I can't see any way to push files using the Terraform provider.

I'm trying to setup Flux2 using Terraform [as per this guide](https://registry.terraform.io/providers/fluxcd/flux/latest/docs/guides/github).

Flux also [supports gitlab](https://fluxcd.io/docs/cmd/flux_bootstrap_gitlab/) for bootstrapping.

However at the point in their example where they push the install/sync and Kustomize files to Github using the following resource(s) below, I'm getting stuck as the Gitlab provider doesn't seem to support the ability to push things to git via the provider.

```
resource "github_repository_file" "install" {
  repository = github_repository.main.name
  file       = data.flux_install.main.path
  content    = data.flux_install.main.content
  branch     = var.branch
}

resource "github_repository_file" "sync" {
  repository = github_repository.main.name
  file       = data.flux_sync.main.path
  content    = data.flux_sync.main.content
  branch     = var.branch
}

resource "github_repository_file" "kustomize" {
  repository = github_repository.main.name
  file       = data.flux_sync.main.kustomize_path
  content    = data.flux_sync.main.kustomize_content
  branch     = var.branch
}
```

Does the Gitlab provider have anything like this to allow us to push files to Gitlab?

If so, I cannot see how - could this be made clearer?

If not, could this be added so that we can push files into a project?

timofurrer:
This change adds a new `gitlab_repository_file` resource.
Most of which origins from #350.

The `start_branch` parameter is not working yet.
AFAIK it's a bug in the GitLab API itself,
see https://gitlab.com/gitlab-org/gitlab/-/issues/342200

I've also dug into the `encoding` parameter situation as discussed in #350.
I experience the same issue that nothing except `base64` is accepted as
valid encoding by the GitLab API (in contrary to what is documented).

I've added a new parameter `overwrite_on_create` which can be used to
overwrite already existing files (therefore performing an update instead
of a create). The GitHub provider has the same parameter.


docs/resources/repository_file.md
```diff
@@ -0,0 +1,53 @@
...
+
+The following arguments are supported:
+
+* `project` - (Required) The ID of the project.
+
+* `file_path` - (Required) The full path of the file.
+
+* `branch` - (Required) Name of the branch to which to commit to.
+
+* `start_branch` - (Required) Name of the branch to start the new commit from.
```
mkjmdski, line 39: According to: https://docs.gitlab.com/ee/api/commits.html start branch is not required argument

docs/resources/repository_file.md
```diff
@@ -0,0 +1,53 @@
...
+
+* `author_email` - (Optional) Email of the commit author.
+
+* `author_name` - (Optional) Name of the commit author.
+
+* `content` - (Required) File content.
```
mkjmdski, line 45: I propose that all `(Required)` are moved to the top of this file

gitlab/resource_gitlab_repository_file.go
```diff
@@ -0,0 +1,209 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"author_name": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"content": {
+				Type:     schema.TypeString,
+				Required: true,
```
mkjmdski, line 53: Maybe this should have a validate func for base64?

docs/resources/repository_file.md
```diff
@@ -0,0 +1,53 @@
...
+
+* `commit_message` - (Required) Commit message.
```
mkjmdski, line 47: I propose that all (Required) are moved to the top of this file

docs/resources/repository_file.md
mkjmdski, line 45: We should probably inform that this is only base64 content

docs/resources/repository_file.md
```diff
@@ -0,0 +1,53 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `project` - (Required) The ID of the project.
+
+* `file_path` - (Required) The full path of the file.
```
mkjmdski, line 35: We should write about the format for file_path, according to the docs it should be relative to the root of the repository without `/` prefix

docs/resources/repository_file.md
timofurrer: good catch. Updated 🎉

docs/resources/repository_file.md
timofurrer: done, sorted them to the documentation order in the API docs. 

Your suggestion makes more sense though!

docs/resources/repository_file.md
timofurrer: done

docs/resources/repository_file.md
timofurrer: done

gitlab/resource_gitlab_repository_file.go
timofurrer: implemented!

gitlab/resource_gitlab_repository_file.go
```diff
@@ -0,0 +1,219 @@
...
+func resourceGitlabRepositoryFileRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gitlab.Client)
+	project := d.Get("project").(string)
+	filePath := d.Id()
+	options := &gitlab.GetFileOptions{
+		Ref: gitlab.String(d.Get("branch").(string)),
+	}
+
+	repositoryFile, _, err := client.RepositoryFiles.GetFile(project, filePath, options)
+	if err != nil {
```
defreng, line 136: it would be great if we could handle the 404 error case a little differently, as in the following example here from the AWS provider:
https://github.com/hashicorp/terraform-provider-aws/blob/main/aws/resource_aws_ami.go#L344

It uses `d.SetId("")` to indicate that the resource no longer exists instead of throwing the API error back to terraform.

From my perspective, this behaviour is highly desirable because it prevents the terraform drift detection stage from crashing if someone manually removed the file before, or the underlying project doesn't exist anymore.

This behavior is present in the entire AWS provider, but missing in the entire Gitlab provider 🙂 not sure what's the best approach here... Start setting an example in this resource or doing a complete refactoring at once...

timofurrer: I've created an issue from this for discussion here: #726 

@mkjmdski what do you think? Should I implement something like this in this PR already or does it make sense to refactor the entire provider all at once?

mkjmdski: I've encountered this issue few times but didn't know the root cause. I think that making this refactor just for this specific resource in this pull request is good idea, and adding it to the others as a part of separated issue is another good idea too :D

gitlab/resource_gitlab_repository_file_test.go
```diff
@@ -0,0 +1,357 @@
+package gitlab
+
```
mkjmdski, line 2: could you add some test for `start_branch` attribitue change? I am not sure if it will not require forcing recreation, we should find this out in the tests.

timofurrer: there actually is a skipped one here: https://github.com/gitlabhq/terraform-provider-gitlab/pull/724/files#diff-939af460cb608f4acf20ee9378fd6898591ec37c96e4ae94bf494c308beb5dd0R59

However, `start_branch` doesn't seem to work properly, see the description of this PR ...

timofurrer:
> Hi @timofurrer. Thanks for the PR! To align with the GitLab API (even if it has a validation bug), the default encoding should be text. Right now this means sending no encoding field in the GitLab request. See my comment here - [#350 (comment)](https://github.com/gitlabhq/terraform-provider-gitlab/pull/350#issuecomment-883617680)

Just pushed a new version which aligns with the GitLab API.

timofurrer:
@armsnyder yes, of course.

timofurrer:
@mkjmdski there is still an open question about the `overwrite_on_create` argument. And I'd like to add the `Importer` today. 

For the former we probably need input from @roidelapluie or @armsnyder

timofurrer:
@mkjmdski @roidelapluie @armsnyder alright, a little update from my side:

The good thing: it now implements a proper Importer

The bad thing: I had to remove the `encoding` again. The main reason for that is, that the `Read` / `Get` API always return the repository file in `base64` encoding leading to the following state issues:

```
=== RUN   TestAccGitlabRepositoryFile_createDefaultEncoding
    testing.go:705: Step 0 error: After applying this step, the plan was not empty:

        DIFF:

        UPDATE: gitlab_repository_file.this
          author_email:   "meow@catnip.com" => "meow@catnip.com"
          author_name:    "Meow Meowington" => "Meow Meowington"
          branch:         "main" => "main"
          commit_message: "feature: add launch codes" => "feature: add launch codes"
          content:        "bWVvdyBtZW93IG1lb3c=" => "meow meow meow"
          encoding:       "base64" => "text"
          file_path:      "meow.txt" => "meow.txt"
          id:             "meow.txt" => "meow.txt"
          project:        "781" => "781"
```

I think it makes sense to just only support `base64` (the only sane option to upload file content embedded in a JSON anyways) and document it accordingly

akselleirv:
Sorry for not getting back to you @timofurrer, but setting the parallelism to 1 was far from optimal since the Terraform code I used was doing more than just adding the files to Gitlab. 

However, I have been working on a workaround for this issue. I managed to create a solution around the issue by synchronizing all the actions when using a for_each on the resource and adding all files in one commit.

Link to the provider: https://registry.terraform.io/providers/akselleirv/gitlabcommit/latest/docs

defreng:
interesting idea!

Another option to work around this issue might be to rename this resource to `gitlab_repository_files` (note the **s**), and add native support for committing multiple files:

```
resource "gitlab_repository_files" "tests" {
  project        = var.gitlab_project_id
  author_name    = "author"
  author_email   = "example"
  commit_message = "msg"
  start_branch   = "main"
  branch         = "main"

  file {
    path = "mytestfile1"
    content = "TEST1"
  }

  file {
    path = "mytestfile2"
    content = "TEST2"
  }
}
```

Then, for the usecase you mentioned, @akselleirv, the terraform [dynamic](https://www.terraform.io/docs/language/expressions/dynamic-blocks.html) block could be used.

akselleirv:
> interesting idea!
> 
> Another option to work around this issue might be to rename this resource to `gitlab_repository_files` (note the **s**), and add native support for committing multiple files:
> 
> ```
> resource "gitlab_repository_files" "tests" {
>   project        = var.gitlab_project_id
>   author_name    = "author"
>   author_email   = "example"
>   commit_message = "msg"
>   start_branch   = "main"
>   branch         = "main"
> 
>   file {
>     path = "mytestfile1"
>     content = "TEST1"
>   }
> 
>   file {
>     path = "mytestfile2"
>     content = "TEST2"
>   }
> }
> ```
> 
> Then, for the usecase you mentioned, @akselleirv, the terraform [dynamic](https://www.terraform.io/docs/language/expressions/dynamic-blocks.html) block could be used.

I believe that approach would also work @defreng. 
It just depends if the maintainers do not find this solution too hacky. The changes required are starting a goroutine from the provider setup and changing `meta` from `*gitlab.Client` to a struct, and then adding the channel in it.

Having support in the official GitLab provider would be great!

mkjmdski:
I like the solution with gitlab_repository_files, however it will still cause issues if somebody decides to loop this resource with for_each, for example for different branches. However I think that the need of community for this feature is so big that we might need to accept this little drawback because blocking managing files with terraform because of some API issues doesn't seem to be better in any terms.

timofurrer:
I agree @mkjmdski. Also, I think that we can iterate over the current implementation and offer improvements once we actually know what kind of issues really matter ...

mkjmdski:
Yeah I also hope that @gitlabhq could actually help solve some issues on their side once this feature is in place. I am more than happy to see some implementation of this feature in the next release even if it's not perfect, as you've mentioned - we can obviously make improvements afterwards.

mkjmdski:
@armsnyder Could we get your opinion on future of that pull request?

mkjmdski:
@roidelapluie could you check if current state of this PR might be merged into future release?

mkjmdski:
Is this repository maintained?

nicholasklick:
@mkjmdski this repository is being maintained. We have many competing priorities and apologize for the delay.

nagyv:
@timofurrer Re the `start_branch` argument, the GitLab PM of this area provided some feedback on the issue you opened: https://gitlab.com/gitlab-org/gitlab/-/issues/342200#note_810898339 Please, check it, and let me know if this PR is okay based on his answer.

I'll try to help you get the PR merged in the following weeks (I'll be out the next week).

timofurrer:
@nagyv just responded in the issue on GitLab - it's okay for me as is.

And it would be nice to finally get this merged. There is also a huge backlog from other open PRs. Let me know if I can be of any help getting the other open ones merged, too ;)

timofurrer:
@nagyv I've resolved the conflicts and enabled the test for the `start_branch` parameter.

The Ci is failing, but it seems to be unrelated to this branch ...

gitlab/resource_gitlab_repository_file.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+}
+
+func resourceGitlabRepositoryFileCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gitlab.Client)
+	project := d.Get("project").(string)
+	filePath := d.Get("file_path").(string)
+
+	var existingRepositoryFile *gitlab.File
+	if d.Get("overwrite_on_create").(bool) {
```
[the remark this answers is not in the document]
timofurrer: btw. I've implemented the importer in the meantime ;)

gitlab/resource_gitlab_repository_file_test.go
timofurrer: I've enabled the tests.

gitlab/resource_gitlab_repository_file.go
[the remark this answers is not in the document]
timofurrer: ... and removed that flag entirely ;)

armsnyder:
@timofurrer The latest CI failure looks related --

```
=== RUN   TestAccGitlabRepositoryFile_create
panic: Invalid address to set: []string{"encoding"}
goroutine 109845 [running]:
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*ResourceData).Set(0xc004703200, 0x126dbf8, 0x8, 0x1081f80, 0xc001c72860, 0x0, 0x0)
	/home/runner/go/pkg/mod/github.com/hashicorp/terraform-plugin-sdk/v2@v2.5.0/helper/schema/resource_data.go:230 +0x371
github.com/gitlabhq/terraform-provider-gitlab/gitlab.resourceGitlabRepositoryFileRead(0xc004703200, 0x11bdf20, 0xc00031b880, 0xc0010372b0, 0x8)
	/home/runner/work/terraform-provider-gitlab/terraform-provider-gitlab/gitlab/resource_gitlab_repository_file.go:131 +0x505
github.com/gitlabhq/terraform-provider-gitlab/gitlab.resourceGitlabRepositoryFileCreate(0xc004703200, 0x11bdf20, 0xc00031b880, 0x0, 0xffffffffffffffff)
```


> And it would be nice to finally get this merged. There is also a huge backlog from other open PRs. Let me know if I can be of any help getting the other open ones merged, too ;)

Every little bit helps! I will make an effort to review any PR where you @mention me. Would you be interested in being added as a collaborator?

timofurrer:
@armsnyder Yes, it definitely is from this change - I'll get this ready tomorrow.

> I will make an effort to review any PR where you @mention me. 

I will do so :)

> Would you be interested in being added as a collaborator?

Yes, please - as I've already briefly discussed with @nagyv today :)

timofurrer:
@armsnyder looks good now :)

gitlab/resource_gitlab_repository_file.go
```diff
@@ -0,0 +1,212 @@
...
+	if startBranch, ok := d.GetOk("start_branch"); ok {
+		options.StartBranch = gitlab.String(startBranch.(string))
+	}
+
+	repositoryFile, _, err := client.RepositoryFiles.CreateFile(project, filePath, options)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(repositoryFile.FilePath)
```
armsnyder, line 110: If the user has two gitlab_repository_file resources for the same file but different projects or branches, wouldn't the ID cause a state conflict? I think we should use the ID format that you have in the importer.

armsnyder (changes requested):

gitlab/resource_gitlab_repository_file.go
timofurrer: Yes, good catch. I've just updated the implementation using `project:branch:filePath` as id and also using the `StatePassThroughContext` importer instead of a custom one.

armsnyder: Perfect

armsnyder (approved):
lgtm

mkjmdski (approved):

nagyv (approved):
LGTM, merging

Issue #729, opened 2021-10-13
Protected Branches on Project Import


briananderson1222:
I would like to understand why this logic exists. https://github.com/gitlabhq/terraform-provider-gitlab/blob/master/gitlab/resource_gitlab_project.go#L519-L531

I understand that I can specify a different default branch than the project I am importing and I like that some of this logic handles updating the new project to use the correct default_branch but I do not understand why assumptions are made about branch protection rules and why those assumptions cause hard errors. I want to be able to define a project and setup my own branch protection rules but this logic ends up either 
1) Causing a failure if there is no branch protection rule on the template project because it fails trying to remove something that is non-existent OR
2) Defines a generic branch protection rule on my new default branch that I can then not control at all with my own branch protection resource

... It seems like this logic assumes that the Group settings for branch protection are always set to "Fully Protected" which is not the case. I know there are other similar-ish issues open related to adding support for controlling the group level settings but this logic seems to be specific to project imports from what I understand..

Can someone shed some light on why this logic is here and if there is an alternative way I could define my own branch protection rules on a project that uses a template or imported project.. Seems like it should either take the group default branch protection rules into account if it wants to create the rule AND not cause a hard failure when trying to remove the old branch protection rule (in case it doesn't exist).

armsnyder:
This logic was originally added while fixing an issue where the `default_branch` attribute was not being respected, in #583. Unfortunately it has caused other issues like the one you're describing.

armsnyder:
Let's fold the discussion into #759.

PR #760, opened 2021-12-03
Check for protected branch before trying to delete protection

### LINKED ISSUE #759: gitlab_project fails to create in group with no branch protection
When creating a new project with `default_branch` defined in a group with `default_branch_protection` set to `0`, creating the project fails with a 404 because on [lines 527-531 of gitlab/resource_gitlab_project.go](gitlab/resource_gitlab_project.go) the provider tries to delete the branch protection without first checking that it exists.

### Terraform Version
0.12.31

### Affected Resource(s)
- resource_gilab_project

### Expected Behavior
Expect to be able to define a default branch other than `main` even when default branch protection for the group is disabled.

### Actual Behavior
When creating a new project with `default_branch` defined in a group with `default_branch_protection` set to `0`, creating the project fails with the below error message:

```shell
Error: Failed to unprotect undesired default branch "main" for project "REDACTED": DELETE https://gitlab.com/api/v4/projects/REDACTED/protected_branches/main: 404 {message: 404 Not found}
  on environments.tf line 23, in resource "gitlab_project" "environments":
  23: resource gitlab_project environments {
```

### Steps to Reproduce
1. Create a Gitlab group with `default_branch_protection` set to `0`
2. Define a `gitlab_project` resource setting `default_branch` to a value other than `main` and `namespace_id` to the ID of the group from step 1
3. `terraform apply`

switchboardOp:
In `resourceGitlabProjectCreate` if `default_branch` is set on the resource, the provider tries to delete branch protection on the old default branch even if it does not exist. If the branch isn't protected this returns a 404 and the provider fails.

This PR adds a check first so that if the old default branch is not protected, the provider can proceed to delete the old default branch, and the resource can be created.

fixes #759

### DIFF
--- a/gitlab/resource_gitlab_project.go
+++ b/gitlab/resource_gitlab_project.go
@@ -524,10 +524,19 @@ func resourceGitlabProjectCreate(d *schema.ResourceData, meta interface{}) error
 			return fmt.Errorf("Failed to protect default branch %q for project %q: %w", newDefaultBranch, d.Id(), err)
 		}
 
-		log.Printf("[DEBUG] unprotect old default branch %q for project %q", oldDefaultBranch, d.Id())
-		_, err = client.ProtectedBranches.UnprotectRepositoryBranches(project.ID, oldDefaultBranch)
-		if err != nil {
-			return fmt.Errorf("Failed to unprotect undesired default branch %q for project %q: %w", oldDefaultBranch, d.Id(), err)
+		log.Printf("[DEBUG] check for protection on old default branch %q for project %q", oldDefaultBranch, d.Id())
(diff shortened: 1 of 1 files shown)


briananderson1222:
Sounds similar to https://github.com/gitlabhq/terraform-provider-gitlab/issues/729 but did not verify

switchboardOp:
@briananderson1222 Yes this is an attempt to resolve one of the hard failures you mentioned in #729, specifically the 404 that's returned when you try to create a project with a non-default default branch in a group with no default branch protection.

armsnyder:
Thanks for the PR. Could you add a regression test? Otherwise there is too much risk of reintroducing the behavior in the future as we untangle this resource.

PR #762, opened 2021-12-06
closes #631, adding the support to block users

### LINKED ISSUE #631: Add the ability to block a user
Hello, this is a feature request for adding an argument `is_blocked` (boolean, default `false`) to the `gitlab_user` resource. Thanks for your consideration.

hallaj:
Just adding a basic implementation for blocking user in the terraform provider

closes #631

### DIFF
--- a/docs/resources/user.md
+++ b/docs/resources/user.md
@@ -51,7 +51,7 @@ resource "gitlab_user" "example" {
 - **projects_limit** (Number) Integer, defaults to 0.  Number of projects user can create.
 - **reset_password** (Boolean) Boolean, defaults to false. Send user password reset link.
 - **skip_confirmation** (Boolean) Boolean, defaults to true. Whether to skip confirmation.
-- **state** - (Optional) String, defaults to "active". Sets the state of the user.
+- **state** (String) String, defaults to 'active'. The state of the user account.
 
 ## Import
 
--- a/docs/resources/user.md
+++ b/docs/resources/user.md
@@ -51,6 +51,7 @@ resource "gitlab_user" "example" {
 - **projects_limit** (Number) Integer, defaults to 0.  Number of projects user can create.
 - **reset_password** (Boolean) Boolean, defaults to false. Send user password reset link.
 - **skip_confirmation** (Boolean) Boolean, defaults to true. Whether to skip confirmation.
+- **state** - (Optional) String, defaults to "active". Sets the state of the user.
 
 ## Import
 
--- a/docs/resources/user.md
+++ b/docs/resources/user.md
@@ -51,7 +51,7 @@ resource "gitlab_user" "example" {
 - **projects_limit** (Number) Integer, defaults to 0.  Number of projects user can create.
 - **reset_password** (Boolean) Boolean, defaults to false. Send user password reset link.
 - **skip_confirmation** (Boolean) Boolean, defaults to true. Whether to skip confirmation.
-- **state** (String) String, defaults to 'active'. The state of the user account.
+- **state** (String) String, defaults to 'active'. The state of the user account. Valid values are either 'active' or 'blocked'
 
 ## Import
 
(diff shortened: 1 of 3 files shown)


gitlab/resource_gitlab_user.go
```diff
@@ -182,6 +188,22 @@ func resourceGitlabUserUpdate(d *schema.ResourceData, meta interface{}) error {
...
 	}
 
+	if d.HasChange("state") {
+		if d.Get("state") == "active" {
+			err := client.Users.UnblockUser(id)
+
+			if err != nil {
+				return err
+			}
+		} else if d.Get("state") == "deactivated" {
```
armsnyder, line 198: This should be `blocked` not `deactivated` right?

gitlab/resource_gitlab_user_test.go
```diff
@@ -146,6 +147,7 @@ type testAccGitlabUserExpectedAttributes struct {
 	SkipConfirmation bool
 	External         bool
 	Note             string
+	State            string
 }
 
 func testAccCheckGitlabUserAttributes(user *gitlab.User, want *testAccGitlabUserExpectedAttributes) resource.TestCheckFunc {
```
armsnyder, line 153: You'll need to update this function to check the new attribute

armsnyder (changes requested):
Thanks for the PR! 😄 Could you rebase please?

We've made some significant changes recently by moving to the Terraform SDK V2. You'll also need to add a `Description` field to the new attribute, and then run `make reviewable` to generate documentation.

See the CONTRIBUTING.md for more up to date info.

gitlab/resource_gitlab_user.go
hallaj: Thanks for catching that, fixed.

gitlab/resource_gitlab_user_test.go
hallaj: Created a new one instead below to cover the blocked use case. Hope that's fine?

gitlab/resource_gitlab_user.go
```diff
@@ -105,6 +111,7 @@ func resourceGitlabUserSetToState(d *schema.ResourceData, user *gitlab.User) {
 	d.Set("is_admin", user.IsAdmin)
 	d.Set("is_external", user.External)
 	d.Set("note", user.Note)
+	d.Set("state", user.State)
 }
 
 func resourceGitlabUserCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
armsnyder, line 117: Can you add this to the "create" function _after_ the call to `d.SetId` --

```go
if d.Get("state") == "blocked" {
	if err := client.Users.BlockUser(id); err != nil {
		return diag.FromErr(err)
	}
}
```

gitlab/resource_gitlab_user.go
```diff
@@ -92,6 +92,12 @@ func resourceGitlabUser() *schema.Resource {
 				Type:        schema.TypeString,
 				Optional:    true,
 			},
+			"state": {
+				Description: "String, defaults to 'active'. The state of the user account.",
```
armsnyder, line 96: The description should mention "blocked" as well. You could also add a field here - `ValidateFunc: validation.StringInSlice([]string{"active", "blocked"}, false)` - to validate values.

gitlab/resource_gitlab_user_test.go
armsnyder: Looks good!

armsnyder (changes requested):

gitlab/resource_gitlab_user.go
hallaj: Added

gitlab/resource_gitlab_user.go
hallaj: Added both

armsnyder (approved):
lgtm, thanks!

PR #763, opened 2021-12-08
resolves #465: add gitlab_group_ldap_link filter support

### LINKED ISSUE #465: gitlab_group_ldap_link should allow filter
Currently, gitlab_group_ldap_link only allows for setting the `cn` but not the `filter`. It'd be neat if `filter` also worked. See [API docs](https://docs.gitlab.com/ee/api/groups.html#add-ldap-group-link-with-cn-or-filter). Take note of the fact that only either `cn` or `filter` may be set.

bittracker:
Added LDAP filter support, see [1] and [2] for more information.

^[1] https://pkg.go.dev/github.com/xanzy/go-gitlab#AddGroupLDAPLinkOptions
^[2] https://docs.gitlab.com/ee/api/groups.html#add-ldap-group-link-starter

### DIFF
--- a/gitlab/resource_gitlab_group_ldap_link.go
+++ b/gitlab/resource_gitlab_group_ldap_link.go
@@ -138,7 +138,6 @@ func resourceGitlabGroupLdapLinkDelete(d *schema.ResourceData, meta interface{})
 	client := meta.(*gitlab.Client)
 	groupId := d.Get("group_id").(string)
 	cn := d.Get("cn").(string)
-	filter := d.Get("filter").(string)
 	ldap_provider := d.Get("ldap_provider").(string)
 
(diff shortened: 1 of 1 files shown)


armsnyder:
@bittracker Can you rebase this?

ielshal:
Any updates here .. i see it only requires rebase then it can be merged .. @bittracker could you please check and rebase or not sure may be someone else can also rebase ?

timofurrer (changes requested):
@ielshal there is more missing than just the rebase / conflict solving: e.g. the `filter` isn't currently in the resource Schema, thus this would never work. It would also be nice to have a test case :)

timofurrer:
I think there is also a little bit more to that `filter` field, see #869 ...

PR #766, opened 2021-12-12
Fix docs for gitlab_project_level_mr_approvals import

coding4food:
The only id in GitLab API for project MR approvals is project id.


armsnyder (approved):
Thanks for the fix!

Issue #448, opened 2020-09-25
Resource gitlab_group_ldap_link has incorrect import documentation


grv87:
### Terraform Version
0.13.3

### Provider Version
3.0.0

### Affected Resource(s)
- `gitlab_group_ldap_link`

### Terraform Configuration Files
```hcl
resource "gitlab_group_ldap_link" "ldap_group" {
  group_id = 123
  cn = "XXX"
  access_level = "owner"
  ldap_provider = "ldapmain"
}
```

`terraform import gitlab_group_ldap_link.ldap_group "ldapmain:XXX`

### Debug Output
https://gist.github.com/grv87/cd8d712124c0939807b8046b6c4653db

### Expected Behavior
According to [documentation](https://registry.terraform.io/providers/gitlabhq/gitlab/latest/docs/resources/group_ldap_link#import), import succeeded

### Actual Behavior
```
Error: resource gitlab_group_ldap_link doesn't support import
```

grv87:
@armsnyder, why it is marked as enhancement when documentation explicitly promises that import is supported? I believe this is a bug in either code or documentation.

armsnyder:
@grv87 I had misinterpreted the issue as wanting to add the ability to import.

PR #328, opened 2020-06-04
Creating custom skip function for group_ldap_link tests.

jasoncarlsen:

PR #771, opened 2021-12-18
fix(group_ldap_link): add missing terraform import feature for group LDAP link

hanks:
**Why**

There is a bug that [the doc](https://registry.terraform.io/providers/gitlabhq/gitlab/latest/docs/resources/group_ldap_link#import) says we can use `terraform import`, but actually the function is not implemented

**Code Changes**

* Added the missing import function and acceptance test case
* ~Fixed a bug to set `access_level` instead of `group_access`, because `access_level` is the one defined in schema~
   * ~Assigned the correct string value to `group_access`, before is an int type~, already fixed in the latest main branch
* Set `force=false` as default when importing, because we can not get any value from the Gitlab API
* Updated the doc about import part

**Design Changes**

* Added `group_id` in the import id string to change it from `ldap_provider:cn` to  `group_id:ldap_provider:cn`, because we need the `group_id` to identify the group

**Tests**

* built local provider, and tested
* added acceptance test

**Close**

* #448

### DIFF
--- a/docs/resources/group_ldap_link.md
+++ b/docs/resources/group_ldap_link.md
@@ -37,4 +37,8 @@ resource "gitlab_group_ldap_link" "test" {
 - **group_access** (String) Minimum access level for members of the LDAP group. Valid values are: `no one`, `minimal`, `guest`, `reporter`, `developer`, `maintainer`, `owner`, `master`
 - **id** (String) The ID of this resource.
 
+GitLab group ldap links can be imported using an id made up of `group_id:ldap_provider:cn`, e.g.
 
+```shell
+terraform import gitlab_group_ldap_link.test "12345:ldapmain:testuser"
+```
--- a/docs/resources/group_ldap_link.md
+++ b/docs/resources/group_ldap_link.md
@@ -37,8 +37,11 @@ resource "gitlab_group_ldap_link" "test" {
 - **group_access** (String) Minimum access level for members of the LDAP group. Valid values are: `no one`, `minimal`, `guest`, `reporter`, `developer`, `maintainer`, `owner`, `master`
 - **id** (String) The ID of this resource.
 
-GitLab group ldap links can be imported using an id made up of `group_id:ldap_provider:cn`, e.g.
+## Import
+
+Import is supported using the following syntax:
 
 ```shell
+# GitLab group ldap links can be imported using an id made up of `group_id:ldap_provider:cn`, e.g.
 terraform import gitlab_group_ldap_link.test "12345:ldapmain:testuser"
 ```
--- a/examples/resources/gitlab_group_ldap_link/import.sh
+++ b/examples/resources/gitlab_group_ldap_link/import.sh
@@ -1,0 +1,2 @@
+# GitLab group ldap links can be imported using an id made up of `group_id:ldap_provider:cn`, e.g.
+terraform import gitlab_group_ldap_link.test "12345:ldapmain:testuser"
--- a/internal/provider/resource_gitlab_group_ldap_link.go
+++ b/internal/provider/resource_gitlab_group_ldap_link.go
@@ -20,6 +20,9 @@ var _ = registerResource("gitlab_group_ldap_link", func() *schema.Resource {
 		CreateContext: resourceGitlabGroupLdapLinkCreate,
 		ReadContext:   resourceGitlabGroupLdapLinkRead,
 		DeleteContext: resourceGitlabGroupLdapLinkDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceGitlabGroupLdapLinkImporter,
+		},
 
 		Schema: map[string]*schema.Schema{
 			"group_id": {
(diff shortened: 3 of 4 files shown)


gitlab/resource_gitlab_group_ldap_link.go
```diff
@@ -20,6 +20,9 @@ func resourceGitlabGroupLdapLink() *schema.Resource {
 		Create: resourceGitlabGroupLdapLinkCreate,
 		Read:   resourceGitlabGroupLdapLinkRead,
 		Delete: resourceGitlabGroupLdapLinkDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceGitlabGroupLdapLinkImporter,
```
timofurrer, line 24: It would be nice to use the `schema.ImportStatePassthroughContext` importer instead of writing our custom one.

To make use of that the `Read` function must be able to set the state on the id from `SetId()` only.
See https://www.terraform.io/plugin/sdkv2/resources/import#importer-state-function

I suggest to use the `group_id:lda_provider:cn` three part id. I'm aware that this requires a little refactoring. Please let me know if you want to continue working on this or not.

timofurrer (changes requested):
Please rebase und run `make reviewable` to generate the documentation.

timofurrer:
@hanks thanks for the follow up. Coincidently, I have had a look at this yesterday - see my outcome here #869.

I also doubt that this change isn't backwards-compatible, because you've changed the id. 

Also, the field for the access level is actually called `group_access` and not `access_level` - The `access_level` attribute is deprecated. 

I suggest that we await the outcome of the #869 discussion :)

timofurrer:
@hanks if you want to push this further: me and @armsnyder had some discussion in #869 and came to the conclusion that at the moment it might be best to implement a custom importer than can import something like `<group_id>:<ldap_provider>:<cn>` and set the state as it is right now, that is writing the `group` state attribute and id `<ldap_provider>:<cn>`. The reason for this is to not break backwards-compatibility.

Would you like to continue working on this? 🎉

timofurrer (changes requested):
@hanks if you want to push this further: me and @armsnyder had some discussion in #869 and came to the conclusion that at the moment it might be best to implement a custom importer than can import something like `<group_id>:<ldap_provider>:<cn>` and set the state as it is right now, that is writing the `group` state attribute and id `<ldap_provider>:<cn>`. The reason for this is to not break backwards-compatibility.

Would you like to continue working on this? 🎉

hanks:
I tried to add the importer test, and before I added the `SkipFunc` to each step, it failed only in `acceptance-ce` with errors below, and then I added the `SkipFunc`, it seems to pass all the cases. Not sure why in ce test, the `terraform apply` is failed.

```
=== RUN   TestAccGitlabGroupLdapLink_import
    resource_gitlab_group_ldap_link_test.go:73: Step 1/2 error: Error running apply: exit status 1
        
        Error: POST http://127.0.0.1:8080/api/v4/groups/168/ldap_group_links: 404 {error: 404 Not Found}
        
          with gitlab_group_ldap_link.foo,
          on terraform_plugin_test.tf line 8, in resource "gitlab_group_ldap_link" "foo":
           8: resource "gitlab_group_ldap_link" "foo" {
        
--- FAIL: TestAccGitlabGroupLdapLink_import (5.96s)
```

timofurrer:
The [LDAP Group Links](https://docs.gitlab.com/ee/api/groups.html#ldap-group-links) are a EE feature :)

timofurrer:
Okay, quick update ... There used to be a testdata file to configure the provider, which I've removed a few days ago in #859. 

The original commit https://github.com/gitlabhq/terraform-provider-gitlab/pull/328/files#diff-e864e9b96eaea85189143b708a2f8f57e3c38fe11d57586fd765d6c65274a6f8L29 used to have a warning:

> t.Skipf("[WARNING] Skipping test until test data is configured in %s.", testDataFile)

why the tests are skipped.

Apparently, it was meant to be run with a valid LDAP provider somewhere. However, from this provider perspective we don't actually care too much.
It seems to work fine with using `default` values here.

About the skipped tests I've just created a PR: 

* https://github.com/gitlabhq/terraform-provider-gitlab/pull/873 

I hope that we can quickly merge it and unblock you here - in the meantime you can try and remove the skip func, or at least add the `isRunningInCE` and remove the third step from the `_basic` acceptance test and implement your import test there :)

timofurrer:
@hanks In the meantime I've merged #873 which enables the tests - please rebase , it should work for you now :)

hanks:
@timofurrer  

> The [LDAP Group Links](https://docs.gitlab.com/ee/api/groups.html#ldap-group-links) are a EE feature

good to know 😄 

thanks for the updates for the test code, and I inserted the import test as the second one, to reuse the Config created from the first step

hanks:
@timofurrer Hi, anything else needs to be changed?

timofurrer (approved):

timofurrer:
@hanks nothing else from your side, just some time from our side! 🎉

PR #784, opened 2022-01-18
Implement configuration of the integration "Microsoft Teams" (#308)

### LINKED ISSUE #308: [Feature Request] gitlab_service_teams
Morning,

I see there are specific resources for slack, jira etc…. to make configuring other integrations simple, with the popularity of Microsoft Teams increasing as of late and would be good if there was a dedicated resource for this as there is with slack etc…

Thanks! 

Pete

MichelHartmann:
This change adds the capability to manage the integration of GitLab with Microsoft Teams.

Fixes #308

### DIFF
--- a/gitlab/resource_gitlab_service_microsoft_teams.go
+++ b/gitlab/resource_gitlab_service_microsoft_teams.go
@@ -1,0 +1,228 @@
+package gitlab
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	gitlab "github.com/xanzy/go-gitlab"
+)
+
+func resourceGitlabServiceMicrosoftTeams() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGitlabServiceMicrosoftTeamsCreate,
+		Read:   resourceGitlabServiceMicrosoftTeamsRead,
+		Update: resourceGitlabServiceMicrosoftTeamsUpdate,
+		Delete: resourceGitlabServiceMicrosoftTeamsDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceGitlabServiceMicrosoftTeamsImportState,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"project": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"title": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"created_at": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"updated_at": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"active": {
+				Type:     schema.TypeBool,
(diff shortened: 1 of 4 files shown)


gitlab/resource_gitlab_service_microsoft_teams_test.go
```diff
@@ -0,0 +1,224 @@
...
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckGitlabServiceMicrosoftTeamsDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGitlabServiceMicrosoftTeamsConfig(rInt),
+			},
+			{
+				ResourceName:      teamsResourceName,
+				ImportStateIdFunc: getTeamsProjectID(teamsResourceName),
```
armsnyder, line 99: Why is this `ImportStateIdFunc` field needed?

gitlab/resource_gitlab_service_microsoft_teams.go
```diff
@@ -0,0 +1,228 @@
...
+)
+
+func resourceGitlabServiceMicrosoftTeams() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGitlabServiceMicrosoftTeamsCreate,
+		Read:   resourceGitlabServiceMicrosoftTeamsRead,
+		Update: resourceGitlabServiceMicrosoftTeamsUpdate,
+		Delete: resourceGitlabServiceMicrosoftTeamsDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceGitlabServiceMicrosoftTeamsImportState,
```
armsnyder, line 19: This resource can use [schema.ImportStatePassthrough](https://www.terraform.io/plugin/sdkv2/resources/import#importer-state-function) which is preferred. You would need to update your *Read function to parse the ID instead of getting the project attribute.

armsnyder (changes requested):
Thanks for the contribution! The code looks excellent, just a couple comments.

gitlab/resource_gitlab_service_microsoft_teams.go
```diff
@@ -0,0 +1,181 @@
...
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"project": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"title": {
```
armsnyder, line 28: Can you double check your attributes one more time? `title` does not appear to be used, and it is not mentioned in the API docs. https://docs.gitlab.com/ee/api/integrations.html#createedit-microsoft-teams-integration

Also a reminder to double check that each attribute is listed on the new documentation page.

MichelHartmann: Sorry, my mistake: The `title` is returned by the API, but apparently only used in the frontend (value is "Microsoft Teams notifications"). I'll remove it and check all other attriutes another time.

armsnyder (approved):
lgtm, thanks!

PR #725, opened 2021-10-04
Upgrade to SDK v2. Closes #552

timofurrer:
This is an actual working upgrade to the Terraform Plugin SDK v2 (not just updating the module reference as e.g. #461) does.

On the way a few lint errors could be removed, because I've switched to non-deprecated features (however, there are more to be updated - on that note: this PR is already quite large, therefore I suggest to get this merged before doing any more changes).

closes #552

PR #791, opened 2022-01-23
Use context-aware CRUD functions and use it in gitlab API calls

timofurrer:
This is a follow up of #725 which uses the context-aware CRUD functions for resources and data sources.
It also uses this contexts for the GitLab API calls.

See https://www.terraform.io/plugin/sdkv2/guides/v2-upgrade-guide#more-support-for-context-context

I've created separate commits for each resource / data source for better traceability of such a massive change set.

### DIFF
--- a/gitlab/data_source_gitlab_group.go
+++ b/gitlab/data_source_gitlab_group.go
@@ -1,16 +1,18 @@
 package gitlab
 
 import (
+	"context"
 	"fmt"
 	"log"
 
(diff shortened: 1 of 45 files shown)


armsnyder (approved):
Incredible! Let's fast track this. The changes look very rote to me, so I think we should merge it unless you had any questions.

timofurrer:
No questions, LGTM.

timofurrer:
@nagyv can you merge this please?

PR #802, opened 2022-01-26
Auto Generate Docs

armsnyder:
### Overview

Fixes #543

This PR adds automatic generation for the entire [/docs](https://github.com/gitlabhq/terraform-provider-gitlab/tree/gen-docs-2/docs) directory, based on the `Description` schema fields in-code and examples in [/examples](https://github.com/gitlabhq/terraform-provider-gitlab/tree/gen-docs-2/examples), as recommended in the [Terraform docs guide](https://www.terraform.io/registry/providers/docs#generating-documentation). It also validates that documentation was generated correctly during the CI workflow.

I am very excited to have this feature, because it will lift a huge burden on PR reviewers to check accuracy of documentation.

Since this PR includes the newly generated documentation, it also fixes all of our existing documentation inaccuracies.

### Notes

* I consolidated our CI workflows for push/pull_request events into a single `test.yaml` workflow file.
* The markdown linter is no longer needed, since markdown is now generated. I added new linters for the Terraform and shell code in the [/examples](https://github.com/gitlabhq/terraform-provider-gitlab/tree/gen-docs-2/examples) directory.
* The tfproviderlint linter is now enforcing [XR005](https://github.com/bflad/tfproviderlint/tree/main/xpasses/XR005) and [XS001](https://github.com/bflad/tfproviderlint/tree/main/xpasses/XS001) which require descriptions in schemas for the purpose of doc generation.
* I moved our tool dependencies out of `tools.go` and into the a [/tools](https://github.com/gitlabhq/terraform-provider-gitlab/tree/gen-docs-2/tools) directory, which is its own module, in the style of the [cloudflare](https://github.com/cloudflare/terraform-provider-cloudflare/blob/master/tools/tools.go) and [newrelic](https://github.com/newrelic/terraform-provider-newrelic/blob/main/tools/tools.go) providers. The upshot is that `make` is slightly faster, and we don't pollute our `go.mod` and `go.sum` with dev dependencies, and dependabot is still able to handle tool version updates.
* I added a single `make` target `make reviewable` as a one-stop-shop for linting and now generating code, inspired by [crossplane](https://github.com/crossplane/crossplane/blob/41bcc71c7703e33e5bd51dd1852fd620269313f2/CONTRIBUTING.md?plain=1#L466).
* The rest of this PR is just the moving of descriptions from the existing [/docs](https://github.com/gitlabhq/terraform-provider-gitlab/tree/gen-docs-2/docs) into the code, and then finally re-generating the documentation.
[...]

PR #803, opened 2022-01-26
Improve documentation around authentication with GitLab. Closes #557

### LINKED ISSUE #557: Documentation: This module needs to define what the required token scope is
The docs for this module, and the Gitops demo at https://gitlab.com/gitops-demo/readme both fail to cover any information about the token required for this module.

Is it:
- A deploy token?
- A users personal access token?
- A group token?

Does it's access need to be:
- Scoped to everything?
- Scoped to just the API?

Really need to a paragraph here - or link to Gitlab documentation. 

It's overly vague at the moment.

timofurrer:


armsnyder (approved):
lgtm. You would normally need to update the markdown too, but as long as we're moving forward with #802 we'll sort it out in the conflict resolution.

timofurrer:
@armsnyder I'm aware and was by intention 🎉

Issue #553, opened 2021-02-05
Schema Audit


armsnyder:
I noticed some inconsistencies in our schemas
* Inconsistencies between attribute names and/or structures between the provider schemas and the upstream GitLab API schemas. (Example: [gitlab_user](https://github.com/gitlabhq/terraform-provider-gitlab/blob/18e9201a44dc2f7867eaf503654b5d923b9b6df7/gitlab/resource_gitlab_user.go#L42) attribute `is_admin` vs [API](https://docs.gitlab.com/ce/api/users.html#user-creation) field `admin`)
* Inconsistencies between attribute names across different resources in this provider, for attributes that have the same function. (Example: `project` vs `id` vs `project_id` vs `full_path` all being used to mean a project name/id)

Correcting inconsistencies will have these benefits
* More consistency between this provider and other API tools users may be using to manage GitLab.
* Increasing the number of resources that could be updated using code generation as the GitLab API evolves.

This will have to be a breaking change and a new major release. I'm thinking of coordinating this with upgrading to the [SDK v2](https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html).

timofurrer:
I think it doesn't make sense to do this now with SDK v2, since users are actually not affected by the SDK upgrade ... Let's schedule a proper v4 release and assign things there we want to have in :)

armsnyder:
Yep my opinion on that changed given how long sdk v2 has been supported in terraform core 👍

armsnyder:
`group` vs `group_id` is another inconsistency 😭

Shocktrooper:
@armsnyder Should we start making a tabled list in the description of this issue? 1 column being what variations there are of a specific variable/attribute, another being what files/resources it exists in and the third column being what the agreed upon final value should be?

PR #806, opened 2022-01-27
Properly allow arguments `id` or `path_with_namespace` for project data source

timofurrer:
The problem with the `gitlab_project` data source was that if the id is a path with namespace and the reading of
the data source is delayed until `apply`, e.g. seeing something like
this:

```
  # module.main.data.gitlab_project.some_project will be read during apply
  # (config refers to values not yet known)
 <= data "gitlab_project" "some_project"  {
      + archived                         = (known after apply)
      + default_branch                   = (known after apply)
      + description                      = (known after apply)
      + http_url_to_repo                 = (known after apply)
      + id                               = "timo.furrer/some-project"
      + issues_enabled                   = (known after apply)
      + lfs_enabled                      = (known after apply)
      + merge_requests_enabled           = (known after apply)
      + name                             = (known after apply)
      + namespace_id                     = (known after apply)
      + path                             = (known after apply)
      + path_with_namespace              = (known after apply)
      + pipelines_enabled                = (known after apply)
      + push_rules                       = (known after apply)
      + remove_source_branch_after_merge = (known after apply)
      + request_access_enabled           = (known after apply)
      + runners_token                    = (known after apply)
      + snippets_enabled                 = (known after apply)
      + ssh_url_to_repo                  = (known after apply)
      + visibility_level                 = (known after apply)
      + web_url                          = (known after apply)
      + wiki_enabled                     = (known after apply)
```

The `id` is the `path_with_namespace`, which ends up like that in the
state (as the string path). However, the data source in fact stores
the actually project id (which it gets from the API) in the state durig
apply
(see https://github.com/gitlabhq/terraform-provider-gitlab/blob/8bc5606d786d73cb9e1947d0f9bc941c2e4f61b0/gitlab/data_source_gitlab_project.go#L173)
which leads to an inconsistent final plan in dependent resources as
shown here:

```
╷
│ Error: Provider produced inconsistent final plan
│
│ When expanding the plan for
│ module.main.gitlab_repository_file.some_file
│ to include new values learned so far during apply, provider
│ "registry.terraform.io/timofurrer/gitlab-repository-files" produced an
│ invalid new value for .project: was
│ cty.StringVal("timo.furrer/some-project"), but now
│ cty.StringVal("30716").
│
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
╵
```

The root cause for this is that the data source always stores the
actual project id in the state no matter if a `path_with_namespace` was
actually given. The `id` attribute is NOT marked as being `Computed`,
but it sometimes is (when the `path_with_namespace` is given).

We cannot mark the `id` attribute as `Required` and `Computed`, because
this is not allowed by terraform (makes sense :D). Therefore, I've used
the same pattern as in other data sources (althought I'd argue that the
handling is not quite consistent here across them :( ).

The implementation here should be (please verify) backwards compatible.
However, people can still run into the same issue as before when
specifying the `path_with_namespace` in the `id` argument, but can
reasolve it by switching to the `path_with_namespace` argument ;)

For long term, and when we have a major release I suggest to break the
interface in these data sources to make it more consistent.

### DIFF
--- a/gitlab/data_source_gitlab_project.go
+++ b/gitlab/data_source_gitlab_project.go
@@ -19,7 +19,13 @@ func dataSourceGitlabProject() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"id": {
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
+				Computed: true,
+				ExactlyOneOf: []string{
+					"id",
+					"path_with_namespace",
+				},
+				Description: "The ID of the project",
 			},
 			"name": {
 				Type:     schema.TypeString,
@@ -31,7 +37,13 @@ func dataSourceGitlabProject() *schema.Resource {
 			},
 			"path_with_namespace": {
 				Type:     schema.TypeString,
+				Optional: true,
 				Computed: true,
+				ExactlyOneOf: []string{
+					"id",
+					"path_with_namespace",
+				},
+				Description: "The full path of the project including the namespace",
 			},
 			"description": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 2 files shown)


gitlab/data_source_gitlab_project.go
```diff
@@ -31,7 +36,12 @@ func dataSourceGitlabProject() *schema.Resource {
 			},
 			"path_with_namespace": {
 				Type:     schema.TypeString,
+				Optional: true,
 				Computed: true,
+				ConflictsWith: []string{
```
armsnyder, line 41: You should use `ExactlyOneOf` instead 😉

armsnyder (changes requested):
Nice writeup. Relates to #553 that we are very inconsistent about `project` vs `id` vs `project_id` vs `full_path` all being used to mean a project name/id. 😭 Some of this we may be able to fix with [state migrations](https://www.terraform.io/plugin/sdkv2/resources/state-migration). The rest will have to be a v4 initiative. Anyway this looks backward compatible to me 👍

gitlab/data_source_gitlab_project.go
timofurrer: done :)

armsnyder (approved):
lgtm

PR #807, opened 2022-01-27
Dependabot/go modules/GitHub.com/hashicorp/terraform plugin sdk/v2 2.10.1

willianpaixao:
Bumping the SDK version  to 2.10 seems to break compatibility with Go 1.15. See [here](https://github.com/gitlabhq/terraform-provider-gitlab/runs/4924487001).

I'm adding one commit on top of [gitlabhq/terraform-provider-gitlab/pull/799](https://github.com/gitlabhq/terraform-provider-gitlab/pull/799) to drop G 1.15 from the unit tests.
Briefly tried to add Go 1.17 but there are refactors to be made, will fill an issue later.


armsnyder (approved):
nice, thanks!

PR #811, opened 2022-01-27
Fix group variable test

### LINKED ISSUE #810: Broken master because of group variable tests in EE
I suspect that the merge of #717 failed our EE Acceptance Tests, see https://github.com/gitlabhq/terraform-provider-gitlab/runs/4971848088?check_suite_focus=true#step:6:142

It's quite later here already, so I'll keep this as a reminder if it won't be fixed in the meantime .. 

/cc @armsnyder @nagyv

armsnyder:
## Description

fixes #810 

There was a bug in one of the test checks. The provider code was correct.

## PR Checklist

- [x] Docs are updated with any new resources or attributes, including how to import the resource.
- [x] New resources should have at minimum a basic test with three steps:
    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/gitlab/provider_test.go
+++ b/gitlab/provider_test.go
@@ -2,6 +2,7 @@ package gitlab
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


timofurrer (approved):
Lgtm

PR #815, opened 2022-01-28
Change provider structure to best practices from terraform-provider-scaffolding project

timofurrer:
At first glance, this change set appears to be quite huge, but actually isn't.

It basically reproduces the best practice provider structure and setup from the [terraform-provider-scaffolding](https://github.com/hashicorp/terraform-provider-scaffolding) project.

The major changes are that:

* provider code moved from `gitlab` to `internal/provider` directory
* change provider code from `gitlab` to `provider` package
* used scaffolding setup in `main.go` and `provider.go`
* For the test code we don't re-use the Gitlab client from the provider, but I've introduced a `testGitlabClient` which is separate from the provider Gitlab client.
* the rest of the changes were all necessary evil


main.go
```diff
@@ -1,11 +1,37 @@
...
+
+	"github.com/gitlabhq/terraform-provider-gitlab/internal/provider"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/plugin"
 )
 
+var (
+	// these will be set by the goreleaser configuration
+	// to appropriate values for the compiled binary
+	version string = "dev"
+)
```
armsnyder, line 16: Is this hooked up, and if not can you open a new issue for setting the appropriate ldflags during release?

main.go
```diff
@@ -1,11 +1,37 @@
...
+
+	var debugMode bool
+
+	flag.BoolVar(&debugMode, "debug", false, "set to true to run the provider with support for debuggers like delve")
+	flag.Parse()
+
+	opts := &plugin.ServeOpts{ProviderFunc: provider.New(version)}
+
+	if debugMode {
+		// TODO: update this string with the full name of your provider as used in your configs
```
armsnyder, line 28: Can you remove this TODO

internal/provider/helper_test.go
```diff
@@ -1,7 +1,6 @@
-package gitlab
+package provider
```
armsnyder, line 1: There are some helper functions in this file that I think go with your change of adding the `testGitlabClient` variable. Can we retire the `testAccNewClient` function, remove the `client *gitlab.Client` argument from the other functions, and remove the `client` field from `testAccGitlabProjectContext`?

armsnyder (changes requested):
Lots of cool sdk v2 improvements!

I just want to acknowledge that moving code to `internal` is going to break almost all PRs and anyone who might be using this repository as a package (which they shouldn't.) What are the benefits of moving code to `internal` now in your opinion?

I looked at some other providers. The major first-party ones have largely moved to `internal`, and the major third-party ones have not yet.

timofurrer:
@armsnyder 

> I just want to acknowledge that moving code to `internal` is going to break almost all PRs and anyone who might be using this repository as a package (which they shouldn't.) What are the benefits of moving code to `internal` now in your opinion?

Yes, I actually share the same hesitation towards this change. First of all, I wouldn't want to merge this before 3.9.0 is out. The reason I went for the change anyways (to be merged after 3.9.0) is because I don't foresee that there will be an "ideal" time to make it. Also I think a v4 wouldn't necessarily be it, because the target audience for the "break" is another one. 

IMHO: moving it to `internal` doesn't provide huge benefits other than (1) that it's aligned (as you mentioned) with other popular providers and the scaffolding project and (2) that no one is using it a go dependency.

main.go
timofurrer: yes, it already was: https://github.com/gitlabhq/terraform-provider-gitlab/blob/6c1a60d79fc76012287e21696384c3c019639ed2/.goreleaser.yml#L8

GNUmakefile
```diff
@@ -41,7 +42,7 @@ lint-examples-sh: tools ## Run shell linter on examples (same as fmt but without
 	$(GOBIN)/shfmt -l -s -d ./examples
 
 SERVICE ?= gitlab-ce
-GITLAB_TOKEN ?= ACCTEST1234567890123
+GITLAB_TOKEN = ACCTEST1234567890123
```
armsnyder, line 45: This would prevent running tests against an external instance.
```suggestion
GITLAB_TOKEN ?= ACCTEST1234567890123
```

armsnyder (changes requested):
>  Also I think a v4 wouldn't necessarily be it

Agree

> (2) that no one is using it a go dependency

That's a good reason, in case we modularize more in the future.

internal/provider/helper_test.go
armsnyder: Thanks!

GNUmakefile
timofurrer: Was more a convenience change for myself, because I sometimes have that variable set for another purpose 🙈 . 
I'll revert this, if it's a use case :)

armsnyder: Oh, gotcha. Maybe we can rename it in this or a future PR then. (It's referenced in the testing section of CONTRIBUTING.md fyi).

armsnyder (approved):

timofurrer:
@armsnyder Are you okay if we merge this early next week?

armsnyder:
@timofurrer Sounds good, as long as it's after the 3.9.1 patch release. I don't know if you saw my two PRs for this branch, in your fork? It's up to you if you want to include those or not. I thought it might make sense to get the changes in the same PR (this one).

timofurrer:
Yes, I'll wait for 3.9.1 - I've marked it as `blocked` in the meantime. 

I haven't seen your PR 🙈  Let me see :)

timofurrer:
@armsnyder I've added the changes from our two PRs - look good to me 🎉

GNUmakefile
armsnyder: Hey!

timofurrer: oups 😖  that wasn't on purpose. Changed it back ;)

armsnyder (approved):

armsnyder (approved):

PR #816, opened 2022-01-28
Implement `archive_on_destroy` attribute for projects. Closes #761

### LINKED ISSUE #761: feature request: archive on delete
Hello, GitHub provider for Terraform supports archive on delete so when a repository is destroyed via TF rather than be deleted it is archived, would it be possible to add this?

timofurrer:
This change implements an `archive_on_destroy` attribute on the
`gitlab_project` resource. It defaults to `false`, if set to `true`, on
a destroy it will *archive* the project instead of deleting it.

This is especially useful in organizations where repositories need to
be kept for regulatory purposes and can't be deleted.

PS: the diff is a mess: I've basically just `if'ed` the `DeleteProject` case depending on the `archive_on_destroy` attribute and called `ArchiveProject` in the new branch :)

### DIFF
--- a/gitlab/resource_gitlab_project_test.go
+++ b/gitlab/resource_gitlab_project_test.go
@@ -383,6 +383,21 @@ func TestAccGitlabProject_initializeWithoutReadme(t *testing.T) {
 	})
 }
 
+func TestAccGitlabProject_archiveOnDestroy(t *testing.T) {
+	rInt := acctest.RandInt()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckGitlabProjectArchivedOnDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGitlabProjectConfigArchiveOnDestroy(rInt),
+			},
+		},
+	})
+}
+
 func TestAccGitlabProject_IssueMergeRequestTemplates(t *testing.T) {
 	var project gitlab.Project
 	rInt := acctest.RandInt()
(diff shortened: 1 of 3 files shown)


gitlab/resource_gitlab_project_test.go
```diff
@@ -383,6 +383,35 @@ func TestAccGitlabProject_initializeWithoutReadme(t *testing.T) {
...
 }
 
+func TestAccGitlabProject_archiveOnDestroy(t *testing.T) {
+	var project gitlab.Project
+	rInt := acctest.RandInt()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckGitlabProjectDestroy,
```
armsnyder, line 393: Doesn't this default `CheckDestroy` fail if the project is archived?

armsnyder: Oh I see, since it was already destroyed in the last step, there is no longer a `gitlab_project` in the state, so this `CheckDestroy` is a no-op.

It would be less messy to write a new `CheckDestroy` instead of using a test step.

armsnyder (approved):
Looks good. I left one optional feedback comment. This differs from the upstream API (normally bad) but I think it's ok in this case because one could look at archiving as being a deletion strategy.

gitlab/resource_gitlab_project_test.go
timofurrer: Yes, good idea :) Just pushed your suggested approach 🎉

timofurrer:
> This differs from the upstream API (normally bad) but I think it's ok in this case because one could look at archiving as being a deletion strategy.

Yes, I agree :)

gitlab/resource_gitlab_project_test.go
```diff
@@ -1343,3 +1379,15 @@ resource "gitlab_project" "foo" {
...
 }
+
+func testAccGitlabProjectConfigArchiveOnDestroy(rInt int) string {
+	return fmt.Sprintf(`
+resource "gitlab_project" "foo" {
+  name = "foo-%d"
+  path = "foo.%d"
+  description = "Terraform acceptance tests"
+  archive_on_destroy = true
+  archived = false
```
armsnyder, line 1390: Hmm in all our other tests we add this config to the `gitlab_project` resource. I'm not entirely sure it is still correct, but for consistency...

```suggestion
  archived = false

  # So that acceptance tests can be run in a gitlab organization with no billing.
  visibility_level = "public"
```

timofurrer: Actually most of the resources in the `resource_gitlab_project_test.go` didn't set that attribute. However, I've pushed a commit on top this branch changing that 🎉

armsnyder:
lgtm 😄

PR #819, opened 2022-01-29
Aligns the *Access Level* situation in this providers with the upstream API. Closes #794

### LINKED ISSUE #794: Streamline access levels across data sources, resources and docs
The current handling of access levels across different data sources and resources and their respective documentation is very inconsistent, as already tried to address in #598.

The documentation of the access levels for the API is here: https://docs.gitlab.com/ee/api/members.html#valid-access-levels

The goal of this issue is to streamline at least these and their docs:

- [ ] group membership data source
- [ ] branch protection resource
- [ ] group share group resource
- [ ] project membership resource
- [ ] project share group resource
- [ ] tag protection resource

Also note this helper here: https://github.com/gitlabhq/terraform-provider-gitlab/blob/8564a07ea91f936b67332a8ef4827d1fa322848d/gitlab/util.go#L238

timofurrer:
The situation is a little more trickier than I've expected.
See my comments in `gitlab/access_level_helpers.go`.

I think that this should be a good enough first improvement and makes
some helpful clarifications in the docs. It also aligns API attributes
in the `gitlab_group_ldap_link` and `gitlab_project_share_group`
resources. The latter even includes a state migration to ensure
compatibility with older states.

Closes #794

### DIFF
--- a/gitlab/access_level_helpers.go
+++ b/gitlab/access_level_helpers.go
@@ -1,0 +1,73 @@
+package gitlab
+
+import (
+	"github.com/xanzy/go-gitlab"
+)
+
+// NOTE:
+// The access level story in the GitLab API is a bit tricky.
+// There are different resources using the same access level names
+// with an identical mapping to int ids. As also defined in the
+// `gitlab.AccessLevelValue` types. However, different endpoints
+// allow all of them or just a subset. There is also endpoints
+// defining an additional `admin` access level, which is nowhere
+// documented and probably not used at all - this provider ignores it.
+// Point being, be careful when using them in a resource or data source
+// and consult the upstream API docs to verify what's possible and keep
+// your fingers crossed it's correct :)
+
+// see the source of truth for `accessLevelNameToValue` and `accessLevelValueToName`
+// here: https://docs.gitlab.com/ee/api/members.html#valid-access-levels
+var validGroupAccessLevelNames = []string{
+	"no one",
+	"minimal",
+	"guest",
+	"reporter",
+	"developer",
+	"maintainer",
+	"owner",
+
+	// Deprecated and should be removed in v4 of this provider
+	"master",
+}
+var validProjectAccessLevelNames = []string{
+	"no one",
+	"minimal",
+	"guest",
+	"reporter",
+	"developer",
+	"maintainer",
+
+	// Deprecated and should be removed in v4 of this provider
+	"master",
+}
+
+// NOTE(TF): the documentation here https://docs.gitlab.com/ee/api/protected_branches.html
+//           mentions an `60 => Admin access` level, but it actually seems to not exist.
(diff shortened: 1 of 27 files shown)


gitlab/access_level_helpers.go
```diff
@@ -0,0 +1,86 @@
...
+	gitlab.NoPermissions:            "no one",
+	gitlab.MinimalAccessPermissions: "minimal",
+	gitlab.GuestPermissions:         "guest",
+	gitlab.ReporterPermissions:      "reporter",
+	gitlab.DeveloperPermissions:     "developer",
+	gitlab.MaintainerPermissions:    "maintainer",
+	gitlab.OwnerPermissions:         "owner",
+}
+
+func validateGroupAccessLevelNames() schema.SchemaValidateDiagFunc {
```
armsnyder, line 76: This is a personal preference, but I don't think we need these `validate*` helper functions. To me it is easier to parse `ValidateDiagFunc: validateValueFunc(validGroupAccessLevelNames)` since I don't need to look up the implementation.

gitlab/resource_gitlab_project_share_group.go
```diff
@@ -53,11 +69,19 @@ func resourceGitlabProjectShareGroupCreate(ctx context.Context, d *schema.Resour
 
 	groupId := d.Get("group_id").(int)
 	projectId := d.Get("project_id").(string)
-	accessLevelId := accessLevelID[d.Get("access_level").(string)]
+
+	var group_access gitlab.AccessLevelValue
```
armsnyder, line 73: nit: camelCase

armsnyder (approved):
I love the Description format strings!

Thanks for aligning the attribute names and writing the migration. Funny how the ldap resource didn't need the migration because it was secretly setting the `group_access` attribute.

The validation here is more restrictive, (e.g. gitlab_branch_protection) but it should be ok because the previous values were not allowed in the API, so these are bug fixes.

Just a couple nitpick comments, otherwise lgtm!

gitlab/access_level_helpers.go
timofurrer: Fine with me :D

gitlab/resource_gitlab_project_share_group.go
timofurrer: Oh yeah, language switches 🤦

armsnyder:
Can you also update the example tf in /examples

armsnyder (changes requested):
Can you also update the example tf in /examples

timofurrer:
@armsnyder I've also pushed a commit on here which removes the `validateValueFunc` completely and uses the equivalents from the SDK. I think we should make use of off the shelf utils whenever we can :)

armsnyder (approved):
lgtm, thanks!

PR #824, opened 2022-01-31
Workflows: Label stale issues and PRs, and issue triage labels

armsnyder:
## Description

Adds workflows for:
- Marking issues and PRs as "stale" and then closing them after 2 weeks
- Adding "needs-triage" and conditionally "bug" labels (based on a regex) for newly opened issues
- Unmarking issues as "stale" or "waiting-response" whenever there is activity
- Removing labels "waiting-response" or "needs-triage" if an issue is closed with those labels

Note that I configured the stale workflow to never close an issue with a "bug", "enhancement", or "keep-open" label.

I tested most of this in my fork, but I have not tested the cron job workflow. It is copied from the action documentation, so it should work.

## PR Checklist

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,44 @@
+# GitHub Action reference: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions
+#
+# Automated handling of stale issues and pull requests.
+#
+# NOTE: When changing "days-before-*" values, please also update the "stale-issue-message" and/or
+# "stale-pr-message" values.
+
+name: stale
+
+on:
+  schedule:
+    #        ┌───────────── minute (0 - 59)
+    #        │ ┌───────────── hour (0 - 23)
+    #        │ │ ┌───────────── day of the month (1 - 31)
+    #        │ │ │ ┌───────────── month (1 - 12 or JAN-DEC)
+    #        │ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
+    #        * * * * *
+    - cron: '8 3 * * *'
+
+permissions:
+  issues: write
+  pull-requests: write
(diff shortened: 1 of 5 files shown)


timofurrer (approved):
Also, very nice 🎉

timofurrer:
What's the behavior for already existing issues and PRs ?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,44 @@
...
+
+on:
+  schedule:
+    #        ┌───────────── minute (0 - 59)
+    #        │ ┌───────────── hour (0 - 23)
+    #        │ │ ┌───────────── day of the month (1 - 31)
+    #        │ │ │ ┌───────────── month (1 - 12 or JAN-DEC)
+    #        │ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
+    #        * * * * *
+    - cron: '8 3 * * *'
```
timofurrer, line 18: that's a very specific time for such a cron job 😛

armsnyder: lol just applying some random jitter

armsnyder:
@timofurrer 

> What's the behavior for already existing issues and PRs ?

The way I configured it, it will sweep all issues and PRs on the first cron run, adding the "stale" label, but not closing anything.

We can optionally provided a min timestamp so it only considers PRs/issues opened after today, but IMO marking existing ones as "stale" is ok.

timofurrer:
@armsnyder sounds good - in any case, we can always reopen :)

armsnyder:
Cool cool, and as a reminder it will not mark issues as `stale` if they have an `enhancement` or `bug` label. It will mark some PRs as `stale` though.

PR #828, opened 2022-01-31
Workflows: Automate applying merge-conflict label

armsnyder:
## Description

Automatically adds and removes the `merge-conflict` label based on whether the PR has merge conflicts.

Action docs: https://github.com/marketplace/actions/label-conflicting-pull-requests

I configured it to comment on the PRs, but we can turn that off if it's too noisy.

Tested in my fork.

## PR Checklist

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/.github/workflows/pr-check-conflicts.yml
+++ b/.github/workflows/pr-check-conflicts.yml
@@ -1,0 +1,28 @@
+# GitHub Action reference: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions
+
+name: pr-check-conflicts
+
+on:
+  # For adding the merge-conflict label.
+  push:
+    branches:
+      - main
+      - master
+  # For removing the merge-conflict label.
+  pull_request_target:
+    types: [synchronize]
+
+permissions:
+  pull-requests: write
+
+jobs:
+  label:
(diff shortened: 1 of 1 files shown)


timofurrer:
I'm not sure how useful that is, at least with what I intended the needs-rebase label to be: To me it's when it is not obvious that the PR needs a rebase, because it would fail as is after a merge. The merge conficts can be spotted easily, so not sure how useful the label is there. I'd be fine with using the needs-rebase manually. If we want a label for conflict I'd use this action with maybe a needs-conflict-resolve label or something.

armsnyder:
Yeah I agree it is useful to have a label for non-obvious needing rebase.

Values I see for this workflow are:
1. Adding the ability to filter PRs and see visually which ones have conflicts on the PR list page.
2. The comment which sets expectations with the contributor that they should rebase.

Maybe it can be a different label like `conflicts` to not collide with the original intention of `needs-rebase`?

Or if you still think there's not enough value we can close.

timofurrer:
> Maybe it can be a different label like merge-conflict to not collide with the original intention of needs-rebase?

Yes, let's do that 👍

timofurrer (approved):

PR #830, opened 2022-01-31
Workflows: Prevent merging PRs with needs-rebase or do-not-merge labels

armsnyder:
## Description

This action fails if either `needs-rebase` or `do-not-merge` labels are present on the PR.

The benefit is we have a way to communicate that a PR should not be merged (`do-not-merge`) that has some enforcement with it.

I don't know how useful this is right now, but we can also expound on it to add other label rules, for example a PR must have one of N different labels before it can be merged. It could be useful for PR triaging purposes.

Tested in my fork.

## PR Checklist

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/.github/workflows/pr-required-labels.yml
+++ b/.github/workflows/pr-required-labels.yml
@@ -1,0 +1,29 @@
+# GitHub Action reference: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions
+#
+# Enforce label requirements on pull requests, to prevent accidental merging.
+
+name: pr-required-labels
+
+on:
+  pull_request_target:
+    types: [opened, reopened, labeled, unlabeled]
+
+permissions:
+  pull-requests: read
+
+jobs:
+  required-labels:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check do-not-merge
+        uses: mheap/github-action-required-labels@v1
+        with:
+          mode: exactly
+          count: 0
+          labels: do-not-merge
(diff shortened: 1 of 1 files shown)


timofurrer:
>  don't know how useful this is right now

Something like that would be very useful even right now, e.g. for https://github.com/gitlabhq/terraform-provider-gitlab/pull/815. 

Let's start with these two labels. Or maybe a third one which would indicate that it should only be merged after the next release. Maybe `wait-upcoming-release` or `merge-after-release` or something like that.

armsnyder:
Heh yep, #815 gave me the idea. I think just the one `do-not-merge` label is ok for now. Juggling lots of different labels with similar meaning can be rough. But yep it's always open for change.

timofurrer:
> Juggling lots of different labels with similar meaning can be rough

Btw. I also think that given the labels meaningful descriptions can help a lot with maintaining them.
Unfortunately, I don't have access to edit them (yet).

armsnyder:
@timofurrer 

> Unfortunately, I don't have access to edit them (yet).

Oh no, I didn't realize. 😮 And I agree, descriptions are valuable. We could use something like https://github.com/marketplace/actions/label-syncer if it's easier to check the descriptions into git. Or you can suggest descriptions and I could add them.

timofurrer (approved):

armsnyder:
That wasn't a problem for very long 😆

timofurrer:
Yes, that's right 💪  😄

PR #764, opened 2021-12-09
Add topic resource support

eliobischof:
Removing a topic resource prints a warning because GitLab doesn't support deleting topics yet. Instead it emptys the topics description.

Also, uploading avatars isn't supported by the go-gitlab library, see  https://github.com/xanzy/go-gitlab/pull/1309#issue-798852929

Closes: #748 

Depends on: https://github.com/xanzy/go-gitlab/pull/1309

Issue #748, opened 2021-11-16
feature request: manage topics via admin API


bufferoverflow:
With the upcoming GitLab release there is an [admin API to manage topics](https://gitlab.com/gitlab-org/gitlab/-/blob/master/doc/api/topics.md), see also https://gitlab.com/gitlab-org/gitlab/-/issues/340920

Would be awesome to become able to manage the topics via the terraform provider.

timofurrer:
I've created another issue in the go-gitlab library to support the API so that it can easily implemented here. see https://github.com/xanzy/go-gitlab/issues/1288

eliobischof:
As the GitLab API doesn't support deleting a topic, implementing a fully functional topic resource isn't possible ATM.

That's what the docs say about resource implementation...

```golang
	// The functions below are the CRUD operations for this resource.
	//
	// The only optional operation is Update. If Update is not implemented,
	// then updates will not be supported for this resource.
	//
	// The ResourceData parameter in the functions below are used to
	// query configuration and changes for the resource as well as to set
	// the ID, computed data, etc.
	//
	// The interface{} parameter is the result of the ConfigureFunc in
	// the provider for this resource. If the provider does not define
	// a ConfigureFunc, this will be nil. This parameter should be used
	// to store API clients, configuration structures, etc.
	//
	// If any errors occur during each of the operation, an error should be
	// returned. If a resource was partially updated, be careful to enable
	// partial state mode for ResourceData and use it accordingly.
	//
	// Exists is a function that is called to check if a resource still
	// exists. If this returns false, then this will affect the diff
	// accordingly. If this function isn't set, it will not be called. You
	// can also signal existence in the Read method by calling d.SetId("")
	// if the Resource is no longer present and should be removed from state.
	// The *ResourceData passed to Exists should _not_ be modified.
	Create CreateFunc
	Read   ReadFunc
	Update UpdateFunc
	Delete DeleteFunc
	Exists ExistsFunc
```

_The only optional operation is Update._

Of course I can pass a delete function that always returns an error saying that deleting a topic is not supported...

bufferoverflow:
@eliobischof for the delete case the option I see at the moment is mapping it to update endpoint and just remove the image and empty the description of the topic. Anyway, first we need https://github.com/xanzy/go-gitlab/issues/1288

eliobischof:
@bufferoverflow I can empty the description but not the avatar, as the update API requires a valid avatar URL

[...]

PR #832, opened 2022-02-01
Add topic resource support, second trial

eliobischof:
## Description

This PR replaces #764. Also see reviews/discussions there. 

## PR Checklist

- [ ] [Examples](/examples) are updated with:
    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/docs/resources/topic.md
+++ b/docs/resources/topic.md
@@ -1,0 +1,30 @@
+# gitlab\_topic
+
+This resource allows you to create and manage topics that are then assignable to projects. Topics are the successors for project tags. Aside from avoiding terminology collisions with Git tags, they are more descriptive and better searchable.
+
+For assigning topics, use the [project](./project.md) resource.
+
+## Example Usage
+
+```hcl
+resource "gitlab_topic" "functional-programming" {
+  name             = "Functional Programming"
+  description      = "In computer science, functional programming is a programming paradigm where programs are constructed by applying and composing functions."
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the approval rule.
+
+* `description` - (Optional) A text describing the topic.
+
+## Import
+
+GitLab topics can also be imported, e.g.
+
+```
+$ terraform import gitlab_topic.functional-programming
+```
--- a/examples/resources/gitlab_topic/import.sh
+++ b/examples/resources/gitlab_topic/import.sh
@@ -1,0 +1,3 @@
+# You can import a topic to terraform state using `terraform import <resource>`.
+# for example:
+terraform terraform import gitlab_topic.functional-programming
--- a/gitlab/provider.go
+++ b/gitlab/provider.go
@@ -108,8 +108,7 @@ func Provider() *schema.Provider {
 			"gitlab_project_freeze_period":      resourceGitlabProjectFreezePeriod(),
 			"gitlab_group_share_group":          resourceGitlabGroupShareGroup(),
 			"gitlab_project_badge":              resourceGitlabProjectBadge(),
-			"gitlab_group_badge":                resourceGitlabGroupBadge(),
-			"gitlab_repository_file":            resourceGitLabRepositoryFile(),
+			"gitlab_topic":                      resourceGitlabTopic(),
 		},
 	}
 
--- a/gitlab/resource_gitlab_topic.go
+++ b/gitlab/resource_gitlab_topic.go
@@ -1,9 +1,9 @@
 package gitlab
 
 import (
-	"fmt"
 	"log"
 	"net/http"
+	"strconv"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/xanzy/go-gitlab"
@@ -44,12 +44,12 @@ func resourceGitlabTopicCreate(d *schema.ResourceData, meta interface{}) error {
 
 	log.Printf("[DEBUG] create gitlab topic %s", *options.Name)
 
-	label, _, err := client.Topics.CreateTopic(options)
+	topic, _, err := client.Topics.CreateTopic(options)
 	if err != nil {
 		return err
 	}
 
-	d.SetId(label.Name)
+	d.SetId(strconv.FormatInt(int64(topic.ID), 10))
 
 	return resourceGitlabTopicRead(d, meta)
 }
@@ -69,7 +69,7 @@ func resourceGitlabTopicRead(d *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
-	d.SetId(fmt.Sprintf("%d", topic.ID))
+	d.SetId(strconv.FormatInt(int64(topic.ID), 10))
 	d.Set("name", topic.Name)
 	d.Set("description", topic.Description)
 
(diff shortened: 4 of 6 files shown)


docs/resources/topic.md
```diff
@@ -0,0 +1,30 @@
+# gitlab\_topic
```
timofurrer, line 1: Please use the `Description` fields in the resource schema definition and run `make reviewable` locally - this will generate that markdown file for you!

eliobischof:
Generating the docs didn't work yet. I'm looking into it.

gitlab/provider.go
```diff
@@ -108,8 +108,7 @@ func Provider() *schema.Provider {
 			"gitlab_project_freeze_period":      resourceGitlabProjectFreezePeriod(),
 			"gitlab_group_share_group":          resourceGitlabGroupShareGroup(),
 			"gitlab_project_badge":              resourceGitlabProjectBadge(),
-			"gitlab_group_badge":                resourceGitlabGroupBadge(),
```
timofurrer, line 111: You probably didn't mean to remove those :)

gitlab/resource_gitlab_topic.go
```diff
@@ -0,0 +1,131 @@
...
+	"log"
+	"strconv"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/xanzy/go-gitlab"
+)
+
+func resourceGitlabTopic() *schema.Resource {
+	return &schema.Resource{
```
timofurrer, line 15: You've had some nice resource description before, can you add it here, please?

timofurrer:
@eliobischof What's the reason for not including the `avatar` field? The API seems to be working correctly now, right? https://github.com/gitlabhq/terraform-provider-gitlab/issues/748#issuecomment-996801092

gitlab/resource_gitlab_topic.go
```diff
@@ -0,0 +1,133 @@
...
+		Description: "This resource allows you to create and manage topics that are then assignable to projects. Topics are the successors for project tags. Aside from avoiding terminology collisions with Git tags, they are more descriptive and better searchable.\n\nFor assigning topics, use the [project](./project.md) resource.",
```
armsnyder, line 16: Could you mention that deleting the resource does not delete the topic, using a [markdown callout](https://www.terraform.io/registry/providers/docs#callouts)

armsnyder (changes requested):

gitlab/resource_gitlab_topic.go
eliobischof: Should I also just print a warning instead of emptying the fields, according to @wwwjon's [comment](https://github.com/gitlabhq/terraform-provider-gitlab/issues/748#issuecomment-990707136)

eliobischof:
> @eliobischof What's the reason for not including the `avatar` field? The API seems to be working correctly now, right? [#748 (comment)](https://github.com/gitlabhq/terraform-provider-gitlab/issues/748#issuecomment-996801092)

@timofurrer uploading an avatar is not supported, see [here](https://github.com/xanzy/go-gitlab/pull/1309#issue-1075604772), so I guess creating an avatar field would not help? @wwwjon mentioned the deletion of an avatar, but if the field isn't really manageable, I can't see the benefit for adding it or how it should be implemented?

timofurrer:
@eliobischof ah I see. Okay, then - let's leave it out for now.

gitlab/resource_gitlab_topic.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+func resourceGitlabTopic() *schema.Resource {
+	return &schema.Resource{
+		Description: `This resource allows you to create and manage topics that are then assignable to projects. Topics are the successors for project tags. Aside from avoiding terminology collisions with Git tags, they are more descriptive and better searchable.  
+
+For assigning topics, use the [project](./project.md) resource.  
+
+
+~> Deleting a resource doesn't delete the corresponding topic as the GitLab API doesn't support deleting topics yet`,
```
timofurrer, line 21: Maybe document what it does instead, too.

wwwjon:
>> @eliobischof What's the reason for not including the avatar field? The API seems to be working correctly now, right? [#748 (comment)](https://github.com/gitlabhq/terraform-provider-gitlab/issues/748#issuecomment-996801092)
> 
> @timofurrer uploading an avatar is not supported, see here, so I guess creating an avatar field would not help? @wwwjon mentioned the deletion of an avatar, but if the field isn't really manageable, I can't see the benefit for adding it or how it should be implemented?

@eliobischof @timofurrer Uploading an avatar is supported, see [here](https://docs.gitlab.com/ee/api/topics.html#upload-a-topic-avatar). This is supported on creating (POST) and updating (PUT) a topic.

gitlab/resource_gitlab_topic.go
armsnyder: Yeah I can see that adding a delete feature in the future would be potentially breaking.

How about we add a new required boolean attribute, default false, called `soft_destroy`? Then in the destroy function you run the behavior you already have if the value is true, and return an error if it is false.

Then in the future if topic deletion is supported in the API we can make the attribute optional.

timofurrer:
@armsnyder @wwwjon Yes, you're probably both right! Let's get this merged asap with or without `avatar` and let's check what `go-gitlab` can do to support it - since it also should be aligned with the upstream API. (@eliobischof I've seen that you've implemented that there, too - so maybe you can check that out another time :) )

It would be nice to get this into 3.9.0! IMHO, only this is preventing a merge: https://github.com/gitlabhq/terraform-provider-gitlab/pull/832#discussion_r797992207

eliobischof:
Yeah, I scoped avatars out in the library because topics would be the first resouce that would support uploading avatars. Like for the terraform plugin it seems to me that this belongs to a dedicated PR over multiple resources.

armsnyder:
It could be the same resource, since there is a field for the avatar url in the response JSON https://docs.gitlab.com/ee/api/topics.html.

I'm imagining it like how [https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_function](aws_lambda_function) handles uploading source code. It has two attributes `filename` and `source_code_hash` to that it can calculate internally whether it needs to upload the file.

examples/resources/gitlab_topic/import.sh
```diff
@@ -0,0 +1,3 @@
+# You can import a topic to terraform state using `terraform import <resource>`.
+# for example:
+terraform terraform import gitlab_topic.functional-programming
```
armsnyder, line 3: The example should include the ID being imported (see other import.sh files)

gitlab/resource_gitlab_topic.go
```diff
@@ -0,0 +1,153 @@
...
+	warning := fmt.Sprintf("[WARN] Not deleting gitlab topic %s as gitlab API doens't support deleting topics", d.Id())
+
+	if softDestroy {
+		warning += ". Instead emptying its description"
+	}
+
+	log.Println(warning)
+
+	if !softDestroy {
+		return nil
```
armsnyder, line 136: This should return an error saying something like "Destroying a topic is not yet supported. You can set soft_destroy=true to suppress this error."

armsnyder (changes requested):

gitlab/resource_gitlab_topic.go
```diff
@@ -0,0 +1,153 @@
...
+	"fmt"
+	"log"
+	"strconv"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/xanzy/go-gitlab"
+)
+
+func resourceGitlabTopic() *schema.Resource {
```
timofurrer, line 14: We've merged a PR which changes how resources are registered. Please look at another resource and use the same pattern. It's a fairly easy change :)

timofurrer (changes requested):

timofurrer:
It's almost ready to merge. @eliobischof are you still working on this or should someone take over?

timofurrer:
@eliobischof thanks for the push, I think a similar mess-up happened as in the first PR ...

PR #834, opened 2022-02-02
Add missing line continuation in example terraform init command

ppejovic:
Fixes missing line continuation in the example `terraform init` command in the GitLab managed state README.


armsnyder (approved):
thanks for the fix!

PR #838, opened 2022-02-03
resource/gitlab_project: correctly handle push rules add and edit

### LINKED ISSUE #836: Creating a project with `push_rule` `member_check` set to `false` leads to an error
Hi there!

When trying to create a project with `push_rule` `member_check` set to `false` the resource creation stops with an error on setting the `push_rule`. 

### Terraform Version
```sh
/ # terraform -v
Terraform v1.1.4
on linux_amd64
```

### Affected Resource(s)
- gitlab_project

```hcl
variable "GITLAB_TOKEN" {
    type = string
}

variable "GITLAB_API" {
    type = string
}

provider "gitlab" {
    token = var.GITLAB_TOKEN
    base_url = var.GITLAB_API
}

terraform {
  backend "local" {
  }
  required_providers {
    gitlab = {
      source = "gitlabhq/gitlab"
      version = "3.8.0"
    }
  }
}

resource "gitlab_project" "terraform-provider-gitlab" {

    name = "terraform-provider-gitlab"
    path = "terraform-provider-gitlab"
    push_rules {
      member_check = "false"
    }
}
```

### Debug Output
https://gist.github.com/janhorstmann/25d4699bc8047f54db442fe1ba6a4154

### Expected Behavior
Project gets created and push rule is set as configured

### Actual Behavior
Resource creation errors on setting the push rule
The project itself is created and manually creating the rule afterwards succeeds:
```
curl --header "Content-Type: application/json" --header "PRIVATE-TOKEN: ${TF_VAR_GITLAB_TOKEN}" --data '{"id": "2392","member_check": "false"}' --request PUT https://XXXXXXXXXXXXXXX/api/v4/projects/2392/push_rule
{"id":2439,"project_id":2392,"created_at":"2022-02-02T16:25:03.329Z","commit_message_regex":"","commit_message_negative_regex":null,"branch_name_regex":"","deny_delete_tag":false,"member_check":false, ...
```
The debug output from above shows no data for the same request via the terraform-provider-gitlab

### Steps to Reproduce
Take above terraform config and run
```sh
export TF_VAR_GITLAB_TOKEN
read -r -s TF_VAR_GITLAB_TOKEN
export TF_VAR_GITLAB_API
read -r TF_VAR_GITLAB_API
export TF_LOG=debug
terraform init
terraform apply
```

timofurrer:
This is a possible approach to fix https://github.com/gitlabhq/terraform-provider-gitlab/issues/836. Even though, I am not sure if that is
really that elegant of a solution ...

The root problem here is that the SDK v2 doesn't support to check if an
attribute value was set in config, state or the plan. During creation of
push rules, we don't have a way to distinguish between the default and
if the user has set the default explicitly. This makes the behavior
working for all cases (AFAIK) (incl. changing the instance default push
rules, no automated test possible yet).

This will be possbile in the new terraform provider framework

### DIFF
--- a/gitlab/resource_gitlab_project.go
+++ b/gitlab/resource_gitlab_project.go
@@ -915,83 +915,88 @@ func resourceGitlabProjectDelete(ctx context.Context, d *schema.ResourceData, me
 func editOrAddPushRules(ctx context.Context, client *gitlab.Client, projectID string, d *schema.ResourceData) error {
 	log.Printf("[DEBUG] Editing push rules for project %q", projectID)
 
-	editOptions := expandEditProjectPushRuleOptions(d)
-	_, _, err := client.Projects.EditProjectPushRule(projectID, editOptions, gitlab.WithContext(ctx))
-	if err == nil {
+	pushRules, _, err := client.Projects.GetProjectPushRules(d.Id(), gitlab.WithContext(ctx))
+	// NOTE: push rules id `0` indicates that there haven't been any push rules set.
+	if err != nil || pushRules.ID == 0 {
+		if addOptions := expandAddProjectPushRuleOptions(d); (gitlab.AddProjectPushRuleOptions{}) != addOptions {
+			log.Printf("[DEBUG] Creating new push rules for project %q", projectID)
+			_, _, err = client.Projects.AddProjectPushRule(projectID, &addOptions, gitlab.WithContext(ctx))
+			if err != nil {
+				return err
+			}
+		} else {
+			log.Printf("[DEBUG] Don't create new push rules for defaults for project %q", projectID)
+		}
+
 		return nil
 	}
 
-	var httpErr *gitlab.ErrorResponse
-	if !errors.As(err, &httpErr) || httpErr.Response.StatusCode != http.StatusNotFound {
-		return err
-	}
-
-	// A 404 could mean that the push rules need to be re-created.
-
-	log.Printf("[DEBUG] Failed to edit push rules for project %q: %v", projectID, err)
-	log.Printf("[DEBUG] Creating new push rules for project %q", projectID)
-
-	addOptions := expandAddProjectPushRuleOptions(d)
-	_, _, err = client.Projects.AddProjectPushRule(projectID, addOptions, gitlab.WithContext(ctx))
-	if err != nil {
-		return err
+	editOptions := expandEditProjectPushRuleOptions(d, pushRules)
+	if (gitlab.EditProjectPushRuleOptions{}) != editOptions {
+		log.Printf("[DEBUG] Editing existing push rules for project %q", projectID)
+		_, _, err = client.Projects.EditProjectPushRule(projectID, &editOptions, gitlab.WithContext(ctx))
+		if err != nil {
+			return err
+		}
+	} else {
+		log.Printf("[DEBUG] Don't edit existing push rules for defaults for project %q", projectID)
 	}
 
 	return nil
 }
 
-func expandEditProjectPushRuleOptions(d *schema.ResourceData) *gitlab.EditProjectPushRuleOptions {
-	options := &gitlab.EditProjectPushRuleOptions{}
+func expandEditProjectPushRuleOptions(d *schema.ResourceData, currentPushRules *gitlab.ProjectPushRules) gitlab.EditProjectPushRuleOptions {
+	options := gitlab.EditProjectPushRuleOptions{}
 
-	if d.HasChange("push_rules.0.author_email_regex") {
+	if d.Get("push_rules.0.author_email_regex") != currentPushRules.AuthorEmailRegex {
 		options.AuthorEmailRegex = gitlab.String(d.Get("push_rules.0.author_email_regex").(string))
 	}
 
-	if d.HasChange("push_rules.0.branch_name_regex") {
+	if d.Get("push_rules.0.branch_name_regex") != currentPushRules.BranchNameRegex {
 		options.BranchNameRegex = gitlab.String(d.Get("push_rules.0.branch_name_regex").(string))
 	}
 
-	if d.HasChange("push_rules.0.commit_message_regex") {
+	if d.Get("push_rules.0.commit_message_regex") != currentPushRules.CommitMessageRegex {
 		options.CommitMessageRegex = gitlab.String(d.Get("push_rules.0.commit_message_regex").(string))
 	}
 
-	if d.HasChange("push_rules.0.commit_message_negative_regex") {
+	if d.Get("push_rules.0.commit_message_negative_regex") != currentPushRules.CommitMessageNegativeRegex {
 		options.CommitMessageNegativeRegex = gitlab.String(d.Get("push_rules.0.commit_message_negative_regex").(string))
 	}
 
-	if d.HasChange("push_rules.0.file_name_regex") {
+	if d.Get("push_rules.0.file_name_regex") != currentPushRules.FileNameRegex {
 		options.FileNameRegex = gitlab.String(d.Get("push_rules.0.file_name_regex").(string))
 	}
 
-	if d.HasChange("push_rules.0.commit_committer_check") {
+	if d.Get("push_rules.0.commit_committer_check") != currentPushRules.CommitCommitterCheck {
 		options.CommitCommitterCheck = gitlab.Bool(d.Get("push_rules.0.commit_committer_check").(bool))
 	}
 
-	if d.HasChange("push_rules.0.deny_delete_tag") {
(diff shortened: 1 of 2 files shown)


gitlab/resource_gitlab_project.go
```diff
@@ -915,129 +915,142 @@ func resourceGitlabProjectDelete(ctx context.Context, d *schema.ResourceData, me
 func editOrAddPushRules(ctx context.Context, client *gitlab.Client, projectID string, d *schema.ResourceData) error {
 	log.Printf("[DEBUG] Editing push rules for project %q", projectID)
 
-	editOptions := expandEditProjectPushRuleOptions(d)
-	_, _, err := client.Projects.EditProjectPushRule(projectID, editOptions, gitlab.WithContext(ctx))
-	if err == nil {
-		return nil
-	}
+	pushRules, _, err := client.Projects.GetProjectPushRules(d.Id(), gitlab.WithContext(ctx))
```
armsnyder, line 918: Missing err check

gitlab/resource_gitlab_project.go
```diff
@@ -915,129 +915,142 @@ func resourceGitlabProjectDelete(ctx context.Context, d *schema.ResourceData, me
...
-	log.Printf("[DEBUG] Creating new push rules for project %q", projectID)
-
-	addOptions := expandAddProjectPushRuleOptions(d)
-	_, _, err = client.Projects.AddProjectPushRule(projectID, addOptions, gitlab.WithContext(ctx))
-	if err != nil {
-		return err
+	editOptions := expandEditProjectPushRuleOptions(d, pushRules)
+	log.Printf("[DEBUG] Editing existing push rules for project %q", projectID)
+	_, _, err = client.Projects.EditProjectPushRule(projectID, editOptions, gitlab.WithContext(ctx))
+	if err == nil {
```
armsnyder, line 937: Weird err check

armsnyder (changes requested):
Thanks for the new test

gitlab/resource_gitlab_project.go
timofurrer: copy&paste weirdness, fixed it 😅

timofurrer:
@armsnyder What do you think about how I solved this? I kinda dislike it and but failed to come up with something better given the constraints of the SDK v2. That exact problem is one of the things addressed in the new framework: https://www.terraform.io/plugin/which-sdk (which we shouldn't switch to just yet, but just saying :D )

gitlab/resource_gitlab_project.go
```diff
@@ -915,129 +915,142 @@ func resourceGitlabProjectDelete(ctx context.Context, d *schema.ResourceData, me
...
 	log.Printf("[DEBUG] Editing push rules for project %q", projectID)
 
-	editOptions := expandEditProjectPushRuleOptions(d)
-	_, _, err := client.Projects.EditProjectPushRule(projectID, editOptions, gitlab.WithContext(ctx))
-	if err == nil {
-		return nil
-	}
+	pushRules, _, err := client.Projects.GetProjectPushRules(d.Id(), gitlab.WithContext(ctx))
+	// NOTE: push rules id `0` indicates that there haven't been any push rules set.
+	if err != nil || is404(err) || pushRules.ID == 0 {
```
armsnyder, line 920: It still looks like you are ignoring any errors from calling GetProjectPushRules, like an HTTP 500 error.

gitlab/resource_gitlab_project.go
```diff
@@ -915,129 +915,142 @@ func resourceGitlabProjectDelete(ctx context.Context, d *schema.ResourceData, me
...
 
 	return nil
 }
 
-func expandEditProjectPushRuleOptions(d *schema.ResourceData) *gitlab.EditProjectPushRuleOptions {
+func expandEditProjectPushRuleOptions(d *schema.ResourceData, currentPushRules *gitlab.ProjectPushRules) *gitlab.EditProjectPushRuleOptions {
 	options := &gitlab.EditProjectPushRuleOptions{}
 
-	if d.HasChange("push_rules.0.author_email_regex") {
+	if d.HasChange("push_rules.0.author_email_regex") || d.Get("push_rules.0.author_email_regex") != currentPushRules.AuthorEmailRegex {
```
armsnyder, line 947: On these `||` conditions, could you remove the left side and just check `d.Get("push_rules.0.author_email_regex") != currentPushRules.AuthorEmailRegex`?

gitlab/resource_gitlab_project.go
```diff
@@ -915,129 +915,142 @@ func resourceGitlabProjectDelete(ctx context.Context, d *schema.ResourceData, me
...
 
-	editOptions := expandEditProjectPushRuleOptions(d)
-	_, _, err := client.Projects.EditProjectPushRule(projectID, editOptions, gitlab.WithContext(ctx))
-	if err == nil {
-		return nil
-	}
+	pushRules, _, err := client.Projects.GetProjectPushRules(d.Id(), gitlab.WithContext(ctx))
+	// NOTE: push rules id `0` indicates that there haven't been any push rules set.
+	if err != nil || is404(err) || pushRules.ID == 0 {
+		if addOptions, needToAdd := expandAddProjectPushRuleOptions(d); needToAdd {
```
armsnyder, line 921: Instead of returning `needToAdd`, you can compare the result `addOptions` with an empty struct in order to detect if it has fields set.

i.e. https://github.com/gitlabhq/terraform-provider-gitlab/blob/289199868e15d647a08f71096a38983ac1c3bacd/gitlab/resource_gitlab_project.go#L635

armsnyder (changes requested):
I didn't look too closely since I was happy with the test. Now that you mention it I see a couple things that could be simplified. 😄

gitlab/resource_gitlab_project.go
armsnyder: Also, does your test still pass if you removed all of the changes you made in this function? In other words is this needed at all?

gitlab/resource_gitlab_project.go
timofurrer: Good question 😄  The thing is that the defaults in the provider (struct) not always match the defaults on the GitLab server, because admins have the possibility to set instance-wide default push rules. And because we can't post an empty body for the push rule endpoint, we have to check if the user specified a value or not (which is not supported by the SDK v2, but only the upcoming framework). This is the only "sane" way I found at the moment to achieve this.

gitlab/resource_gitlab_project.go
timofurrer: Yes, the pass at the moment even without this change, but ONLY if the instance defaults are the same as the provider defaults. See my comments here: https://github.com/gitlabhq/terraform-provider-gitlab/pull/838#discussion_r800220299

As soon as we can manage instance-wide push rules in the provider, we can test this .. may with some effort we could make that API call manually in the tests ...

gitlab/resource_gitlab_project.go
timofurrer: Yes, actually by intention - I thought for an error message to the user it might be more intuitive to have the subsequent `AddProjectPushRule` fail, because that was what the user tried to achieve: set the push rules. 

But, I see your point and can easily change it. What's our general approach here? 🤔

gitlab/resource_gitlab_project.go
armsnyder: I think my comment stands though. `needToAdd` can be calculated because all of the fields of `gitlab.EditProjectOptions` are pointers, so you can easily detect if any of the fields have been set, by comparing it with a zero-value `gitlab.EditProjectOptions`.

gitlab/resource_gitlab_project.go
armsnyder: Sounds good. Since we only call this method after a call to `d.GetOk("push_rules.0")`, that should be ok. Can you hold off merging this until after we can release https://github.com/gitlabhq/terraform-provider-gitlab/milestone/5 ?

gitlab/resource_gitlab_project.go
armsnyder: I see what you mean. That's all right, as long as continuing after an unexpected error won't harm anything. In that case the `is404(err)` check is redundant.

gitlab/resource_gitlab_project.go
timofurrer: Yes, I'll wait for that release for sure!

gitlab/resource_gitlab_project.go
timofurrer: True, ... pointers 😄  I've changed it!

armsnyder (approved):

PR #856, opened 2022-02-09
Support creating project in group without default branch protection

### LINKED ISSUE #759: gitlab_project fails to create in group with no branch protection

timofurrer:
This change includes what is in #760 from @switchboardOp, but adds an acceptance tests.

I also had to fix the `gitlab_group` resource to support groups setting the default branch protection to `0`.

Closes #759 
Closes #760

### DIFF
--- a/internal/provider/resource_gitlab_group.go
+++ b/internal/provider/resource_gitlab_group.go
@@ -208,7 +208,9 @@ func resourceGitlabGroupCreate(ctx context.Context, d *schema.ResourceData, meta
 		options.ParentID = gitlab.Int(v.(int))
 	}
 
-	if v, ok := d.GetOk("default_branch_protection"); ok {
+	// nolint:staticcheck // SA1019 ignore deprecated GetOkExists
+	// lintignore: XR001 // TODO: replace with alternative for GetOkExists
+	if v, ok := d.GetOkExists("default_branch_protection"); ok {
 		options.DefaultBranchProtection = gitlab.Int(v.(int))
 	}
 
--- a/internal/provider/resource_gitlab_project.go
+++ b/internal/provider/resource_gitlab_project.go
@@ -600,8 +600,8 @@ func resourceGitlabProjectCreate(ctx context.Context, d *schema.ResourceData, me
 		}
 
 		log.Printf("[DEBUG] check for protection on old default branch %q for project %q", oldDefaultBranch, d.Id())
-		branch, resp, err := client.ProtectedBranches.GetProtectedBranch(project.ID, oldDefaultBranch, gitlab.WithContext(ctx))
-		if err != nil && resp.StatusCode != http.StatusNotFound {
+		branch, _, err := client.ProtectedBranches.GetProtectedBranch(project.ID, oldDefaultBranch, gitlab.WithContext(ctx))
+		if err != nil && !is404(err) {
 			return diag.Errorf("Failed to check for protected default branch %q for project %q: %v", oldDefaultBranch, d.Id(), err)
 		}
 		if branch == nil {
--- a/internal/provider/resource_gitlab_project.go
+++ b/internal/provider/resource_gitlab_project.go
@@ -599,10 +599,19 @@ func resourceGitlabProjectCreate(ctx context.Context, d *schema.ResourceData, me
 			return diag.Errorf("Failed to protect default branch %q for project %q: %s", newDefaultBranch, d.Id(), err)
 		}
 
-		log.Printf("[DEBUG] unprotect old default branch %q for project %q", oldDefaultBranch, d.Id())
-		_, err = client.ProtectedBranches.UnprotectRepositoryBranches(project.ID, oldDefaultBranch, gitlab.WithContext(ctx))
-		if err != nil {
-			return diag.Errorf("Failed to unprotect undesired default branch %q for project %q: %s", oldDefaultBranch, d.Id(), err)
+		log.Printf("[DEBUG] check for protection on old default branch %q for project %q", oldDefaultBranch, d.Id())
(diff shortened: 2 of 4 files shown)


internal/provider/resource_gitlab_group.go
```diff
@@ -208,7 +208,9 @@ func resourceGitlabGroupCreate(ctx context.Context, d *schema.ResourceData, meta
 		options.ParentID = gitlab.Int(v.(int))
 	}
 
-	if v, ok := d.GetOk("default_branch_protection"); ok {
+	// nolint:staticcheck // SA1019 ignore deprecated GetOkExists
+	// lintignore: XR001 // TODO: replace with alternative for GetOkExists
+	if v, ok := d.GetOkExists("default_branch_protection"); ok {
```
armsnyder, line 213: Not a critique but I struggle with how we should handle default values in the provider. If we were adding this as a new resource, would we use `d.Get` here, in order to allow a zero-value, and set a default value for `default_branch_protection` in the schema? Or use `d.GetOkExists`? Regardless I agree with your handling here for compatibility reasons.

internal/provider/resource_gitlab_project.go
```diff
@@ -599,10 +599,19 @@ func resourceGitlabProjectCreate(ctx context.Context, d *schema.ResourceData, me
 			return diag.Errorf("Failed to protect default branch %q for project %q: %s", newDefaultBranch, d.Id(), err)
 		}
 
-		log.Printf("[DEBUG] unprotect old default branch %q for project %q", oldDefaultBranch, d.Id())
-		_, err = client.ProtectedBranches.UnprotectRepositoryBranches(project.ID, oldDefaultBranch, gitlab.WithContext(ctx))
-		if err != nil {
-			return diag.Errorf("Failed to unprotect undesired default branch %q for project %q: %s", oldDefaultBranch, d.Id(), err)
+		log.Printf("[DEBUG] check for protection on old default branch %q for project %q", oldDefaultBranch, d.Id())
+		branch, resp, err := client.ProtectedBranches.GetProtectedBranch(project.ID, oldDefaultBranch, gitlab.WithContext(ctx))
+		if err != nil && resp.StatusCode != http.StatusNotFound {
```
armsnyder, line 604: Could you use `is404` here, since resp could be nil and panic

armsnyder (changes requested):

internal/provider/resource_gitlab_group.go
armsnyder: Right, `GetOk()` doesn't work on zero-values. I was more just seeing if you would help me touch grass here. Since `GetOkExists()` is apparently discouraged, if we were starting over on `gitlab_group` from scratch, would you still use it? Or would you use `Get()` and set an opinionated default in the provider?

armsnyder (approved):

internal/provider/resource_gitlab_group.go
timofurrer: hmmm 🤔 

> Or would you use Get() and set an opinionated default in the provider?

So, just to make sure I understand correctly, this will only work if this "opinionated default in the provider" is set to something we'd always know it's the "optional default" and the user would never set this in the config themselves, right?
And if it's set to that specific default, we'd not send the field.
The problem is that `Get()` will always return something and we'd never know if we have to include it in the request or not.

I think that even when writing the resource from scratch I'd still use `GetOkExists()` as long as it's working properly, even if discouraged ... I don't see another way to implement this functionality at the moment ..

armsnyder: I'm saying we could use `Get()` and always send the field. And since the API docs say that the default is `2`, we could make that our default also.

Anyway this is hypothetical so I'll resolve. 😄

timofurrer: oh, shoot you are right 🤦  I somehow had in mind that it would use some default set on instance-level which we wouldn't know in the provider ... that seems to be wrong though. Must have confused it with another resource 🤔 

In that case, yes, we could use `Get()` ..

timofurrer: oh, that's where my confusion originated: 

image

I'm not sure if `2` really is the default or just what the default of a pristine GitLab instance is 🤔

armsnyder: I guess I'm not sure if we care if the GitLab instance is configured with a different default. I'll probably research some other providers when I get a chance.

timofurrer: hmmm ... I think here it's not too crucial, but I think we (or someone else) at some point will care 🙃

switchboardOp:
Thanks for following up with the tests @timofurrer! I got busy with the day job.

PR #859, opened 2022-02-09
Remove over-engineered testdata JSON file and just create the struct in the test

timofurrer:
Let's keep things simple :)


armsnyder (approved):
I like simple

PR #862, opened 2022-02-11
resolves #800: Add WaitForState support for resource_gitlab_project branch protection

### LINKED ISSUE #800: Flaky test TestAccGitlabProject_initializeWithReadme
The `TestAccGitlabProject_initializeWithReadme` test has been failing on and off.

```
=== RUN   TestAccGitlabProject_initializeWithReadme
    resource_gitlab_project_test.go:328: Step 1/1 error: Check failed: Check 2/3 error: expected default branch to be protected
```

https://github.com/gitlabhq/terraform-provider-gitlab/runs/4928564294

PatrickRice-KSC:
This fixes an issue where branch protection ran in the background when creating a project, causing the default branch to occasionally not be protected.

## Description

Closes #800 
There was a race condition apparent in project creation where occasionally the branch protection would return `false` even though the branch had been protected asynchronously. This could cause issues if downstream resources used that output value. This also caused several tests that checked the branch protection status to be flakey and seemingly appear to fail rendomly when they were, in fact, failing properly.

## PR Checklist

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/internal/provider/resource_gitlab_project.go
+++ b/internal/provider/resource_gitlab_project.go
@@ -2,10 +2,8 @@ package provider
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"log"
-	"net/http"
 	"strconv"
 	"time"
 
@@ -565,12 +563,11 @@ func resourceGitlabProjectCreate(ctx context.Context, d *schema.ResourceData, me
 
 	if _, ok := d.GetOk("push_rules"); ok {
 		err := editOrAddPushRules(ctx, client, d.Id(), d)
-		var httpError *gitlab.ErrorResponse
-		if errors.As(err, &httpError) && httpError.Response.StatusCode == http.StatusNotFound {
-			log.Printf("[DEBUG] Failed to edit push rules for project %q: %v", d.Id(), err)
-			return diag.Errorf("Project push rules are not supported in your version of GitLab")
-		}
 		if err != nil {
+			if is404(err) {
+				log.Printf("[DEBUG] Failed to edit push rules for project %q: %v", d.Id(), err)
+				return diag.Errorf("Project push rules are not supported in your version of GitLab")
+			}
 			return diag.Errorf("Failed to edit push rules for project %q: %s", d.Id(), err)
 		}
 	}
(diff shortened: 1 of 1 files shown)


PatrickRice-KSC:
OK, well this is embarrassing given I'd run those tests locally. Investigating.

PatrickRice-KSC:
OK, note to self. Don't forget when running make testacc locally to remove the limiter you put on it to only run your specific tests. Whoops.

The issue here was that if you don't set `default_branch` or `initialize_with_readme`, then the default branch obviously throws a 404 when you attempt to query it. Similar to other situations where this happens, I now catch the 404 status and exit the WaitForState loop.

I'm oddly still getting one failing test with this change, related to a data source for an archived project not returning an empty state. I wouldn't have changed that behavior I'd think, but I'm going to go ahead and commit the changes for tonight and keep looking at it tomorrow.

internal/provider/resource_gitlab_project.go
```diff
@@ -620,6 +621,35 @@ func resourceGitlabProjectCreate(ctx context.Context, d *schema.ResourceData, me
...
+	// before we continue. Note this check should only be required when there is a custom default branch set
+	// See issue 800: https://github.com/gitlabhq/terraform-provider-gitlab/issues/800
+	stateConf := &resource.StateChangeConf{
+		Pending: []string{"false"},
+		Target:  []string{"true"},
+		Timeout: 2 * time.Minute, //The async action usually completes very quickly, within seconds. Don't wait too long.
+		Refresh: func() (interface{}, string, error) {
+			branch, response, err := client.Branches.GetBranch(project.ID, project.DefaultBranch, gitlab.WithContext(ctx))
+			if err != nil {
+				if response.StatusCode == 404 {
```
armsnyder, line 634: `response` may be nil here. Prefer to use the `is404(err)` helper function.

armsnyder (dismissed):

internal/provider/resource_gitlab_project.go
PatrickRice-KSC: There's one other reference using similar syntax as this one in the project resource, I'll go ahead and fix both of them tomorrow so we're consistent!

armsnyder: Thanks!

timofurrer:
@PatrickRice-KSC thanks for the deep-dive into what caused this and thanks for the fix! Much appreciated 🎉

PR #864, opened 2022-02-11
Draft next release changelog

timofurrer:
This is just a start and we can continue with it before the actual release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,32 @@
+## 3.10.0 (Unreleased)
+
+FEATURES:
+
+* resource/gitlab_project_hook: added `releases_events` attribute (#773)
+* resource/gitlab_user: added `state` attribute to allow blocking users (#762)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,32 @@
...
+ENHANCEMENTS:
+
+* docs: better provider introduction (#852)
+
+NOTES:
+
+* provider: aligned project structure to terraform provider scaffolding project (#815)
+* repository: changed default branch from `master` to `main` (#852)
+* repository: introduction of multiple GitHub workflows to improve developer experience
+
```
armsnyder, line 29: We only need to include items in the changelog which affect users of the provider. Otherwise it can create confusion. I think all of these can be cut. Other examples of things we can omit would be testing changes and refactors (usually).

```suggestion
```

armsnyder (changes requested):

CHANGELOG.md
timofurrer: Okay, sounds good :)

armsnyder (approved):

Issue #861, opened 2022-02-10
gitlab_group_variable unable to update value of previously applied key


Liberis:
### Terraform Version
Terraform v1.1.5

### Affected Resource(s)
resource "gitlab_group_variable"

### Terraform Configuration Files
```hcl
resource "gitlab_group_variable" "group_variable" {
  for_each          = local.gitlab_variables
  group             = var.group_id
  key               = each.key
  value             = each.value
  protected         = true
  masked            = false
  environment_scope = var.env_value
}
```
### Expected Behavior
What should have happened?
Ability to update gitlab group variable with the same name.
### Actual Behavior
What actually happened?
 `terraform apply`  failed with the following error:
Error: PUT https://gitlab.com/api/v4/groups/*******/variables/var_name: 400 {message: {key: [(var_name) has already been taken]}}
│ 
│   with gitlab_group_variable.group_variable["group_var"],
│   on main.tf line 32, in resource "gitlab_group_variable" "group_variable":
│   32: resource "gitlab_group_variable" "group_variable" {
|
### Steps to Reproduce
1. Create a gitlab group variable with a fixed name.
2. `terraform apply` to create the variable on gitlab group.
3.  Change the variable value to something else.
4. Reapply

timofurrer:
Okay, I'll dig into that later ... Out of curiosity: does it also happen
without the environment scoping?

Liberis:
It seems that if the environment scope is specified gitlab api cannot update the variable.

Also replicated using the api :
curl --request POST --header "PRIVATE-TOKEN: <your_access_token>" \
     "https://gitlab.example.com/api/v4/groups/1/variables/NEW_VARIABLE" --form "value=value" --form "environment-scope=env"
     
 curl --request PUT --header "PRIVATE-TOKEN: <your_access_token>" \
     "https://gitlab.example.com/api/v4/groups/1/variables/NEW_VARIABLE" --form "value=updated_value" --form "environment-scope=env"

timofurrer:
@Liberis I'm still unable to reproduce this. What responses do you get from the API calls you posted.

Creating a simple non-protected / non-masked variable with an environment scope:

```
curl -X POST http://localhost:8080/api/v4/groups/108/variables --header "PRIVATE-TOKEN: ACCTEST1234567890123" --form "key=some_var" --form "value=some_value" --form "environment_scope=env" | jq
{
  "variable_type": "env_var",
  "key": "some_var",
[...]

PR #866, opened 2022-02-11
Add failing group variable update test to test issue #861

timofurrer:
It's currently sporadically failing, because of the upstream bug here: https://gitlab.com/gitlab-org/gitlab/-/issues/333296

I've also refactored the `withEnvironmentScopeFilter` function into a separate file.


armsnyder (approved):
lgtm, ack the new disabled test

PR #871, opened 2022-02-14
Add topic resource support, third trial

eliobischof:
## Description

Cherry picked the commits from #832 that has a messed up commit history. 

## PR Checklist

    - Create the resource
    - Update the attributes
    - Import the resource


timofurrer:
Thanks @eliobischof for creating yet another PR for this! 🎉 

I'm still super unhappy about the tradeoffs we had to make here (absolutely not your fault @eliobischof :D )- but for the lack of a better solution I think we can merge this.

I think I'll follow up with some changes in go-gitlab to support removing the avatar (since https://docs.gitlab.com/ee/api/topics.html#remove-a-topic-avatar) is available and also include the `total_projects_count` and `avatar_url` as `Computed` attributes. But it shouldn't block that PR from finally going through :)

timofurrer (approved):

examples/resources/gitlab_topic/import.sh
```diff
@@ -0,0 +1,4 @@
+# You can import a topic to terraform state using `terraform import <resource> <id>`.
+# The `id` must be an integer for the id of the topic you want to import,
+# for example:
+terraform terraform import gitlab_topic.functional-programming 1
```
armsnyder, line 4: Let's use underscore in name. Also double terraform typo.
```suggestion
terraform import gitlab_topic.functional_programming 1
```

examples/resources/gitlab_topic/resource.tf
```diff
@@ -0,0 +1,4 @@
+resource "gitlab_topic" "functional-programming" {
```
armsnyder, line 1: Style - underscores in names.

```suggestion
resource "gitlab_topic" "functional_programming" {
```

armsnyder (dismissed):
Looks good! Just some typos and style things.

timofurrer:
I took the liberty and applied @armsnyder suggestions. Once the checks pass I'll merge!

Thanks @eliobischof for your effort ❤️  much appreciated 🎉

PR #873, opened 2022-02-15
Fix gitlab_group_ldap_link panic on group_access attribute and re-enable tests

timofurrer:
Refs: #771 

@armsnyder do you know anything about that `force` flag? And how it's supposed to work?
The test currently is failing (it was never executed, so we didn't notice) and it's apparent why, but I don't see what the intention was and how to make it "work" again :D


armsnyder (approved):
LGTM!

(We discussed in DMs) The failing test was not correctly testing the `force` attribute, so we're deleting it.

Issue #874, opened 2022-02-17
Transfering or renaming a project doesn't compute attributes


QuingKhaos:
### GitLab Provider version

3.9.1

### GitLab version

GitLab.com Enterprise Edition 14.8.0-pre

### Terraform version

v1.1.6

### Relevant Terraform Configuration

First run create a project:

```hcl
resource "gitlab_group" "foo" {
  name = "foogroup"
  path = "foogroup"
  visibility_level = "public"
}

resource "gitlab_project" "foo" {
  name = "foo"
  description = "Terraform test"
  namespace_id = gitlab_group.foo.id

  # So that acceptance tests can be run in a gitlab organization
  # with no billing
  visibility_level = "public"
  build_coverage_regex = "foo"
}

resource "gitlab_project_variable" "foo" {
  project = gitlab_project.foo.id

  key = "FOO"
  value = gitlab_project.foo.path_with_namespace
}
```

Second run to transfer it:

```hcl
resource "gitlab_group" "foo" {
  name = "foogroup"
  path = "foogroup"
  visibility_level = "public"
}

resource "gitlab_group" "foo2" {
  name = "foo2group"
  path = "foo2group"
  visibility_level = "public"
}

resource "gitlab_project" "foo" {
  name = "foo"
  description = "Terraform test"
  namespace_id = gitlab_group.foo2.id

  # So that acceptance tests can be run in a gitlab organization
  # with no billing
  visibility_level = "public"
  build_coverage_regex = "foo"
}

resource "gitlab_project_variable" "foo" {
  project = gitlab_project.foo.id

  key = "FOO"
  value = gitlab_project.foo.path_with_namespace
}
```

### Relevant log output

`gitlab_project_variable.foo` is not included in the updated plan:

```shell        
        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
          + create
          ~ update in-place
        
        Terraform will perform the following actions:
        
          # gitlab_group.foo2 will be created
          + resource "gitlab_group" "foo2" {
              + auto_devops_enabled               = false
              + default_branch_protection         = 2
              + emails_disabled                   = false
              + full_name                         = (known after apply)
              + full_path                         = (known after apply)
              + id                                = (known after apply)
              + lfs_enabled                       = true
              + mentions_disabled                 = false
              + name                              = "foo2group"
              + parent_id                         = 0
              + path                              = "foo2group"
              + project_creation_level            = "maintainer"
[...]

PR #875, opened 2022-02-17
[gitlab_project] re-compute project attributes if path or namespace changed

QuingKhaos:
## Description

All project attributes which are affected by a namespace or path change are re-computed if changed and dependent resources on those attributes will be updated in the same run.

Closes: #874

## PR Checklist

- [x] **(not applicable)** Resource attributes match 1:1 the names and structure of the API resource in [the GitLab API documentation](https://docs.gitlab.com/ee/api/).
- [x] **(not applicable)** [Examples](/examples) are updated with:
- [x] **(not applicable)** New resources have at minimum a basic test with three steps:
    - Create the resource
    - Update the attributes
    - Import the resource
- [x] Fix is covered via tests 

### DIFF
--- a/internal/provider/resource_gitlab_project.go
+++ b/internal/provider/resource_gitlab_project.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
@@ -388,6 +389,12 @@ var _ = registerResource("gitlab_project", func() *schema.Resource {
 			StateContext: schema.ImportStatePassthroughContext,
 		},
 		Schema: resourceGitLabProjectSchema,
+		CustomizeDiff: customdiff.All(
+			customdiff.ComputedIf("path_with_namespace", namespaceOrPathChanged),
+			customdiff.ComputedIf("ssh_url_to_repo", namespaceOrPathChanged),
+			customdiff.ComputedIf("http_url_to_repo", namespaceOrPathChanged),
+			customdiff.ComputedIf("web_url", namespaceOrPathChanged),
+		),
 	}
 })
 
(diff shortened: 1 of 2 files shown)


internal/provider/resource_gitlab_project.go
```diff
@@ -1137,3 +1144,7 @@ func flattenProjectPushRules(pushRules *gitlab.ProjectPushRules) (values []map[s
 		},
 	}
 }
+
+func namespaceOrPathChanged(ctx context.Context, d *schema.ResourceDiff, meta interface{}) bool {
+	return d.HasChange("namespace_id") || d.HasChange("path")
```
timofurrer, line 1149: Isn't the `name` argument itself also affected by the problem at hand here?

QuingKhaos: Hmm, on the first thought I would say no, as the resource doesn't return a `name_with_namespace` attribute which is directly affected by a change of `name`.

But you're right, as `name` is the only required attribute and not `path` which can be derived from the `name`. I always explicitly set the `path` so it didn't come up on the tests..

QuingKhaos: Meeeh, I can't find a way to cover that case if `path` isn't set and to trigger on the `name`. I don't just want to trigger on `name` generally as it may not change the depending attributes always. But a change on that e.g. may trigger recreating a dependent project on that `import_url`.

QuingKhaos: @timofurrer Meeeh, I can't find a way to cover that case if `path` isn't set and to trigger on the `name`. I don't just want to trigger on `name` generally as it may not always change the depending attributes. But a change on that e.g. may trigger recreating a dependent project on that `import_url`.

timofurrer: > Yes, you are right - it's only during creation, when path isn't set, right? Or did you uncover another situation?

... for which it actually isn't needed anyways 🤔

QuingKhaos: Right, `CustomizeDiff` is applicable to updates not creation.

QuingKhaos: > Yes, you are right - it's only during creation, when `path` isn't set, right? Or did you uncover another situation?

Ah, yes. `name` is only used as `path` during creation, when no `path` is send to the API. Afterwards `name` changes only with an unset `path` don't trigger changes in the repository path.

timofurrer (approved):
LGTM 🎉

Issue #758, opened 2021-11-30
allow for override for variable in gitlab_project_variable to be non-sensitive


antonym:
Currently the variables in gitlab_project_variable are sensitive by default and cannot be overridden. There are some variables that I'd like to view the output for when doing a plan. 

```
Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
  ~ update in-place
Terraform will perform the following actions:
  # gitlab_project_variable.terraform_version-project will be updated in-place
  ~ resource "gitlab_project_variable" "terraform_version-project" {
        id                = "12345:TERRAFORM_VERSION:*"
      ~ value             = (sensitive value)
        # (6 unchanged attributes hidden)
    }
Plan: 0 to add, 1 to change, 0 to destroy.
```

In this case, we're setting a default TERRAFORM_VERSION and would like to be able to review the variable that is going to be rolled out when doing the plan. Would it be possible to allow a sensitive override to false for these type of scenarios? I would imagine there are a lot other valid scenarios for this feature as well.

We are running Terraform 1.0.11 with version 3.7.0 of terraform-provider-gitlab.

armsnyder:
In theory this could be supported with a new attribute `value_plaintext` which would be mutually exclusive (`ExactlyOneOf`) with `value`, and would not have the sensitive attribute. However I have not found precedence for a Terraform plugin having this functionality, and it would add considerable tech debt managing the field, which is not a part of the upstream GitLab API, for all the variable resources (group, project, instance -level).

QuingKhaos:
IMHO I wouldn't expect the provider to set the `value` field as sensitive, unless the value supplied itself is a sensitive value within terraform itself. For example from an input/output declared as sensitive, or another resource returns the value as sensitive, or declared sensitive with the [`sensitive()`](https://www.terraform.io/language/functions/sensitive) function.

As an CI/CD variable is very generic it shouldn't make any assumptions about the value sensitivity of the value. I understand the thoughts about doing so, although.

Even if `masked = true` is set, it may sound logical to set the value as sensitive too. On first I would agree, but on second thought this should still be left at the discretion of terraform (referring to a sensitive value) and the developer using the resource (using `sensitive()`)

[...]

PR #876, opened 2022-02-17
CI/CD variable values should be non-sensitive by default

QuingKhaos:
## Description

Removes the sensitive setting from instance/group/project level CI/CD variables. It should be at the discretion of Terraform and the infra developer to mark values as sensitive.

In a very big gitops project of us all sensitive values are still marked as sensitive in the state as they are either input variables set as sensitive or derived from sensitive values like the `runner_token`.

Closes: #758

## PR Checklist

- [x] **(not applicable)** Resource attributes match 1:1 the names and structure of the API resource in [the GitLab API documentation](https://docs.gitlab.com/ee/api/).
- [x] **(not applicable)** [Examples](/examples) are updated with:
- [x] Documentation re-generated.
- [x] **(not applicable)** New resources have at minimum a basic test with three steps:
    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/docs/resources/group_variable.md
+++ b/docs/resources/group_variable.md
@@ -34,7 +34,7 @@ resource "gitlab_group_variable" "example" {
 
 - **group** (String) The name or id of the group.
 - **key** (String) The name of the variable.
-- **value** (String, Sensitive) The value of the variable.
+- **value** (String) The value of the variable.
 
 ### Optional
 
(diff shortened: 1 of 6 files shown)


armsnyder:
Thanks for making the PR! The code looks good. I'm just going to mark this as do-not-merge until we resolve the conversation in #758. 😉

PR #877, opened 2022-02-19
Add branch protection support for allow_force_push

### LINKED ISSUE #709: [Feature Request] Add support for allow_force_push on branch protection
`gitlab_branch_protection` does not support setting `allow_force_push` right now, which causes issues when used to create a mirror from an other repo like a remote github or gitlab for example.

This wasn't an issue before we updated our local environment to gitlab 14.x since that feature wasn't there.

beekeep:
Issue 709 requested the `allow_force_push` option to be added to the branch protection resource.  These changes add that feature, along with the appropriate tests.  The API does not support updating the allow force push configuration, so any changes will force a new resource to be created.

## Description

Closes #709 .  Adds `allow_force_push` attribute to branch protection resource.

## PR Checklist

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/internal/provider/resource_gitlab_branch_protection.go
+++ b/internal/provider/resource_gitlab_branch_protection.go
@@ -78,6 +78,13 @@ var _ = registerResource("gitlab_branch_protection", func() *schema.Resource {
 				Required:         true,
 				ForceNew:         true,
 			},
+			"allow_force_push": {
+				Description: "Can be set to true to allow users with push access to force push.",
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				ForceNew:    true,
+			},
 			"allowed_to_push":  schemaAllowedTo(),
 			"allowed_to_merge": schemaAllowedTo(),
 			"code_owner_approval_required": {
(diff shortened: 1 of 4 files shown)


beekeep:
@armsnyder , @timofurrer, @nagyv  This is ready for your review.  Let me know if there is anything I need to change.  Thanks!

internal/provider/resource_gitlab_branch_protection.go
```diff
@@ -174,6 +183,10 @@ func resourceGitlabBranchProtectionRead(ctx context.Context, d *schema.ResourceD
 		}
 	}
 
+	if err := d.Set("allow_force_push", pb.AllowForcePush); err != nil {
```
timofurrer, line 186: NIT: we usually don't implement explicit error checking when setting primitive types (they are caught by the compiler I guess) ... but doesn't hurt either :)

timofurrer (approved):
Thanks for the change! Looks good to me!

Comparing with the [upstream API](https://docs.gitlab.com/ee/api/protected_branches.html#protect-repository-branches) I've noticed that we are also missing the `allowed_to_unprotect` field. I've created #878 - in case you want to implement this in a separate PR 🎉

internal/provider/resource_gitlab_branch_protection.go
beekeep: I wasn't sure the right approach.  I saw it being done with `code_owner_approval_required` so figured it shouldn't hurt.  If we don't want it done, I can exclude that in the future.

beekeep:
@timofurrer I noticed that attribute missing as well, but figured I would keep this PR to just the one attribute.  I will get another PR ready to add the `allowed_to_unprotect` attribute.

timofurrer (approved):

PR #881, opened 2022-02-21
Fix incompatible assignments for ResourceData.Set() reported by tfproviderlint. Closes #436

### LINKED ISSUE #436: Fix R004: ResourceData.Set() incompatible value type
Check: [R004](https://github.com/bflad/tfproviderlint/tree/master/passes/R004)

```
./gitlab/data_source_gitlab_user.go:198:22: R004: ResourceData.Set() incompatible value type: *time.Time
./gitlab/resource_gitlab_group_membership.go:154:22: R004: ResourceData.Set() incompatible value type: *github.com/xanzy/go-gitlab.ISOTime
./gitlab/resource_gitlab_service_slack.go:148:40: R004: ResourceData.Set() incompatible value type: func(b []byte) error
./gitlab/resource_gitlab_service_slack.go:149:38: R004: ResourceData.Set() incompatible value type: func(b []byte) error
```

We should fix these reported issues, and then remove this check from the list of ignored checks in the gitlabproviderlint make target.

To reproduce, run `go run ./gitlabproviderlint ./gitlab` (pending #432)

timofurrer:
I have to admit though that I kinda like the typed approach which was used there.
However, in favor of having compatible type assignments and consistency it probably makes sense to merge this.

### DIFF
--- a/internal/provider/resource_gitlab_branch_protection.go
+++ b/internal/provider/resource_gitlab_branch_protection.go
@@ -169,30 +169,30 @@ func resourceGitlabBranchProtectionRead(ctx context.Context, d *schema.ResourceD
 	d.Set("project", project)
 	d.Set("branch", pb.Name)
 
-	pushAccessLevels := convertAllowedAccessLevelsToBranchAccessDescriptions(pb.PushAccessLevels)
-	if len(pushAccessLevels) > 0 {
-		if err := d.Set("push_access_level", pushAccessLevels[0].AccessLevel); err != nil {
+	if pushAccessLevel, err := firstValidAccessLevel(pb.PushAccessLevels); err == nil {
+		if err := d.Set("push_access_level", accessLevelValueToName[*pushAccessLevel]); err != nil {
 			return diag.Errorf("error setting push_access_level: %v", err)
 		}
+	} else {
+		return diag.Errorf("unable to get push_access_level: %v", err)
 	}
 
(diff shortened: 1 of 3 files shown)


internal/provider/resource_gitlab_branch_protection.go
```diff
@@ -169,30 +169,30 @@ func resourceGitlabBranchProtectionRead(ctx context.Context, d *schema.ResourceD
 	d.Set("project", project)
 	d.Set("branch", pb.Name)
 
-	pushAccessLevels := convertAllowedAccessLevelsToBranchAccessDescriptions(pb.PushAccessLevels)
-	if len(pushAccessLevels) > 0 {
-		if err := d.Set("push_access_level", pushAccessLevels[0].AccessLevel); err != nil {
+	if pushAccessLevel, err := firstValidAccessLevel(pb.PushAccessLevels); err == nil {
```
armsnyder, line 172: There's a change in behavior here. Previously if no access level was found, it would not set the attribute. Now, due to the `flattenNonZeroBranchAccessDescriptions` change, it is returning a "no valid access level found" error.

I think this is ok because `push_access_level` and `merge_access_level` are required attributes. Feel free to resolve if you agree.

armsnyder (approved):

internal/provider/resource_gitlab_branch_protection.go
timofurrer: Yes, that is my assumption, too.

PR #889, opened 2022-02-23
Finialize changelog for v3.10.0 release

timofurrer:
Changelog for v3.10.0

I've also kept the wording and formatting to what's proposed in https://www.terraform.io/plugin/sdkv2/best-practices/versioning ;)


armsnyder (approved):
lgtm. I like the consistent formatting; makes it much easier to review, thanks!

btw at some point we should fix up the issue links in the changelog. They should be markdown links; referencing issues by number doesn't work in markdown, only in issue comments.

timofurrer:
@armsnyder yes, you are right. Let me do that real quick. Regex where invented for something, right? 😝

CHANGELOG.md
```diff
@@ -1,21 +1,33 @@
-## 3.10.0 (Unreleased)
+## 3.10.0 (2022-02-22)
```
armsnyder, line 1:
```suggestion
## 3.10.0 (2022-02-23)
```

armsnyder (changes requested):

armsnyder (approved):

Issue #853, opened 2022-02-08
Duplicated code between resources and data sources


armsnyder:
For the data sources that we have, the code is almost entirely duplicated from the corresponding resource. This creates problems:
1. They become out of sync (For example, someone adds an attribute to a resource without updating the data source).
2. Managing a data source for each resource is a large enough burden that we usually do not do it.

This is especially egregious for large resources like `gitlab_project` with many fields that are difficult to audit for consistency across the resource and data source.

I'm wondering if we can generate the data source based on the resource? The "read" function is always the same across the two. The only differences between a data source and resource are:
1. Data source fields are always computed
2. The required fields can be different (for example a `gitlab_user` can be looked up by its ID, username, or email)

Possible strategies:
https://github.com/hashicorp/terraform-provider-google/blob/1a72f93a8dcf6f1e59d5f25aefcb6d794a116bf5/google/datasource_helpers.go#L13

timofurrer:
We've worked a little bit towards this with the `datasourceSchemaFromResourceSchema`, `constructSchema`, `attributeNamesFromSchema`, `setStateMapInResourceData` helper functions.

While there is still a lot of duplication and manual processes I think the major issue is solved - new data sources are usually only a couple of lines. 

I'm going to close this.

Issue #746, opened 2021-11-15
Add `gitlab_project_issue` resource


mkjmdski:
Hi :)

It would be great to implement `gitlab_project_issue` or `gitlab_issue` resource. The use case is following:

we are using terraform to provision projects from scratch. As media company we've got kind of the same set of issues that are required at the beginnig of the project. We currently manage this manually or with some extra scripts. Would be good to keep terraform as single source of truth and provision standard issues with terraform too.

armsnyder:
API docs: https://docs.gitlab.com/ee/api/issues.html

armsnyder:
@timofurrer I changed the milestone, since I assume you kept this open to use it for further improvements.

timofurrer:
@armsnyder not sure actually, we can close this since the initial request is solved and it's possible to do what was asked for. 

IMHO, we can close it ...

armsnyder:
Sounds good!

Closed via #891

Released in v3.12.0 of the provider

PR #891, opened 2022-02-23
Add initial Gitlab Project Issue resource and data source support

timofurrer:
This is an initial iteration of the:

* `gitlab_project_issue` resource
* `gitlab_project_issue` data source
* `gitlab_project_issues` data source

It mainly covers the most simple CRUD functions.

Additional issue API endpoints like subscriptions, todos, time
estimation, merge request and issue relations, ... are NOT yet
supported.

I believe that the state as-is in this change set already provides a
great set of value for potentially interested users.

I imagine that the main use case for this is to bootstrap a project with
some getting started / welcome issues and not to track time etc.

Refs #746

@armsnyder this change set also incorporates some very basic approach to reduce duplication between resources and data sources as mentioned in #853. It's by no means perfect, but I think it already reduces some maintenance burden :)


internal/provider/schema_gitlab_project_issue.go
```diff
@@ -0,0 +1,364 @@
...
+		},
+		// NOTE: it keep things simple users of this resource should use the `gitlab_user` data source to
+		//       get more information about the author if desired.
+		"author_id": {
+			Description: "The id of the author of the issue. Use `gitlab_user` data source to get more information about the user.",
+			Type:        schema.TypeInt,
+			Computed:    true,
+		},
+		"updated_at": {
+			Description: "When the issue was last. Date time string, ISO 8601 formatted, for example 2016-03-11T03:45:40Z.",
```
jtymes, line 141:
```suggestion
			Description: "When the issue was updated. Date time string, ISO 8601 formatted, for example 2016-03-11T03:45:40Z.",
```

internal/provider/schema_gitlab_project_issue.go
```diff
@@ -0,0 +1,364 @@
...
+		},
+		// NOTE: not part of `CREATE`, but part of `UPDATE` with the `state_event` field.
+		"state": {
+			Description:      fmt.Sprintf("The state of the issue. Valid values are: %s.", renderValueListForDocs(validIssueStates)),
+			Type:             schema.TypeString,
+			Optional:         true,
+			Default:          "opened",
+			ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice(validIssueStates, false)),
+		},
+		// NOTE: it keep things simple users of this resource should use the `gitlab_user` data source to
```
jtymes, line 133:
```suggestion
		// NOTE: to keep things simple, users of this resource should use the `gitlab_user` data source to
```

internal/provider/schema_gitlab_project_issue.go
```diff
@@ -0,0 +1,364 @@
...
+		"closed_at": {
+			Description: "When the issue was closed. Date time string, ISO 8601 formatted, for example 2016-03-11T03:45:40Z.",
+			Type:        schema.TypeString,
+			Computed:    true,
+			Optional:    true,
+		},
+		// NOTE: it keep things simple users of this resource should use the `gitlab_user` data source to
+		//       get more information about the closer if desired.
+		"closed_by_user_id": {
+			Description: "The id of the user that closed the issue. Use `gitlab_user` data source to get more information about the user.",
```
jtymes, line 155: nit: Noticing inconsistencies between using "id" and "ID" in descriptions. Not sure which is preferred.

internal/provider/schema_gitlab_project_issue.go
```diff
@@ -0,0 +1,364 @@
...
+			Computed:    true,
+			Optional:    true,
+		},
+		"closed_at": {
+			Description: "When the issue was closed. Date time string, ISO 8601 formatted, for example 2016-03-11T03:45:40Z.",
+			Type:        schema.TypeString,
+			Computed:    true,
+			Optional:    true,
+		},
+		// NOTE: it keep things simple users of this resource should use the `gitlab_user` data source to
```
jtymes, line 152:
```suggestion
		// NOTE: to keep things simple, users of this resource should use the `gitlab_user` data source to
```

internal/provider/schema_gitlab_project_issue.go
```diff
@@ -0,0 +1,364 @@
...
+			Type:        schema.TypeInt,
+			Computed:    true,
+		},
+		"epic_issue_id": {
+			Description: "The id of the epic issue.",
+			Type:        schema.TypeInt,
+			Computed:    true,
+			Optional:    true,
+		},
+		"epic_id": {
```
jtymes, line 254: Isn't there a comment on L83-89 saying this isn't supported?

docs/data-sources/project_issue.md
```diff
@@ -0,0 +1,94 @@
...
+  value = data.gitlab_project_issue.web_url
+}
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- **iid** (Number) The internal ID of the project's issue (requires administrator or project owner rights)
```
jtymes, line 38: Is this true for reading or just creating?

timofurrer: only for setting it, clarified.

internal/provider/schema_gitlab_project_issue.go
timofurrer: good catch ;)

timofurrer:
@armsnyder I've put it back to "draft", because I want to add a few more tests, however, I'm still happy if you'd review :)

internal/provider/data_source_gitlab_project_issues_test.go
```diff
@@ -0,0 +1,69 @@
...
+					resource.TestCheckResourceAttr("data.gitlab_project_issues.this", "issues.1.title", "Issue 2"),
+					resource.TestCheckResourceAttr("data.gitlab_project_issues.this", "issues.1.description", "Description 2"),
+				),
+			},
+		},
+	})
+}
+
+func testAccDataGitlabProjectIssuesConfig(rInt int) string {
+	return fmt.Sprintf(`
```
armsnyder, line 35: Point of information, and a bit of a nitpick for a data source, but, I would like to move in a direction where we are only including resources in the test TF config that are being tested.
- Makes tests faster
- Reduces issues where we cannot represent the "setup" behavior in TF config alone, providing good examples for future tests
- Allows us to use the `CheckDestroy` TestCase field correctly to check the destroy function, where we are destroying the tested resource without any of its dependencies like the project it lives in. (Not applicable to data sources, but still...)

So for instance here, the project and the two issues could be created in code before the call to `resource.Test...`. What do you think?

internal/provider/data_source_gitlab_project_issues.go
```diff
@@ -0,0 +1,379 @@
...
+	"priority",
+	"due_date",
+	"relative_position",
+	"label_priority",
+	"milestone_due",
+	"popularity",
+	"weight",
+}
+
+var validIssueScopeValues = []string{"created_by_me", "assigned_to_me", "all"}
```
armsnyder, line 27: For this and other resources, can you move your `valid*` variables into the `register*` function itself, to keep the global scope clean? This is mostly a readability concern for me.

internal/provider/resource_gitlab_project_issue.go
```diff
@@ -0,0 +1,294 @@
...
+		updateOptions.StateEvent = gitlab.String(issueStateToStateEvent[stateEvent.(string)])
+	}
+	if updateOptions != (gitlab.UpdateIssueOptions{}) {
+		_, _, err := client.Issues.UpdateIssue(project, issue.IID, &updateOptions, gitlab.WithContext(ctx))
+		if err != nil {
+			return diag.Errorf("failed to update issue %d in project %s right after creation: %v", issue.IID, project, err)
+		}
+	}
+
+	d.SetId(resourceGitLabProjectIssueBuildId(project, issue.IID))
```
armsnyder, line 126: Always set the ID as soon as you can. In this case, after the `client.Issues.CreateIssue` call. The reason is it makes the creation idempotent if the second API call to update the issue fails. As the code is written now, if the update call fails, then the whole resource will not be tracked in state, and the user would need to either import it or delete it manually.

internal/provider/resource_gitlab_project_issue.go
```diff
@@ -0,0 +1,294 @@
...
+		}
+	}
+
+	return nil
+}
+
+func resourceGitLabProjectIssueParseId(id string) (string, int, error) {
+	project, issue, err := parseTwoPartID(id)
+	if err != nil {
+		return "", -1, err
```
armsnyder, line 241: Go convention is to use the zero-value when there is an error. The caller is expected to inspect the error first.

internal/provider/helper_test.go
```diff
@@ -277,3 +278,10 @@ func testAccGitlabProjectStart(t *testing.T) testAccGitlabProjectContext {
 		project: project,
 	}
 }
+
+// testCheckResourceAttrLazy works like resource.TestCheckResourceAttr, but lazy evaluates the value parameter.
```
armsnyder, line 282: Usually `resource.TestCheckResourceAttrPtr` can be used. Might be good to add that context here.

```suggestion
// testCheckResourceAttrLazy works like resource.TestCheckResourceAttr, but lazy evaluates the value parameter.
// See also: resource.TestCheckResourceAttrPtr.
```

internal/provider/resource_gitlab_project_issue_test.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+
+		log.Printf("[DEBUG] testAccCheckGitlabProjectIssueExists: %#v", gotIssue)
+
+		*issue = *gotIssue
+		return nil
+	}
+}
+
+func testAccCheckGitlabProjectIssueDestroyAsClosed(s *terraform.State) error {
```
armsnyder, line 138: You could rewrite this as a single function for both `testAccCheckGitlabProjectIssueDestroyAsClosed` and `testAccCheckGitlabProjectIssueDestroy`, and alter your assertion based on the `delete_on_destroy` value. IMO this would be less error prone for test refactors.

internal/provider/util.go
```diff
@@ -272,3 +285,118 @@ func is404(err error) bool {
...
+
+func attributeNamesFromSchema(schema map[string]*schema.Schema) []string {
+	names := make([]string, 0, len(schema))
+	for name := range schema {
+		names = append(names, name)
+	}
+	return names
+}
+
+// datasourceSchemaFromResourceSchema is a recursive func that
```
armsnyder, line 341: 👏 👏 👏

armsnyder (changes requested):
Nice!

internal/provider/data_source_gitlab_project_issues_test.go
timofurrer: Yes, absolutely, I'm also nitpicking this usually. It was for testing convenience to get something - and was one of the points I was planning do to when implementing more tests 👍

internal/provider/helper_test.go
timofurrer: I'll reference it. In the case we are using the *lazy* function here, we are calling a function on the ptr, so that wouldn't be solved with the *ptr* function, right?

armsnyder: Yes in your case the function makes sense since you need to transform a value on the issue object.

docs/data-sources/project_issue.md
```diff
@@ -0,0 +1,94 @@
...
+  value = data.gitlab_project_issue.web_url
+}
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- **iid** (Number) The internal ID of the project's issue
```
armsnyder, line 38: nitpick but for consistency should end in period

docs/data-sources/project_issue.md
```diff
@@ -0,0 +1,94 @@
...
+- **web_url** (String) The web URL of the issue.
+- **weight** (Number) The weight of the issue. Valid values are greater than or equal to 0.
+
+<a id="nestedatt--task_completion_status"></a>
+### Nested Schema for `task_completion_status`
+
+Read-Only:
+
+- **completed_count** (Number)
+- **count** (Number)
```
armsnyder, line 92: Wouldn't `count` collide with the `count` meta-argument?

https://www.terraform.io/language/meta-arguments/count

internal/provider/resource_gitlab_project_issue_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+	var testIssue gitlab.Issue
+
+	testProject := testAccCreateProject(t)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: providerFactories,
+		CheckDestroy:      testAccCheckGitlabProjectIssueDestroy,
+		Steps: []resource.TestStep{
```
armsnyder, line 25: These tests are a little weak because they do not check what happened upstream. So this will not catch bugs resulting from not passing attributes to the upstream API and not setting the attribute in the READ function.

internal/provider/helper_test.go
armsnyder: Looks like your force-push reverted the applied suggestions.

armsnyder (dismissed):

docs/data-sources/project_issue.md
timofurrer: I'm not 100% sure, but I think it's okay. Other prominent provider use it, too in nested schemas at least:

* https://cs.github.com/hashicorp/terraform-provider-aws/blob/8d7b2519ddcfd2a4d6e9b907cee38aafe302733a/internal/service/ec2/instance_type_data_source.go?q=language%3Ago+org%3Ahashicorp+%5C%22count%5C%22+repo%3Ahashicorp%2Fterraform-provider-aws#L126
* https://cs.github.com/hashicorp/terraform-provider-aws/blob/8d7b2519ddcfd2a4d6e9b907cee38aafe302733a/internal/service/wafv2/web_acl.go?q=language%3Ago+org%3Ahashicorp+%5C%22count%5C%22+repo%3Ahashicorp%2Fterraform-provider-aws#L111

Also I think you can't access `count` as an attribute on a resource in case it used `count = N` (even though I think it used, too) - You'll get something like:

```
The special "count" attribute is no longer supported after Terraform v0.12. Instead, use length(gitlab_project.for_user) to count resource instances.
```

I think it's okay to use here, because it's a nested resource and not an resource block level and because it's read-only.

Please correct me if I'm wrong :D

armsnyder: Aaaaaaaaaaaah oops I thought I deleted this comment. Yeah I came to that conclusion when I realized it was a nested schema. Thanks for double checking.

internal/provider/resource_gitlab_project_issue_test.go
timofurrer: As discussed on discord :D

armsnyder (approved):
lgtm!

PR #895, opened 2022-02-24
resource/gitlab_group_variable: better error message for invalid masked variable values. Closes #371

### LINKED ISSUE #371: gitlab_project_variable: Add `masked` checks

timofurrer:


armsnyder (approved):
lgtm. Can you update `gitlab_instance_variable` as well?

timofurrer:
@armsnyder I've haded it for the instance variables, too 🎉

armsnyder (approved):

PR #898, opened 2022-02-24
Auto remove safe to test on acceptance EE run and push

DonNicoJs:
## Description

This MR updates the acceptance-EE workflow to auto remove safe to test label when:
- a new commit is pushed to the branch
-  the workflow runs

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/.github/workflows/pr-acceptance-ee.yml
+++ b/.github/workflows/pr-acceptance-ee.yml
@@ -60,6 +60,9 @@ jobs:
     runs-on: ubuntu-latest
     needs: [go-version, license-encryption-password]
     steps:
+      - uses: actions-ecosystem/action-remove-labels@v1
+        with:
+          labels: 'safe to test'
       - uses: actions/setup-go@v2
         with:
           go-version: ${{ needs.go-version.outputs.go-version }}
--- a/.github/workflows/pr-push.yml
+++ b/.github/workflows/pr-push.yml
@@ -1,10 +1,10 @@
 name: pr-push
 
 on: 
-  push:
-    branches-ignore:
-      - 'main'
-
+  workflows: ["pr-required-labels"]
+    types: 
+      - completed
+  
 concurrency: 
   group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
   cancel-in-progress: true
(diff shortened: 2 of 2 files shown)


DonNicoJs:
@timofurrer @armsnyder this is ready for your review!

timofurrer:
I guess the `skip ee test` can be ignored here, right? Once we've set it we trust that no relevant code which would impact ee tests has been pushed?

floh96:
Maybe also automatically add the safe-to-test label to pull requests from Collaborators?

DonNicoJs:
> I guess the `skip ee test` can be ignored here, right? Once we've set it we trust that no relevant code which would impact ee tests has been pushed?

i am not sure what is the ask here? To remove the check on `skip ee test`? Or to not run this workflow if `skip ee test` is set?

timofurrer:
@DonNicoJs It was more that I wanted to confirm my assumption that we don't target `skip ee test` here, too, because we assume that there won't be EE tests necessary to run once labelled. The risk here is that we don't run EE tests before a merge, because we once labelled it `skip ee test`, but then EE relevant code was pushed afterwards and we don't clear that label.

armsnyder (changes requested):
Hi @DonNicoJs. I think this needs more testing.

This would need to run on the `pull_request` event (not `push`) since `push` is not triggered in a PR from a fork.

Additionally, I believe there would be a race condition between this workflow and the workflow that checks that the PR has the required labels. Assuming this is rare, maybe we can live with it, especially if we are still considering #907

DonNicoJs:
@armsnyder @timofurrer I've updated this to run the new workflow AFTER `pr-required-labels` is completed this way we always avoid the race condition and we trigger the workflow at the right moment

timofurrer (approved):

.github/workflows/pr-acceptance-ee.yml
```diff
@@ -60,6 +60,9 @@ jobs:
     runs-on: ubuntu-latest
     needs: [go-version, license-encryption-password]
     steps:
+      - uses: actions-ecosystem/action-remove-labels@v1
```
armsnyder, line 63: This shouldn't work, because we made this pipeline run with read-only permissions. Did you test this in a fork? I'm concerned it will break the pipeline after we merge.

timofurrer:
I've just noticed that this change would also remove the label when I assign it when creating the PR, because the pr-label workflow wouldn't have run yet.

PR #899, opened 2022-02-24
resource/gitlab_user: Add support for `deactivated` user state. Closes #897

### LINKED ISSUE #897: 3.10.0 wants to re-activate all the users
### GitLab Provider version

3.10.0

### GitLab version

14.7.1-ee

### Terraform version

1.0.11

```hcl
resource "gitlab_user" "user" {
  name        = "Name"
  username    = "username"
  email       = "e@ma.il"
}
```

### Relevant log output

```shell
# gitlab_user.user will be updated in-place
~ resource "gitlab_user" "user" {
      id                = "2"
      name              = "Name"
    ~ state             = "deactivated" -> "active"
      # (8 unchanged attributes hidden)
  }
```

### Description

We just upgraded to 3.10.0 and the provider want to re-activate all the deactivated users.

If I set the `state` field with the value `deactivated`, then I get this error message:

```
expected state to be one of [active blocked], got deactivated
```

timofurrer:
This change will fix a regression from v3.9.0 to v3.10.0, see report in #897.

However, the regression is just because the user state has been changed outside of terraform, while it was supposed to be owner by terraform. It was changed outside of terraform, because the `gitlab_user` resource didn't support deactivating users.

Therefore, we could argue that this is actually a new feature and not a bug fix. But whatever, @ilpianista how urgent is the upgrade to 3.10 for you?

### DIFF
--- a/internal/provider/resource_gitlab_user.go
+++ b/internal/provider/resource_gitlab_user.go
@@ -14,6 +14,12 @@ import (
 	gitlab "github.com/xanzy/go-gitlab"
 )
 
+var validUserStateValues = []string{
+	"active",
+	"deactivated",
+	"blocked",
+}
+
 var _ = registerResource("gitlab_user", func() *schema.Resource {
 	return &schema.Resource{
 		Description: "This resource allows you to create and manage GitLab users.\n" +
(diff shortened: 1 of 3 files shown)


ilpianista:
I can build and test this myself without a new release if this is what you are asking.

timofurrer:
@ilpianista this would be nice :) But, it was more along the lines if we should release a 3.11 just for this or not :)

ilpianista:
Yay! I just ran `terraform plan` on the 3.10.1 release + this fix and I got:

> No changes. Your infrastructure matches the configuration.

Thank you a lot. No rush for a 3.11 release, I can use the version I just built meantime.

internal/provider/resource_gitlab_user.go
```diff
@@ -224,18 +236,33 @@ func resourceGitlabUserUpdate(ctx context.Context, d *schema.ResourceData, meta
 	}
 
 	if d.HasChange("state") {
-		if d.Get("state") == "active" {
-			err := client.Users.UnblockUser(id)
-
+		oldState, newState := d.GetChange("state")
+		var err error
+		// NOTE: yes, this can be written much more consice, however, for the sake of understanding the behavior,
+		//       of the API and the allowed state transitions of GitLab, let's keep it as-is and enjoy the readability.
```
armsnyder, line 242: <3 this

armsnyder (approved):

Issue #896, opened 2022-02-24
Cannot download provider 3.10.1


cmd-werner-diers:
### GitLab Provider version

3.10.1

### GitLab version

GitLab Enterprise Edition 14.9.0-pre 67dde8cf25d

### Terraform version

1.1.6

### Relevant Terraform Configuration

```hcl
terraform {
  required_providers {
    # Used for gitlab configuration
    gitlab = {
      source  = "gitlabhq/gitlab"
      version = "3.10.1"
    }
  }
}


provider "gitlab" { }
```

### Relevant log output

```shell
❯ tree -a
.
└── versions.tf

0 directories, 1 file


❯ terraform init

Initializing the backend...

Initializing provider plugins...
- Finding gitlabhq/gitlab versions matching "3.10.1"...
╷
│ Error: Failed to query available provider packages
│
│ Could not retrieve the list of available versions for provider gitlabhq/gitlab: no available releases match the given constraints 3.10.1
╵
```

### Description

Current behaviour:
Terraform cannot download the latest provider. 

Expected behaviour:
Terraform successfully installs the provider with the specified version

timofurrer:
@nagyv I can confirm that the 3.10.1 provider hasn't been properly released somehow. Can you please check?

timofurrer:
The pipeline apparently ran: https://github.com/gitlabhq/terraform-provider-gitlab/actions/runs/1891749446

nagyv:
Interesting. 3.10.1 is definitely not available in the Terraform registry: https://registry.terraform.io/providers/gitlabhq/gitlab

nagyv:
I triggered the webhook call for 3.10.1 again

armsnyder:
@nagyv I also tried running `terraform init` using the 3.10.1 version and it is not downloading.

```
Could not retrieve the list of available versions for provider gitlabhq/gitlab: no available releases match the given constraints 3.10.1
```

Looking at our releases, I don't see anything different about the 3.10.1 release, looking at the assets and workflow logs.

Can you double check that our GPG public key still exists at https://registry.terraform.io/settings/gpg-keys ?

nagyv:
> Can you double check that our GPG public key still exists at https://registry.terraform.io/settings/gpg-keys ?

The GPG key is there.

floh96:
@timofurrer @nagyv  the 3.10.1 version is not in the registry, maybe the webhook call needs to be triggered again?

[...]

PR #901, opened 2022-02-24
Fix the release process

nagyv:
## Description

Check the changes if we update the release process as described in https://www.terraform.io/registry/providers/publishing

Closes https://github.com/gitlabhq/terraform-provider-gitlab/issues/896

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,8 +15,18 @@ on:
     tags:
       - 'v*'
 jobs:
+  go-version:
+    runs-on: ubuntu-latest
+    outputs:
+      go-version: ${{ steps.go-version.outputs.go-version }}
+    steps:
+      - uses: actions/checkout@v2
+      # Read the .go-version file and output it for other jobs to use.
+      - id: go-version
+        run: echo "::set-output name=go-version::$(cat .go-version)"
   goreleaser:
     runs-on: ubuntu-latest
+    needs: [go-version]
     steps:
       -
         name: Checkout
@@ -28,7 +38,7 @@ jobs:
         name: Set up Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.14
+          go-version: ${{ needs.go-version.outputs.go-version }}
       -
         name: Import GPG key
         id: import_gpg
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,12 +1,12 @@
 # This GitHub action can publish assets for release when a tag is created.
 # Currently its setup to run on any tag that matches the pattern "v*" (ie. v0.1.0).
 #
-# This uses an action (paultyng/ghaction-import-gpg) that assumes you set your
+# This uses an action (hashicorp/ghaction-import-gpg) that assumes you set your 
 # private key in the `GPG_PRIVATE_KEY` secret and passphrase in the `PASSPHRASE`
 # secret. If you would rather own your own GPG handling, please fork this action
 # or use an alternative one for key handling.
 #
-# You will need to pass the `--batch` flag to `gpg` in your signing step
+# You will need to pass the `--batch` flag to `gpg` in your signing step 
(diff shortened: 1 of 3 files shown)


nagyv:
@timofurrer @armsnyder As I went through the Terraform provider publishing docs and updated the files you can see here, I think the only major difference is that we don't have a `terraform-registry-manifest.json` file.

The current, recommended release workflow is basically the same that we have. The GoReleaser configuration file differs in references to `terraform-registry-manifest.json`. Could this be the issue with the release process?

See https://www.terraform.io/registry/providers/publishing#terraform-registry-manifest-file

.github/workflows/release.yml
```diff
@@ -15,39 +15,35 @@ on:
...
+        uses: actions/checkout@v2.4.0
+      -
+        name: Unshallow
         run: git fetch --prune --unshallow
-      - uses: paultyng/ghaction-import-gpg@v2.1.0
+      -
+        name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.14
```
armsnyder, line 31: Could you revert this line and use the go-version from the go-version job, so we use the same go-version in the release as we do in the build/test?

.goreleaser.yml
```diff
@@ -1,5 +1,14 @@
+# Visit https://goreleaser.com for documentation on how to customize this
+# behavior.
+before:
+  hooks:
+    # this is just an example and not a requirement for provider building/publishing
```
armsnyder, line 5: Let's remove this hook since it's just an example

.goreleaser.yml
```diff
@@ -24,17 +33,28 @@ archives:
 - format: zip
   name_template: '{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}'
 checksum:
+  extra_files:
+    - glob: 'terraform-registry-manifest.json'
```
armsnyder, line 37: We would need to add this file to the root of the repository.

https://www.terraform.io/registry/providers/publishing#terraform-registry-manifest-file

> This file is conventionally named terraform-registry-manifest.json at the root of the provider repository and will be renamed in the release assets.

armsnyder (changes requested):

timofurrer:
I'm very curious what has changed that we are unable to release ... does it have to do with GitHub settings we made regarding protection of EE ?

.github/workflows/release.yml
nagyv: Done.

.goreleaser.yml
nagyv: Done.

.goreleaser.yml
nagyv: I added the file.

nagyv:
@timofurrer I think we did not change the release specific jobs. My guess is that we did not have the `terraform-registry-manifest.json` until now.

nagyv:
@armsnyder I follow-up on your requests. Please, check and approve.

timofurrer:
Can we actually build a `-pre` release to test this?

armsnyder (dismissed):
Certainly worth trying, although to be clear I don't understand what went wrong in the first place. Did you see [my comment](https://github.com/gitlabhq/terraform-provider-gitlab/issues/896#issuecomment-1050549569) in the issue?

timofurrer:
@nagyv I can't spot the release 👀

nagyv:
> @nagyv I can't spot the release 👀

Neither do I. This release did not succeed either.

timofurrer:
Just noticed, too. However, I don't think that it's related ;)

timofurrer:
Worked: https://registry.terraform.io/providers/gitlabhq/gitlab/3.11.0-pre

Issue most likely was the missing manifest file (not sure why that appeared now 🤔 ) 

Also, the terraform upstream docs had a bug which I fixed here and in the upstream docs itself:

* https://github.com/hashicorp/terraform-website/pull/2161

PR #716, opened 2021-09-14
Add gitlab_user_sshkey resource

eonpatapon:
Closes #40

Allows to manage users ssh keys

PR #903, opened 2022-02-26
resource/gitlab_user_sshkey: add new resource. Closes #40, 716 

### LINKED ISSUE #40: Feature request: Support user SSH keys management
https://docs.gitlab.com/ce/api/users.html#add-ssh-key-for-user

timofurrer:
I took the liberty to finish #716 and therefore made some significant changes to it to make it work properly.
Thanks @eonpatapon for getting this started 🎉 

Closes: #40 
Closes #716 

I'll followup up next with on this PR with other PRs for global ssh key resources and MR in GitLab itself to improve the API.


go.mod
```diff
@@ -12,6 +12,6 @@ require (
 	github.com/mitchellh/hashstructure v1.1.0
 	github.com/onsi/gomega v1.18.1
 	github.com/xanzy/go-gitlab v0.55.1
-	golang.org/x/crypto v0.0.0-20210616213533-5ff15b29337e
+	golang.org/x/crypto v0.0.0-20220214200702-86341886e292 // indirect
```
eonpatapon, line 15: Looks like this is not needed anymore

eonpatapon:
Thanks for moving this forward!

go.mod
armsnyder: This crypto dependency was always there; it was just hidden before. Starting in Go 1.17 you are required to include all indirect dependencies in the go.mod anyway. This go.mod is still using 1.16 as the min Go version.

timofurrer: Exactly, you'll see that "change" in `go.sum`. I'll leave it as is, since it's marked as `// indirect` anyways.

internal/provider/resource_gitlab_user_sshkey.go
```diff
@@ -0,0 +1,193 @@
...
+		return diag.FromErr(err)
+	}
+
+	return nil
+}
+
+func resourceGitlabUserSSHKeyParseID(id string) (int, int, error) {
+	userIDFromID, keyIDFromID, err := parseTwoPartID(id)
+	if err != nil {
+		return -1, -1, err
```
armsnyder, line 181: Ah, same comment as in my other review. Convention is to use the zero-value for other return values when the last return value is a non-nil error.

internal/provider/resource_gitlab_user_sshkey_test.go
```diff
@@ -0,0 +1,179 @@
...
+		userID, keyID, err := resourceGitlabUserSSHKeyParseID(rs.Primary.ID)
+		if err != nil {
+			return fmt.Errorf("failed to parse user ssh key resource ID: %s", err)
+		}
+
+		keys, _, err := testGitlabClient.Users.ListSSHKeysForUser(userID, &gitlab.ListSSHKeysForUserOptions{})
+		// User deleted as well as its keys
+		if err != nil {
+			if is404(err) {
+				return nil
```
armsnyder, line 80: This should be an error if the user doesn't exist IMO

armsnyder (changes requested):

armsnyder:
Forget to push? 🤔

timofurrer:
oups, check again 😇

armsnyder (approved):

PR #907, opened 2022-03-01
Workflows: Run EE tests on PRs without label requirement

armsnyder:
## Description

Addresses an issue where EE tests only run when the `safe to test` label is added by returning to the previous behavior of running EE tests on all pushes to a PR.

I'm opening this as an alternative to #898 which I believe is now feasible since #902 was merged.

    - Create the resource
    - Update the attributes
    - Import the resource


timofurrer (approved):
LGTM 🎉

Issue #904, opened 2022-02-28
Add any_approver option for gitlab_project_approval_rule


petrmvala:
### Feature Description

This is a part of GitLab API. Setting `approval_rule = "any_approver"` enables setting the Eligible approvers in the project, enabling for the proper use of the CODEOWNERS file.

It is currently blocked in the go-gitlab library, but that should be fixed with this PR:

- https://github.com/xanzy/go-gitlab/pull/1396

### Do you want to implement this?

- [X] I would like to implement this myself 👷

petrmvala:
Unblocked

PR #913, opened 2022-03-03
feat: Add rule_type parameter to merge_request_approval_rule

petrmvala:
This is working in the API, and is necessary in order to enable
Eligible Approvers.

## Description

#904 

    - Create the resource
    - Update the attributes
    - Import the resource


internal/provider/resource_gitlab_project_approval_rule.go
```diff
@@ -72,6 +77,7 @@ func resourceGitlabProjectApprovalRuleCreate(ctx context.Context, d *schema.Reso
 	options := gitlab.CreateProjectLevelRuleOptions{
 		Name:               gitlab.String(d.Get("name").(string)),
 		ApprovalsRequired:  gitlab.Int(d.Get("approvals_required").(int)),
+		RuleType:           gitlab.String(d.Get("rule_type").(string)),
```
timofurrer, line 80: If an attribute is not set in the config and is optional it's good to also not send it to the API, therefore, I'd do it like this:

https://github.com/gitlabhq/terraform-provider-gitlab/blob/278f2b394ec57caf9e415c2b42d31bdb48b76e78/internal/provider/resource_gitlab_topic.go#L56

I'm aware that it also applies to the attributes already in here. It's never to late to start though and if you want feel free to apply it for the others, too 🌮

internal/provider/resource_gitlab_project_approval_rule_test.go
```diff
@@ -57,12 +58,82 @@ func TestAccGitLabProjectApprovalRule_basic(t *testing.T) {
...
+			{
+				ResourceName:      "gitlab_project_approval_rule.foo",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func TestAccGitLabProjectApprovalRule_any_approver(t *testing.T) {
```
timofurrer, line 84: NIT:

```suggestion
func TestAccGitLabProjectApprovalRule_anyApprover(t *testing.T) {
```

timofurrer (changes requested):
Looks very promising, nice work! I've left a few minor comments.

internal/provider/resource_gitlab_project_approval_rule.go
```diff
@@ -43,6 +43,11 @@ var _ = registerResource("gitlab_project_approval_rule", func() *schema.Resource
 				Type:        schema.TypeInt,
 				Required:    true,
 			},
+			"rule_type": {
+				Description: "The type of rule. `any_approver` is a pre-configured default rule with `approvals_required` at `0`. Other rules are `regular`.",
```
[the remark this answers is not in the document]
petrmvala: Yes, exactly. I took the docs text from here: https://docs.gitlab.com/14.7/ee/api/merge_request_approvals.html#create-project-level-rule

Thanks for the comments and let me see :)

internal/provider/resource_gitlab_project_approval_rule.go
petrmvala: I would prefer to touch only this one in this scope and then go through the others separately, would that be ok?

timofurrer: sure thing! You can also leave it and I'll sweep over it afterwards ;)

internal/provider/resource_gitlab_project_approval_rule_test.go
petrmvala: Actually, shouldn't it be `_AnyApprover` ?

internal/provider/resource_gitlab_project_approval_rule.go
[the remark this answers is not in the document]
timofurrer: Yes, I saw that, too - docs are sometimes incomplete (🤫 ) that's why I've asked

internal/provider/resource_gitlab_project_approval_rule.go
```diff
@@ -3,17 +3,24 @@ package provider
...
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 	gitlab "github.com/xanzy/go-gitlab"
 )
 
 // https://docs.gitlab.com/ee/api/merge_request_approvals.html#create-project-level-rule
 var errApprovalRuleNotFound = errors.New("approval rule not found")
 
+var validRuleTypeValues = []string{
```
timofurrer, line 19: Let's move this in the `registerResource()` call between current line 24 / 25

timofurrer (approved):
LGTM 🎉

timofurrer:
@DonNicoJs @armsnyder I've merged this because all checks passed and I've assigned the `safe to test` label (this is an EE feature) - However, seems like EE tests were not executed and `main` is now failing :'( Not sure what went wrong here. But I'll try to make `main` pass again ;)

PR #916, opened 2022-03-04
feat: Add rule_type parameter to merge_request_approval_rule

petrmvala:
This is working in the API, and is necessary in order to enable
Eligible Approvers.

Closes: #904

    - Create the resource
    - Update the attributes
    - Import the resource


internal/provider/resource_gitlab_project_approval_rule.go
```diff
@@ -43,6 +49,13 @@ var _ = registerResource("gitlab_project_approval_rule", func() *schema.Resource
 				Type:        schema.TypeInt,
 				Required:    true,
 			},
+			"rule_type": {
+				Description:      fmt.Sprintf("String, defaults to 'regular'. The type of rule. `any_approver` is a pre-configured default rule with `approvals_required` at `0`. Valid values are %s.", renderValueListForDocs(validRuleTypeValues)),
+				Type:             schema.TypeString,
+				Optional:         true,
+				Default:          "regular",
```
timofurrer, line 56: NIT:

```suggestion
```

I'd leave it up to GitLab to decide on the `rule_type` ... thus, we don't send it, if it wasn't set :)

timofurrer (changes requested):
One minor thing, otherwise LGTM 🎉

internal/provider/resource_gitlab_project_approval_rule.go
petrmvala: Ah, right. makes sense :)

timofurrer (dismissed):
LGTM 🎉

internal/provider/resource_gitlab_project_approval_rule.go
```diff
@@ -43,6 +49,12 @@ var _ = registerResource("gitlab_project_approval_rule", func() *schema.Resource
 				Type:        schema.TypeInt,
 				Required:    true,
 			},
+			"rule_type": {
+				Description:      fmt.Sprintf("String, defaults to 'regular'. The type of rule. `any_approver` is a pre-configured default rule with `approvals_required` at `0`. Valid values are %s.", renderValueListForDocs(validRuleTypeValues)),
+				Type:             schema.TypeString,
+				Optional:         true,
```
timofurrer, line 55:
```suggestion
				Optional:         true,
				Computed:      true,
```

timofurrer (approved):

timofurrer:
@petrmvala I think we have a problem ...

This change doesn't handle the update at all. I've just double checked go-gitlab and the upstream API and neither support the *rule type* (see https://docs.gitlab.com/ee/api/merge_request_approvals.html#update-project-level-rule)

I think for now we have to mark it with [`ForceNew`](https://www.terraform.io/plugin/sdkv2/schemas/schema-behaviors#forcenew). That being said a test changing the rule type would be great.

timofurrer (changes requested):
@petrmvala I think we have a problem ...

This change doesn't handle the update at all. I've just double checked go-gitlab and the upstream API and neither support the *rule type* (see https://docs.gitlab.com/ee/api/merge_request_approvals.html#update-project-level-rule)

I think for now we have to mark it with [`ForceNew`](https://www.terraform.io/plugin/sdkv2/schemas/schema-behaviors#forcenew). That being said a test changing the rule type would be great.

internal/provider/resource_gitlab_project_approval_rule_test.go
```diff
@@ -79,15 +79,83 @@ func TestAccGitLabProjectApprovalRule_basic(t *testing.T) {
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGitlabProjectApprovalRuleExists("gitlab_project_approval_rule.bar", &projectApprovalRule),
+					testAccCheckGitlabProjectApprovalRuleAttributes_AnyApprover(&projectApprovalRule, &testAccGitlabProjectApprovalRuleExpectedAttributes_AnyApprover{
+						Name:              "bar",
+						ApprovalsRequired: 2,
+						RuleType:          "any_approver",
+					}),
+				),
+			},
+			// Re-create rule
```
timofurrer, line 122:
```suggestion
			// Update rule
```

timofurrer (approved):

Issue #636, opened 2021-06-16
I should be able to configure a project's `resolve_outdated_diff_discussions`


alexbowers:
### Expected Behavior
I should be able to configure `resolve_outdated_diff_discussions` option via Edit Project API call.

It is listed on here: https://docs.gitlab.com/ee/api/projects.html (search for resolve_outdated_diff_discussions)

### Actual Behavior
Nothing

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- https://github.com/pulumi/pulumi-gitlab/issues/125

Issue #541, opened 2021-01-30
Feature Request - gitlab_project - additional argument reference (option) - Skip outdated deployment jobs


Gvazzana:
### Feature Request Details:
Please add the ability to configure the following ci/cd general project settings on the 'Gitlab Project' Resource. Toggle Options could be a Boolean (true / false) or enable, disable
- Auto-cancel redundant, pending pipelines
- Skip outdated deployment jobs

armsnyder:
It sounds like this is an ask to add one more setting: `auto_cancel_pending_pipelines` to the `gitlab_project` resource.

RomanNess:
Thought, I could do a quick PR for this before dinner. Tough luck, `AutoCancelPendingPipelines` is not implemented in `xanzy/go-gitlab` where the `project` is read from API. 
https://github.com/xanzy/go-gitlab/blob/d803e68bb3041a361bb749dbb951026b8e4607bb/projects.go#L40

However, it is already included in POST and PUT.
https://github.com/xanzy/go-gitlab/blob/d803e68bb3041a361bb749dbb951026b8e4607bb/projects.go#L517
https://github.com/xanzy/go-gitlab/blob/d803e68bb3041a361bb749dbb951026b8e4607bb/projects.go#L654

Gvazzana:
Thank you for the update. Is there anyway to request this feature be added to the gitlab API, with more weight coming from the Terraform Provider group?

RomanNess:
I'm not sure if it is missing in the REST API or if it is just not yet implemented in `xanzy/go-gitlab`. The first step would be to clarify this. 
Then for the latter case there's always the option to contribute the missing attributes or at least to file an issue in `xanzy/go-gitlab` that links this issue.

ngoduykhanh:
**Auto-cancel redundant, pending pipelines** is `auto_cancel_pending_pipelines`
**Skip outdated deployment jobs** is `ci_forward_deployment_enabled`

They are both available in Gitlab API but missing from `xanzy/go-gitlab`.
| Name                          | GitLab-API GET | GitLab-API POST | GitLab-API PUT | xanzy/go-gitlab GET | xanzy/go-gitlab POST | xanzy/go-gitlab PUT |
|-------------------------------|----------------|-----------------|----------------|---------------------|----------------------|---------------------|
| auto_cancel_pending_pipelines |        x       |        x        |        x       |                     |           x          |          x          |
[...]

Issue #477, opened 2020-11-09
Support all attributes supported by API in gitlab_project resource


Zyava:
I'm very surprised that `gitlab_project` resource doesn't support all attributes available in the API:
https://docs.gitlab.com/ee/api/projects.html#create-project. Please support all attributes API supports - like it is done in other providers (AWS, k8s, etc.)

### Terraform Version
TF 0.13.5, Gitlab provider version 3.1.0

### Affected Resource(s)
- gitlab_project

### Expected Behavior
All attributes accepted by Gitlab API (https://docs.gitlab.com/ee/api/projects.html#create-project) are supported by `gitlab_project` resource.

### Actual Behavior
Some random subset of attributes is supported.

kitos9112:
This would be great - Do you have plans to get a close 100% coverage of the existing GitLab v4 API to create projects?

@armsnyder

cmd-werner:
This would be a great addition

timofurrer:
I've attempted this (at least for now) in #917 

However, this is a never-ending story, thus I'll close this issue once #917 is merged and we can open dedicated issues for missing attributes afterwards.

Issue #351, opened 2020-06-25
[feature request] forking_access_level and pages_access_level


FernandoMiguel:
### Affected Resource(s)
Please list the resources as a list, for example:
- pages_access_level
- forking_access_level

ideally keeping in sync with https://docs.gitlab.com/ee/api/projects.html#create-project

tobiasbueschel:
Thanks a lot for everyone who has been maintaining this repo!

I agree with @FernandoMiguel, would be great if the GitLab repo creation properties are kept in sync, e.g. I have a use case for disabling GitLab's page feature across all repos that I create with Terraform.

PR #917, opened 2022-03-05
resource/gitlab_project: Add many attributes to align with upstream API

timofurrer:
This is work in progress and is currently blocked by:

* https://github.com/xanzy/go-gitlab/pull/1401

Also missing:

- [x] Data Source alignment
- [x] Testing

Closes: #636 
Closes: #286 
Closes: #351 
Closes: #928 
Closes: #946
Closes: #477 
Closes: #541 
Closes: #947


internal/provider/resource_gitlab_project.go
```diff
@@ -387,11 +403,201 @@ var resourceGitLabProjectSchema = map[string]*schema.Schema{
...
+		Type:        schema.TypeBool,
+		Optional:    true,
+	},
+	"autoclose_referenced_issues": {
+		Description: "Set whether auto-closing referenced issues on default branch.",
+		Type:        schema.TypeBool,
+		Optional:    true,
+	},
+	"build_git_strategy": {
+		Description: "The Git strategy. Defaults to fetch.",
```
jtymes, line 440: This could use a `ValidateDiagFunc` for values `clone, fetch`. https://gitlab.com/gitlab-org/gitlab/-/blob/89d18cb96aad3e6522228948f6e17bf7f89a9904/lib/api/helpers/projects_helpers.rb#L13

jtymes (commented):
Since you're in here 😉, I found this change the other day when I was looking through the `docs` folder in GitLab for the upcoming release.

For `path`:
>Repository name for new project. Generated based on name if not provided (generated as lowercase with dashes). Starting with GitLab 14.9, path must not start or end with a special character and must not contain consecutive special characters. 

I don't know that this needs its own validation function, though that could be considered.

timofurrer:
> I don't know that this needs its own validation function, though that could be considered.

hmmm, I'm always torn about what to validate in the provider - especially since we are implicitly supporting multiple GitLab versions. We may implement a check though for this one outside of this PR :)

jtymes (approved):

internal/provider/resource_gitlab_project_test.go
```diff
@@ -525,66 +616,48 @@ func TestAccGitlabProject_MergeTrains(t *testing.T) {
...
-					testAccCheckAggregateGitlabProject(&willError, &received),
-				),
+				Config: fmt.Sprintf(`
+					resource "gitlab_project" "this" {
+						name = "foo-%d"
+						repository_storage = "non-existing"
+
+						visibility_level = "public"
+					}`, rInt),
+				ExpectError: regexp.MustCompile(`repository_storage_moves: \[is invalid\]`),
```
armsnyder, line 649: Where does the string `repository_storage_moves` come from?

internal/provider/resource_gitlab_project_test.go
```diff
@@ -1620,3 +1876,68 @@ resource "gitlab_project" "foo" {
...
+  squash_commit_template = "hello squash"
+  merge_commit_template = "hello merge"
+
+  // EE features
+  approvals_before_merge = 2
+  external_authorization_classification_label = "test"
+  requirements_access_level = "enabled"
+  // are tested in separate test case
+  // mirror_trigger_builds = true
+  // mirror = true
```
armsnyder, line 1940: style: to not mix comment styles...

```suggestion
  # EE features
  approvals_before_merge = 2
  external_authorization_classification_label = "test"
  requirements_access_level = "enabled"
  # are tested in separate test case
  # mirror_trigger_builds = true
  # mirror = true
```

armsnyder (changes requested):
This is a huge win, thanks!

internal/provider/resource_gitlab_project_test.go
timofurrer: I asked that myself - it's what is returned by the upstream API ... but I guess it's related to the `repository_storage` stuff which this test does intentionally wrong ...

armsnyder: I think this is worth a comment, or, changing the regex to just be "is invalid", just because `repository_storage_moves` is kind of foreign to the provider code and not really relevant to the asserted behavior

timofurrer: done

armsnyder (approved):

PR #920, opened 2022-03-06
Support file_template_project_id attribute in the gitlab_group resource

PatrickRice-KSC:
Adds support for file_template_project_id

## Description

Closes #736

    - Create the resource
    - Update the attributes
    - Import the resource


PatrickRice-KSC:
This is not yet ready for merger, however I needed to open the PR to let the remote EE tests run. The shared_runner_limits require both an admin token and a local instance of GitLab EE to test, and my trial for ultimate local ran out :)

Within CE, the attributes will only ever return 0 so far in my tests, so I'd like to see the results from the EE tests.

internal/provider/resource_gitlab_group.go
```diff
@@ -148,11 +148,33 @@ var _ = registerResource("gitlab_group", func() *schema.Resource {
...
 				Optional:    true,
-				Default:     false,
+			},
+			"file_template_project_id": {
+				Description: "The ID of a project to load custom file templates from. Requires GitLab Premium.",
+				Type:        schema.TypeInt,
+				Optional:    true,
+			},
+			"shared_runners_minutes_limit": {
+				Description: "Maximum number of monthly CI/CD minutes for this group. Can be nil " +
```
timofurrer, line 162: I think it makes sense to use back ticks here :)

internal/provider/resource_gitlab_group_test.go
```diff
@@ -275,6 +275,72 @@ func TestAccGitlabGroup_PreventForkingOutsideGroup(t *testing.T) {
 	})
 }
 
+func TestAccGitlabGroup_updatedAttributes(t *testing.T) {
```
timofurrer, line 278: NIT:

```suggestion
func TestAccGitlabGroup_u=UpdatedAttributes(t *testing.T) {
```

timofurrer (changes requested):

internal/provider/resource_gitlab_group.go
```diff
@@ -284,6 +321,10 @@ func resourceGitlabGroupRead(ctx context.Context, d *schema.ResourceData, meta i
 	d.Set("runners_token", group.RunnersToken)
 	d.Set("share_with_group_lock", group.ShareWithGroupLock)
 	d.Set("default_branch_protection", group.DefaultBranchProtection)
+	d.Set("membership_lock", group.MembershipLock)
```
jtymes, line 324: I'm not seeing `membership_lock` coming back when running a GET on a group, so I'm a bit confused where this would be coming from.

PatrickRice-KSC: You know, I think you're right~ I just took another read through their Group API documentation, and it doesn't list this attribute as one that's returned. It's kinda vague on _exactly_ which attributes are returned other than providing an example, but their example doesn't explicitly mentioned it and it's not mentioned as an "additional" attribute when licensed at a higher level either. Really odd that you can pass it in both create/update, but it doesn't come back in "get", but that exactly described the behavior I'm seeing in the test.

Nice catch @jtymes !

jtymes: I agree that it's strange. It _seems_ like it should be fairly easy for them to expose. I'll take a look and maybe make an upstream MR.

PatrickRice-KSC:
I'm seeing the exact same behavior in EE for the shared runner limits as I'm seeing within CE - namely that the `get` API is returning  0 for the shared runner limits, even though they're being set in the Create API. I'm going to go ahead and back out those changes for now, because I don't think it makes sense to attempt to troubleshoot those API oddities here.

internal/provider/resource_gitlab_group.go
timofurrer: Yes, this happens .. I've had a similar encounter yesterday : https://gitlab.com/gitlab-org/gitlab/-/issues/354657

PatrickRice-KSC:
I'm going to try one more thing - I was thinking about this overnight and wonder if the fact that the group is "public" means that it's not working properly since the additional shared runner limits are explicitly noted in the API as requiring a license. It's a long shot, but it's just one commit to test it out, then I'll revert.

internal/provider/resource_gitlab_group.go
PatrickRice-KSC: fixed, thanks for the reminder 😃 , my Java background leaking out here!

PatrickRice-KSC:
OK, I reverted the membership_lock and runner limits. I'm going to squash just since there has been so much iteration, then we should be good for review.

internal/provider/resource_gitlab_group.go
```diff
@@ -235,6 +239,14 @@ func resourceGitlabGroupCreate(ctx context.Context, d *schema.ResourceData, meta
...
 	}
 
+	if v, ok := d.GetOk("file_template_project_id"); ok {
+		updateOptions.FileTemplateProjectID = gitlab.Int(v.(int))
+	}
+
+	if v, ok := d.GetOk("prevent_forking_outside_group"); ok {
+		updateOptions.PreventForkingOutsideGroup = gitlab.Bool(v.(bool))
+	}
+
```
armsnyder, line 249: Duplicate code block

```suggestion
```

internal/provider/resource_gitlab_group_test.go
```diff
@@ -275,6 +275,67 @@ func TestAccGitlabGroup_PreventForkingOutsideGroup(t *testing.T) {
...
+				Config:   testAccGitLabGroupWithProjectTemplate(rInt),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGitlabGroupExists("gitlab_group.test-group", &group),
+					testAccCheckGitlabProjectExists("gitlab_project.foo", &project),
+					resource.TestCheckResourceAttr("gitlab_group.test-group",
+						"file_template_project_id", fmt.Sprintf("%d", project.ID)),
+					resource.TestCheckResourceAttr("gitlab_group.test-group",
+						"shared_runners_minutes_limit", "1"),
+					resource.TestCheckResourceAttr("gitlab_group.test-group",
+						"extra_shared_runners_minutes_limit", "1"),
```
armsnyder, line 301: Leftover test code which can be deleted

```suggestion
```

armsnyder (changes requested):
Thanks for the contribution! Some comments...

PatrickRice-KSC:
@armsnyder  - Thanks for the catches... missed removing that dead code. Too many iterations on this PR!

internal/provider/resource_gitlab_group_test.go
```diff
@@ -275,6 +275,63 @@ func TestAccGitlabGroup_PreventForkingOutsideGroup(t *testing.T) {
...
+	var project gitlab.Project
+	rInt := acctest.RandInt()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: providerFactories,
+		CheckDestroy:      testAccCheckGitlabGroupDestroy,
+		Steps: []resource.TestStep{
+			{
+				//This test requires EE because without it, the two "Shared Runner" settings will always return 0,
```
timofurrer, line 289: Is this comment still valid?

internal/provider/resource_gitlab_group_test.go
```diff
@@ -275,6 +275,63 @@ func TestAccGitlabGroup_PreventForkingOutsideGroup(t *testing.T) {
...
+				//  even though they will work. They also require an admin token.
+				SkipFunc: isRunningInCE,
+				Config:   testAccGitLabGroupWithProjectTemplate(rInt),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGitlabGroupExists("gitlab_group.test-group", &group),
+					testAccCheckGitlabProjectExists("gitlab_project.foo", &project),
+					resource.TestCheckResourceAttr("gitlab_group.test-group",
+						"file_template_project_id", fmt.Sprintf("%d", project.ID)),
+				),
+			},
```
timofurrer, line 299: With a subsequent import verification step you'd verify that not only the terraform state is correctly set, but also the upstream API is correctly called.

timofurrer (changes requested):

internal/provider/resource_gitlab_group_test.go
PatrickRice-KSC: I'm working on adding this, but I'm getting some intermittent failures when running tests locally that I'm not happy with, so I'm doing some debugging.

PatrickRice-KSC: So, turns out there are some weird debugging quirks with GitPod that I'll see if I can figure out in a different issue. Swapping to local development resolved the intermittent failures. 

I updated the comment line because the test still does require EE I believe (otherwise the file_template_project_id just doesn't return), and added an import test.

timofurrer (approved):

PatrickRice-KSC:
@timofurrer | @armsnyder - Ok, we've got a bit of a problem here. So the EE test fails because the project ID never gets set successfully, even though the option is included properly. The reason for that is kind of silly, and I'm frustrated at myself for not catching this earlier (so much time spent for all of us, sorry!): *It's not the project_id that gets set on the child group, it's the child group's ID that gets set on the parent group, and then all projects in the child are available as templates.* The child group is always coming back with 0 because if you pass it a project ID instead of a group ID it just ignores it (instead of throwing an error for an invalid parameter, which is what I'd expect).

The reason this is a problem is because I'm not sure there is an actual way to simulate the required flow necessary to do this within a terraform configuration. [The flow needed is to](https://docs.gitlab.com/ee/user/group/custom_project_templates.html#set-up-group-level-project-templates):

1. Create the parent group (in the test config, `gitlab_group.foo`)
2. Create the child group (in the test config, `gitlab_group.test-group`)
3. Create the child project in the child group (in the test config, `gitlab_project.foo`)
4. *Update the parent group with the ID of the child group in the `file_template_project_id` field*

It's obviously that last part that's the problem since Terraform is deterministic in nature - It creates a circular dependency in the execution path. The only way I can think of fixing this issue is by creating a completely new "gitlab_group_level_template" resource, which will then call the UpdateGroup API behind the scenes. That way the terraform resource dependencies will flow like:

- Create Parent Group
- Create Child Project
- Create Child Group
- Create new "gitlab_group_level_template" resource which links the parent group and child group.

This is less than intuitive, given that the it's not a "many to many" relationship which is where you usually see this pattern, and users of the provider would _expect_ to see the  file_template_project_id within the group resource given it's position in the API. I'm not sure how else to solve the issue though.

Thoughts?

Stromweld:
I think that acceptable to make it it’s own resource. TF team in the AWS provider started doing similar pattern to also help break up the s3 resource as it had so many large attributes and to help with code maintainability.

timofurrer:
@PatrickRice-KSC 

> It's not the project_id that gets set on the child group, it's the child group's ID that gets set on the parent group, and then all projects in the child are available as templates. 

So, that also means that the upstream docs are wrong, describing the field as:

> The ID of a project to load custom file templates from.

> The only way I can think of fixing this issue is by creating a completely new "gitlab_group_level_template" resource, which will then call the UpdateGroup API behind the scenes.

I'm afraid I can't think of a better solution than that ...

Maybe you can also create an issue upstream for the misleading docs ...

armsnyder:
Yeah, I agree with the conclusion that it should be a new resource.

PR #921, opened 2022-03-06
Improve documentation and reference upstream API for every data source and resource

timofurrer:
This change improves the consistency of the data source and resource documentation and consistently references the upstream API documentation, which IMHO is usually essential to understand the full functionality of the underlying data source or resource.

### DIFF
--- a/docs/data-sources/branch.md
+++ b/docs/data-sources/branch.md
@@ -3,12 +3,15 @@
 page_title: "gitlab_branch Data Source - terraform-provider-gitlab"
 subcategory: ""
 description: |-
-  Provide details about a gitlab project branch
+  The gitlab_branch data source allows details of a repository branch to be retrieved by its name and project.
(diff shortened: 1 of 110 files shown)


armsnyder (approved):
Wow, thanks for the extensive change. I agree the upstream API should always be linked in some manner. The doc page is a good place to do it. Assuming it is simple to do, should we enforce this as a CI check so we don't regress on this? I could take a stab at writing an OPA policy and run it against the markdown files, unless anyone knows a better way.

timofurrer:
@armsnyder OPA might be a little bit of an overkill, maybe some simple `grep` could already do the job :D 

However, up to you - I'm happy if there is *something* 🎉

armsnyder:
@timofurrer Yeah I also want to avoid overkill. My thinking is, since this would be such a specific assertion, it should be added to the CI in an extensible way.

armsnyder:
@timofurrer Done in https://github.com/gitlabhq/terraform-provider-gitlab/pull/942 using grep

PR #923, opened 2022-03-06
Automatically add changelog as release notes to GitHub release

timofurrer:
There are other providers which do basically the same thing, e.g. see https://github.com/hashicorp/terraform-provider-google/blob/main/.github/workflows/release.yml


armsnyder (approved):
Love it! 🎉

PR #924, opened 2022-03-06
Correctly handle deletion outside of terraform

timofurrer:
This change fixes the read function of the following resources to properly handle out of bound deletions:

- `gitlab_project` 
- `gitlab_project_badge`
- `gitlab_group_badge`

In case a resource is deleted outside of terraform the tf state drift and we need to remove it if the read function received a `404`.

@armsnyder any ideas for how to test this in a non-hacky way?


armsnyder (approved):
lgtm. For testing, each TestStep has a `PreConfig` field you can use to run code like deleting the resource upstream.

timofurrer:
> For testing, each TestStep has a PreConfig field you can use to run code like deleting the resource upstream.

Ah yes, right! I think I'll address this later in #205

Issue #69, opened 2018-12-01
Support for label priority


alanbchristie:
Can support be added for label priorities? According to GitLab the priority of the label (an optional field) must be greater or equal than zero or null to remove the priority.

### Terraform Version
```
Terraform v0.11.10
+ provider.gitlab v1.0.0
```

### Affected Resource(s)
Please list the resources as a list, for example:
- gitlab_label

timofurrer:
Implementing in go-gitlab pending to be merged and released: 

* https://github.com/xanzy/go-gitlab/pull/1403

teekhandekar:
@alanbchristie Can I work on this ?

timofurrer:
@teekhandekar there is already ongoing work for that in https://github.com/gitlabhq/terraform-provider-gitlab/pull/926 (btw. you can check that in the `Development` section to the right of the issue description ;)).

The problem at the moment is that it's blocked because we can't support it properly without doing some very suboptimal changes in go-gitlab. Thus, until we have a way to correctly support it I guess we won't.

PR #926, opened 2022-03-07
resource/gitlab_label: Add priority attribute 

timofurrer:
Closes: #69

Blocked by:

* https://github.com/xanzy/go-gitlab/pull/1407

The PR above highlights the issue why this is blocked: basically go-gitlab doesn't support nullable fields in responses, which would be required for the `priority` field.

### DIFF
--- a/internal/provider/resource_gitlab_label.go
+++ b/internal/provider/resource_gitlab_label.go
@@ -78,7 +78,6 @@ func resourceGitlabLabelCreate(ctx context.Context, d *schema.ResourceData, meta
 	}
 
 	d.SetId(label.Name)
-
 	return resourceGitlabLabelRead(ctx, d, meta)
 }
 
--- a/internal/provider/resource_gitlab_label.go
+++ b/internal/provider/resource_gitlab_label.go
@@ -2,8 +2,10 @@ package provider
 
 import (
 	"context"
+	"encoding/json"
 	"log"
 
+	"github.com/hashicorp/go-retryablehttp"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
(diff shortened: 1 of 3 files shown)


internal/provider/resource_gitlab_label_test.go
```diff
@@ -52,36 +54,7 @@ func TestAccGitlabLabel_basic(t *testing.T) {
...
-				Check: resource.ComposeTestCheckFunc(
-					testAccCheckGitlabLabelExists("gitlab_label.fixme.20", &label),
-					testAccCheckGitlabLabelExists("gitlab_label.fixme.30", &label),
-					testAccCheckGitlabLabelExists("gitlab_label.fixme.40", &label),
-					testAccCheckGitlabLabelExists("gitlab_label.fixme.10", &label),
-					testAccCheckGitlabLabelAttributes(&label, &testAccGitlabLabelExpectedAttributes{
-						Name:        "FIXME11",
-						Color:       "#ff0000",
-						Description: "red label",
+						Priority:    5,
```
jtymes, line 57: Should there be a test validating the state is correct?

go.mod
```diff
@@ -11,7 +11,12 @@ require (
...
 	github.com/mitchellh/hashstructure v1.1.0
 	github.com/onsi/gomega v1.18.1
-	github.com/xanzy/go-gitlab v0.57.0
+	github.com/xanzy/go-gitlab v0.58.0
 	golang.org/x/crypto v0.0.0-20220214200702-86341886e292 // indirect
+	golang.org/x/net v0.0.0-20220225172249-27dd8689420f // indirect
+	golang.org/x/oauth2 v0.0.0-20220223155221-ee480838109b // indirect
+	golang.org/x/time v0.0.0-20220224211638-0e9765cccd65 // indirect
 	google.golang.org/api v0.34.0 // indirect
+	google.golang.org/appengine v1.6.7 // indirect
```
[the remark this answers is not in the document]
timofurrer: It was indirectly updated from `go-gitlab` ... I guess it's okay.

internal/provider/resource_gitlab_label.go
```diff
@@ -43,6 +44,12 @@ var _ = registerResource("gitlab_label", func() *schema.Resource {
 				Type:        schema.TypeString,
 				Optional:    true,
 			},
+			"priority": {
+				Description:      "The priority of the label. Must be greater or equal than zero or null to remove the priority.",
```
armsnyder, line 48: Did you test if it is possible to remove the priority once it has been set?

timofurrer: I've added a test case :)

armsnyder: The new test is comparing the value to `0`, which would be a non-null priority of value 0

PatrickRice-KSC:
@timofurrer  - I checked out the graphql API for labels to see if that helped out with this issue... there seems to be an [API to create labels](https://docs.gitlab.com/ee/api/graphql/reference/index.html#mutationlabelcreate), and there is an API to [read labels for projects](https://docs.gitlab.com/ee/api/graphql/reference/index.html#projectlabel) along with an API to [read labels for groups](https://docs.gitlab.com/ee/api/graphql/reference/index.html#grouplabel) however there appears to be no mutation for actually updating a label. So no love here for helping with this issue :-/

timofurrer:
> Hey @timofurrer - I wonder if we could apply the same workaround for go-gitlab here that we did over here: [#971 (comment)](https://github.com/gitlabhq/terraform-provider-gitlab/pull/971#discussion_r835801777)

@PatrickRice-KSC hmmm, where would you see the point to hook into the go-gitlab response handling? I guess what we could do is just query the API ourselves with a custom response struct ...

PatrickRice-KSC:
@timofurrer  - Oh shoot, my bad. It's in the _response_, not the _request_. I misread the note. Seems like we'd have to make a custom call at this point...

PR #929, opened 2022-03-07
New Resource `gitlab_system_hook`

timofurrer:
This change also required the `nightly` GitLab CE and EE images for the acceptance tests, because the feature is only available from 14.9 onwards.

@armsnyder do you think that's okay? I'd like to give it a shot and see how stable it is, we can always go back ...


docker-compose.yml
```diff
@@ -4,7 +4,7 @@ version: '3'
 # Only one of these services should be run at a time.
 services:
   gitlab-ce:
-    image: gitlab/gitlab-ce
+    image: gitlab/gitlab-ce:nightly
```
timofurrer, line 7: Okay, this doesn't work because the `/version` API doesn't return `-ee` and the `isRunningInEE` checks will fail.

```json
{"version":"14.9.0-pre","revision":"76141b84bb7"}
```

internal/provider/resource_gitlab_system_hook.go
```diff
@@ -0,0 +1,179 @@
...
+
+	hook, _, err := client.SystemHooks.AddHook(options, gitlab.WithContext(ctx))
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(fmt.Sprintf("%d", hook.ID))
+	if options.Token != nil {
+		d.Set("token", options.Token)
+	}
```
armsnyder, line 133: This shouldn't be needed. After all you set `options.Token` by getting the `"token"` attribute already.
```suggestion
```

armsnyder (changes requested):
looks great! one suggestion.

internal/provider/resource_gitlab_system_hook_test.go
```diff
@@ -0,0 +1,127 @@
...
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"token"},
+			},
+			// Update the hook to toggle all the values to their inverse
+			{
+				Config: testAccGitlabSystemHookUpdateConfig(rInt),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGitlabSystemHookExists("gitlab_system_hook.this", &hook),
+				),
```
PatrickRice-KSC, line 43: @timofurrer  - tiny nitpick, but since this will force the creation of a new resource, do we want to logically check that the "created_at" date changed in this test as opposed to line 28?

Otherwise this looks good to me!

timofurrer: hmmm, the reason I didn't is that from an acceptance test perspective I actually don't really care. Even if an update endpoint would be implemented one day, the essence of the feature still remains the same and so the tests.

On the other side it makes sense to test ALL the current behavior, but I'd mark this an implementation detail and I think we are good not asserting it in the acceptance tests.

PatrickRice-KSC: That's fair. I'd say this to good to approve then.

PatrickRice-KSC (approved):

nagyv (approved):
Approved following the approval from @PatrickRice-KSC

PR #934, opened 2022-03-07
Add `allowed_to_unprotect` and `unprotect_access_level` fields to branch protection

### LINKED ISSUE #878: Support `allowed_to_unprotect` field in branch protection resource
### Feature Description

The `gitlab_branch_protection` resource is missing the `allowed_to_unprotect` field.

The goal is to implement this to align with the upstream API.

While at it, it may also make sense to double check the branch protection data sources if they are aligned with upstream. (I quickly checked and I'm not sure if the upstream docs example is complete - it may make sense to quickly manually try what the response is.)

beekeep:
## Description

Closes #878.

Adds `unprotect_access_level` and `allowed_to_unprotect` attributes to the branch protection resource, which should bring the resource in line with the API.  Also changes `merge_access_level` and `push_access_level` to optional, as they were previously marked as required, but they are not required in the API.  A default is also configured, to match the default that is set via the API.

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/docs/resources/branch_protection.md
+++ b/docs/resources/branch_protection.md
@@ -4,7 +4,7 @@ page_title: "gitlab_branch_protection Resource - terraform-provider-gitlab"
 subcategory: ""
 description: |-
   The gitlab_branch_protection resource allows to manage the lifecycle of a protected branch of a repository.
-  ~> The allowedtopush, allowedtomerge and codeownerapproval_required attributes require a GitLab Enterprise instance.
+  ~> The allowed_to_push, allowed_to_merge, allowed_to_unprotect, unprotect_access_level and code_owner_approval_required attributes require a GitLab Enterprise instance.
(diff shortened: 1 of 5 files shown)


armsnyder (dismissed):
lgtm.

The test is a good candidate to be simplified using our testing helpers `testAccCurrentUsers` and `testAccCurrentGroups` in `helper_test.go`.

Hopefully this would also reduce the code in the `testAccCheckGitlabBranchProtectionAttributes` function, and if it doesn't, then probably one of our matching library dependencies like Gomega could be used there.

However since you were continuing the existing testing pattern this shouldn't hold up this PR. Thanks!

Adding a reference to #205 to help track this.

armsnyder (approved):

timofurrer (approved):

Issue #927, opened 2022-03-07
Implement gitlab_repository_file data source


petrmvala:
### Feature Description

It is in the Gitlab API and also in xanzy/go-gitlab library

### Do you want to implement this?

- [X] I would like to implement this myself 👷

Issue #820, opened 2022-01-30
Sporadic Failure in `TestAccGitlabLabel_basic` Acceptance Test


timofurrer:
During one of the CE acceptance test runs I've got this failure: https://github.com/gitlabhq/terraform-provider-gitlab/runs/4993701443?check_suite_focus=true#step:6:93

> === RUN   TestAccGitlabLabel_basic
>    resource_gitlab_label_test.go:17: Step 5/5 error: Error running apply: exit status 1
>       
>     Error: Provider produced inconsistent result after apply
>        
>       When applying changes to gitlab_label.fixme[79], provider
>       "provider[\"registry.terraform.io/hashicorp/gitlab\"]" produced an unexpected
>       new value: Root resource was present, but now absent.
>      
>       This is a bug in the provider, which should be reported in the provider's own
>       issue tracker.

timofurrer:
I haven't seen this ever since. @armsnyder did you? Otherwise I'd close this ...

timofurrer:
It still exists: https://github.com/gitlabhq/terraform-provider-gitlab/runs/5461795214?check_suite_focus=true#step:6:100

/cc @armsnyder @petrmvala

PR #939, opened 2022-03-08
feat: Add gitlab_repository_file data source

petrmvala:
This data source is part of Gitlab API.

Closes: #927

    - Create the resource
    - Update the attributes
    - Import the resource


timofurrer (commented):
Could you give this approach a shot:

https://github.com/gitlabhq/terraform-provider-gitlab/blob/main/internal/provider/data_source_gitlab_project_issue.go

For that to work you may also add the additional read-only fields to the `gitlab_repository_file` resource as `Computed`. 

If you don't want I can later or tomorrow give it a shot, no worries. 

Also, the examples are missing ;)

petrmvala:
Well, I posted this because I was not sure why does this error out and wanted to ask for advice https://github.com/gitlabhq/terraform-provider-gitlab/runs/5461795214?check_suite_focus=true#step:6:34

But the approach with the schema seems interesting.

timofurrer:
@petrmvala It's unrelated, but probably related to #820 ...

internal/provider/data_source_gitlab_repository_file.go
```diff
@@ -0,0 +1,107 @@
...
+	"log"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	gitlab "github.com/xanzy/go-gitlab"
+)
+
+var _ = registerDataSource("gitlab_repository_file", func() *schema.Resource {
+	return &schema.Resource{
+		Description: "Allows you to receive information about file in repository like name, size, content. File content is Base64 encoded. This endpoint can be accessed without authentication if the repository is publicly accessible.",
```
timofurrer, line 15: Please use the same description pattern as here:

https://github.com/gitlabhq/terraform-provider-gitlab/blob/c6a706e3ab6e907e2da0c92ee50138544ff501dd/internal/provider/data_source_gitlab_project_issue.go#L13

w.r.t to wording and the upstream docs link

internal/provider/data_source_gitlab_repository_file_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+		file := s.RootModule().Resources[src]
+		fileAttr := file.Primary.Attributes
+
+		search := s.RootModule().Resources[n]
+		searchAttr := search.Primary.Attributes
+
+		testAttributes := []string{
+			"project",
+			"file_path",
```
timofurrer, line 39:
```suggestion
```

timofurrer (changes requested):
Looks good, I think we leave the things with `datasourceSchemaFromResourceSchema()` for now and go with the proposed approach.

internal/provider/data_source_gitlab_repository_file.go
petrmvala: I will take a look again probably on Sunday. I had to actually implement something at my workplace :D

timofurrer:
@petrmvala I took the liberty and followed up on this.

I'll merge once the checks passed!

Thank you very much for the contribution! 🎉

timofurrer (approved):

PR #941, opened 2022-03-08
resource/gitlab_deploy_token: Implement pagination when reading tokens

timofurrer:
This fixes an issue where resources get recreated in case there are more
tokens in a group or project then the page size.

I'm also going to implement a `GET /projects/:id/deploy_tokens/:deploy_token` API upstream in GitLab ...


armsnyder (approved):
Looks great, thanks!

PR #945, opened 2022-03-09
Fix docs rendering issue

timofurrer:
This change removes this colon rendering issue:

image

### DIFF
--- a/internal/provider/resource_gitlab_project_issue.go
+++ b/internal/provider/resource_gitlab_project_issue.go
@@ -23,7 +23,7 @@ var _ = registerResource("gitlab_project_issue", func() *schema.Resource {
 
 -> During a terraform destroy this resource will close the issue. Set the delete_on_destroy flag to true to delete the issue instead of closing it.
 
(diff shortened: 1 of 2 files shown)


internal/provider/resource_gitlab_project_issue.go
```diff
@@ -23,7 +23,7 @@ var _ = registerResource("gitlab_project_issue", func() *schema.Resource {
 
 -> During a terraform destroy this resource will close the issue. Set the delete_on_destroy flag to true to delete the issue instead of closing it.
 
-~> **Experimental**: while the base functionality of this resource works, it may be subject to minor change.
+~> **Experimental** while the base functionality of this resource works, it may be subject to minor change.
```
armsnyder, line 26: Screen Shot 2022-03-09 at 8 08 13 PM

```suggestion
~> **Experimental** While the base functionality of this resource works, it may be subject to minor change.
```

armsnyder (changes requested):

armsnyder (approved):

PR #952, opened 2022-03-13
Update managed_license resources to deprecate existing values and add GitLab 15+ values

### LINKED ISSUE #951: Update the "managed_license" resource to follow the new conventions necessary for 15.0
### Feature Description

Per the managed license API documentation, the current `approval` and `blacklisted` verbiage is now deprecated and will be removed in 15.0 in favor of `allowed` and `denied`. 

This issue will update the allowed values in the terraform provider to accept these values in addition to the current `approval` and `blacklisted` values, and we'll likely want to open a second PR in may that removes the existing values.

PatrickRice-KSC:
## Description

Closes #951

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/internal/provider/helper_test.go
+++ b/internal/provider/helper_test.go
@@ -15,6 +15,8 @@ import (
 	"github.com/xanzy/go-gitlab"
 )
 
+type SkipFunc = func() (bool, error)
+
 func init() {
 	// We are using the gomega package for its matchers only, but it requires us to register a handler anyway.
 	gomega.RegisterFailHandler(func(_ string, _ ...int) {
@@ -68,6 +70,22 @@ func testAccCheck(t *testing.T) {
 
 	if os.Getenv(resource.EnvTfAcc) == "" {
 		t.Skip(fmt.Sprintf("Acceptance tests skipped unless env '%s' set", resource.EnvTfAcc))
+	}
+}
+
+// orSkipFunc accepts many skipFunc and returns "true" if any returns true.
+func orSkipFunc(input ...SkipFunc) SkipFunc {
+	return func() (bool, error) {
+		for _, item := range input {
+			result, err := item()
+			if err != nil {
+				return false, err
+			}
+			if result {
+				return result, nil
+			}
+		}
+		return false, nil
 	}
 }
 
--- a/internal/provider/resource_gitlab_managed_license.go
+++ b/internal/provider/resource_gitlab_managed_license.go
@@ -12,6 +12,10 @@ import (
 	"github.com/xanzy/go-gitlab"
 )
 
+var managedLicenseAllowedValues = []string{
+	"approved", "blacklisted", "allowed", "denied",
+}
+
 var _ = registerResource("gitlab_managed_license", func() *schema.Resource {
 	return &schema.Resource{
 		Description: `The ` + "`" + `gitlab_managed_license` + "`" + ` resource allows to manage the lifecycle of a managed license.
@@ -44,11 +48,14 @@ var _ = registerResource("gitlab_managed_license", func() *schema.Resource {
 				Description: "The name of the managed license (I.e., 'Apache License 2.0' or 'MIT license')",
 			},
 			"approval_status": {
-				Type:         schema.TypeString,
-				Required:     true,
-				ForceNew:     false,
-				ValidateFunc: validation.StringInSlice([]string{"approved", "blacklisted"}, true),
-				Description:  "Whether the license is approved or not. Only 'approved' or 'blacklisted' allowed.",
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         false,
+				ValidateFunc:     validation.StringInSlice(managedLicenseAllowedValues, true),
+				DiffSuppressFunc: checkDeprecatedValuesForDiff,
+				Description: fmt.Sprintf(`The approval status of the license. Valid values are: %s. "approved" and "blacklisted" 
+				have been deprecated in favor of "allowed" and "denied"; use "allowed" and "denied" for GitLab versions 15.0 and higher. 
+				Prior to version 15.0 and after 14.6, the values are equivalent.`, renderValueListForDocs(managedLicenseAllowedValues)),
 			},
 		},
 	}
@@ -149,6 +156,11 @@ func stringToApprovalStatus(s string) *gitlab.LicenseApprovalStatusValue {
 	lookup := map[string]gitlab.LicenseApprovalStatusValue{
 		"approved":    gitlab.LicenseApproved,
 		"blacklisted": gitlab.LicenseBlacklisted,
+
+		// This is counter-intuitive, but currently the API response from the non-deprecated
+		// values is the deprecated values. So we have to map them here.
+		"allowed": gitlab.LicenseApproved,
+		"denied":  gitlab.LicenseBlacklisted,
 	}
 
 	value, ok := lookup[s]
@@ -170,4 +182,21 @@ func projectIdAndLicenseIdFromId(id string) (string, int, error) {
 	}
 
 	return projectId, licenseId, nil
+}
+
+func checkDeprecatedValuesForDiff(k, oldValue, newValue string, d *schema.ResourceData) bool {
+	approvedValues := []string{"approved", "allowed"}
+	deniedValues := []string{"blacklisted", "denied"}
+
+	// While we could technically combine these two "if" blocks, this seems more readable
+	// and should have the same execution pattern.
+	if contains(approvedValues, oldValue) && contains(approvedValues, newValue) {
+		return true
+	}
+
+	if contains(deniedValues, oldValue) && contains(deniedValues, newValue) {
+		return true
+	}
+
+	return false
 }
--- a/internal/provider/resource_gitlab_managed_license_test.go
+++ b/internal/provider/resource_gitlab_managed_license_test.go
@@ -21,19 +21,20 @@ func TestAccGitlabManagedLicense_basic(t *testing.T) {
 		CheckDestroy:      testAccCheckManagedLicenseDestroy,
 		Steps: []resource.TestStep{
 			{
-				// Create a managed license with an "approved" state
+				// Create a managed license with an "allowed" state
 				SkipFunc: isRunningInCE,
-				Config:   testManagedLicenseConfig(rInt, "approved"),
+				Config:   testManagedLicenseConfig(rInt, "allowed"),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckGitlabManagedLicenseExists("gitlab_managed_license.fixme", &managedLicense),
 				),
 			},
 			{
-				// Update the managed license to have a blacklisted state
+				// Update the managed license to have a denied state
 				SkipFunc: isRunningInCE,
-				Config:   testManagedLicenseConfig(rInt, "blacklisted"),
+				Config:   testManagedLicenseConfig(rInt, "denied"),
 				Check: resource.ComposeTestCheckFunc(
-					testAccCheckGitlabManagedLicenseStatus("gitlab_managed_license.fixme", &managedLicense),
+					//Even though the API accepts "denied", it still returns "blacklisted" until 15.0
+					testAccCheckGitlabManagedLicenseStatus("gitlab_managed_license.fixme", "blacklisted", &managedLicense),
 				),
 			},
 			{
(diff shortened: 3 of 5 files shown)


PatrickRice-KSC:
Note - I chose to leave the existing values under a deprecated test, which will automatically stop running once we update to GitLab 15+. This ensures that we don't accidentally regress in the meantime.

PatrickRice-KSC:
Hm, jumped the gun a bit on this PR I think. Was just testing this on GitLab.com, and it appears the new values are not accepted yet, so folks will have a "hard" cut-over in version 15. That's a bit annoying. I'll likely close this PR and then re-push it post 15.0 (May).

PatrickRice-KSC:
And nevermind once again. Arg. I was testing with an older project and having issues with my tests. New approval statuses are implemented, so this PR is valid.

PatrickRice-KSC:
Since I don't have rights to re-open a PR, can someone reopen this so that I don't duplicate PRs in the repo and clutter it up? This should be ready for testing.

internal/provider/resource_gitlab_managed_license_test.go
```diff
@@ -144,3 +180,25 @@ resource "gitlab_managed_license" "fixme" {
 }
 	`, rInt, status)
 }
+
+// This method is used as a skipfunc to ensure we're both running in EE, and the EE
+// version is less than 15, since existing values were deprecated in v15.
+func checkCEAndDeprecatedVersion() (bool, error) {
```
armsnyder, line 186: We have a generic function similar to this already:

https://github.com/gitlabhq/terraform-provider-gitlab/blob/49db985b79f667da9244a57d9ff4f107f337b506/internal/provider/util.go#L227

Could you do something like this for the `SkipFunc` instead?

```golang
SkipFunc: orSkipFuncs(isRunningInCE, isGitLabVersionLessThan(testGitlabClient, "15.0")
```

(`orSkipFuncs` doesn't exist; you'd need to write that)

internal/provider/resource_gitlab_managed_license.go
```diff
@@ -47,8 +47,8 @@ var _ = registerResource("gitlab_managed_license", func() *schema.Resource {
 				Type:         schema.TypeString,
 				Required:     true,
 				ForceNew:     false,
-				ValidateFunc: validation.StringInSlice([]string{"approved", "blacklisted"}, true),
-				Description:  "Whether the license is approved or not. Only 'approved' or 'blacklisted' allowed.",
+				ValidateFunc: validation.StringInSlice([]string{"approved", "blacklisted", "allowed", "denied"}, true),
+				Description:  "Whether the license is approved or not. 'allowed' or 'denied' are acceptable values. 'approved' and 'blacklist' are accepted, but removed in GitLab 15.0",
```
timofurrer, line 51: Please use this function:

https://github.com/gitlabhq/terraform-provider-gitlab/blob/49db985b79f667da9244a57d9ff4f107f337b506/internal/provider/util.go#L17

To keep the docs and validation in sync. I also suggest that you use the same wording for docs as other users of that function.

timofurrer (changes requested):

internal/provider/resource_gitlab_managed_license_test.go
PatrickRice-KSC: @armsnyder  - For sure. Which of these do you think is better?

`func orSkipFun(first func() (bool, error), second func() (bool, error)) func() (bool error)`
or 
`orSkipFunc(functions []func() (bool, error)) func() (bool, error)`

The first one is going to result in code within the func itself that's a lot more readable, and a skipFunc declaration that matches what your review finding is suggesting, but it would only ever accept 2 functions so if we wanted to do a 3-deep "or" chain we'd need to chain the "orSkipFunc" method twice. The second option would allow for many skipFuncs, but the skipFunc declaration in the test itself would need to be something like:

```
SkipFunc: orSkipFunc([]func() (bool, error){
	isRunningInCE,
	isGitLabVersionLessThan(testGitlabClient, "15"),
}),
```

I'm leaning to the first option, just because we haven't needed an "OR" function to-date, so the likelihood of needing one 3-deep is pretty slim, so I favor the readability for now. Thoughts?

internal/provider/resource_gitlab_managed_license.go
PatrickRice-KSC: Yep, I'll get that done at the same time as I fix up what armsnyder is asking for.

internal/provider/resource_gitlab_managed_license_test.go
timofurrer: What about something like this:

```go
func orSkipFunc(skipFuncs ...func() (bool, error)) (bool, error) {
	for _, skipFunc := range skipFuncs {
		skip, err := skipFunc()
		if err != nil {
			return false, err
		}
		if skip {
			return true, nil
		}
	}
	return false, nil
}
```

... then call it like this:

```go
SkipFunc: orSkipFunc(isRunningInCE, isGitLabVersionLessThan(testGitlabClient, "15"))
```

armsnyder: Getting closer... the func needs to return a func though. Combining both of your suggestions...

```golang
func orSkipFunc(skipFuncs ...func() (bool, error)) func() (bool, error) {
	switch len(skipFuncs) {
	case 0:
		panic("zero arguments")
	case 1:
		return skipFuncs[0]
	case 2:
		return func() (bool, error) {
			v0, err := skipFuncs[0]()
			if err != nil {
				return false, err
			}
			v1, err := skipFuncs[1]()
			if err != nil {
				return false, err
			}
			return v0 || v1, nil
		}
	default:
		return orSkipFunc(skipFuncs[0], orSkipFunc(skipFuncs[1:]...))
	}
}
```

armsnyder: Getting closer... the func needs to return a func though. Combining both of your suggestions...

```golang
type SkipFunc = func() (bool, error)

func orSkipFunc(a SkipFunc, b ...SkipFunc) SkipFunc {
	switch len(b) {
	case 0:
		return a
	case 1:
		return func() (bool, error) {
			va, err := a()
			if err != nil {
				return false, err
			}
			vb, err := b[0]()
			if err != nil {
				return false, err
			}
			return va || vb, nil
		}
	default:
		return orSkipFunc(a, orSkipFunc(b[0], b[1:]...))
	}
}
```

PatrickRice-KSC: Perfect, I actually implemented the "only 2" option in the last commit that has almost explicitly what you have in "case 1" implemented, but I'll update the approach. I really like defining a specific SkipFunc type, makes the code _way_ easier to read.

Just out of curiosity, is there a reason to use recursion as opposed to just using range to loop through the SkipFunc and short-circuit returning on the first "true" or error value?

internal/provider/resource_gitlab_managed_license.go
PatrickRice-KSC: Sorry, should have left a comment, but this was fixed in the last commit.

internal/provider/resource_gitlab_managed_license_test.go
timofurrer: I'd say personal preference and background :)

internal/provider/helper_test.go
```diff
@@ -71,6 +73,22 @@ func testAccCheck(t *testing.T) {
 	}
 }
 
+// Helper Function that accepts two skipFunc and returns "true" if either returns true.
+func orSkipFunc(input ...SkipFunc) func() (bool, error) {
```
timofurrer, line 77:
```suggestion
// orSkipFunc accepts many skipFunc and returns "true" if any returns true.
func orSkipFunc(input ...SkipFunc) SkipFunc {
```

timofurrer (changes requested):
looks good, except a minor comment :)

internal/provider/helper_test.go
PatrickRice-KSC: Of _course_ I forgot to update the comment 🤦‍♂️ .

internal/provider/resource_gitlab_managed_license_test.go
PatrickRice-KSC: Fair enough. I tend to prefer avoiding recursion, so I committed the one I'd written above. Feel free to leave comments if we don't like it 😄

PatrickRice-KSC:
So weird thing about the EE failure - this is happening because the "Read" API always returns the old deprecated results no matter which value is passed in. You pass in "allowed", you get back "approved" with a success. You pass in "denied", you get back "blacklisted" with a success. You can then update them interchangeably, but you'll always get back the old value.

I see a couple different ways of fixing this and I'm interested in your guys opinions:

- Map the values in the "Read" function based on what's pass into the resource config. I.e., if "approved" is received, and "allowed" is in the config, store "allowed" in the resource returned. I don't love this because we're manipulating data coming back from the API, but it's cleanest way of making the response idempotent between runs
- Ignore the change in the test. I don't love this because it means we're essentially ignoring test failures
- Change the non-deprecated test to only run on  GitLab 15+, and the deprecated test to run on GitLab 15-. I don't love this because it means there is essentially an "untested" path in the code.

It seems like option 1 may be the "best" path, and it's relatively easy to do. We'd have to leave it in the codebase for a while though, since there may be users who use the tfprovider on earlier code versions, so we'd have to pre-emptively deprecate it for 4.0 I think.

timofurrer:
@PatrickRice-KSC that's a pity -.^ Do you know if there exists an upstream issue for that?

Actually there is another way (and probably the cleanest one), which is to use a [`DiffSupressFunc`](https://www.terraform.io/plugin/sdkv2/schemas/schema-behaviors#diffsuppressfunc) and treat those values as the "same" ...

However, I'm not sure what would happen if you'd switch from the deprecated value to the new value, if you can make it so that the API is still being called ...

I suggest you play around with it a little bit and see if we can use it :)

PatrickRice-KSC:
> that's a pity -.^ Do you know if there exists an upstream issue for that?

As far as I can tell, this is actually how they intended this to work. It's noted this way in the deprecation notice after I went back and read it a second time (after observing the behavior): https://docs.gitlab.com/ee/update/deprecations.html#legacy-approval-status-names-from-license-compliance-api

> Actually there is another way (and probably the cleanest one), which is to use a [DiffSupressFunc](https://www.terraform.io/plugin/sdkv2/schemas/schema-behaviors#diffsuppressfunc) and treat those values as the "same" ...

Sweet, I'll take a look. I'll definitely need to play around with that a bit since I've not needed to use it before.

PatrickRice-KSC:
I think this will be pretty simple. I'm going to assume that in 15.0 that GitLab will automatically migrate everything from "allowed" to "approved" and from "blacklisted" to "denied", just because there is no way they'll make everyone re-submit all their licenses. 

So if I set "allowed == approved", then it'll suppress the diff the second time around (and make the tests pass). The only odd thing here is that if a user has a config that was applied pre-15.0, it'll pass until they attempt to add a new license, then that new license will fail with an invalid value and force them to update the verbiage to "approved" for all licenses which will ignore applying for all old licenses, and just apply for the new licenses.  Thus, I don't think we need to make it call the API even if they're attempting to update from "allowed -> approved".

Testing and committing.

internal/provider/resource_gitlab_managed_license.go
```diff
@@ -171,3 +181,20 @@ func projectIdAndLicenseIdFromId(id string) (string, int, error) {
 
 	return projectId, licenseId, nil
 }
+
+func checkDeprecatedValuesForDiff(k, oldValue, newValue string, d *schema.ResourceData) bool {
+	approvedValues := []string{"approved", "allowed"}
+	deniedValues := []string{"blacklisted", "denied"}
+
+	//While we could technically combine these two "if" blocks, this seems more readable
```
timofurrer, line 189:
```suggestion
	// While we could technically combine these two "if" blocks, this seems more readable
```

I love the comment - well, the prio for readability :D

internal/provider/resource_gitlab_managed_license.go
```diff
@@ -149,6 +154,11 @@ func stringToApprovalStatus(s string) *gitlab.LicenseApprovalStatusValue {
 	lookup := map[string]gitlab.LicenseApprovalStatusValue{
 		"approved":    gitlab.LicenseApproved,
 		"blacklisted": gitlab.LicenseBlacklisted,
+
+		// This is counter-intuitive, but currently the API response from the non-deprecated
+		//  values is the deprecated values. So we have to map them here.
```
timofurrer, line 159:
```suggestion
		// values is the deprecated values. So we have to map them here.
```

timofurrer (changes requested):

internal/provider/resource_gitlab_managed_license.go
```diff
@@ -44,11 +48,12 @@ var _ = registerResource("gitlab_managed_license", func() *schema.Resource {
...
-				Required:     true,
-				ForceNew:     false,
-				ValidateFunc: validation.StringInSlice([]string{"approved", "blacklisted"}, true),
-				Description:  "Whether the license is approved or not. Only 'approved' or 'blacklisted' allowed.",
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         false,
+				ValidateFunc:     validation.StringInSlice(managedLicenseAllowedValues, true),
+				DiffSuppressFunc: checkDeprecatedValuesForDiff,
+				Description:      fmt.Sprintf("The approval status of the license. Valid values are: %s", renderValueListForDocs(managedLicenseAllowedValues)),
```
timofurrer, line 56: Should we document that the valid values are consisting of deprecated and new values and that some are "treated" as equivalent?

PatrickRice-KSC: Sure - I added a note to the end stating the equivalence and that "allowed" and "denied" should be favored since the others will go away in 15.0. I also added a note about when "allowed" and "denied" were added in the event someone is still using the tf provider on older versions of self-hosted.

PatrickRice-KSC:
Once we're done with reviews, let me know please and I'll squash the commits. I'm not a fan of 9+ commits on a PR 😄

timofurrer (dismissed):
LGTM 🎉 

You can do the squashing and once tests passed I'll merge :)

Nice work 🌮

PatrickRice-KSC:
Done - I rebased as long as I was in there as well, so we should be good when the tests finish.

timofurrer (approved):

PR #954, opened 2022-03-13
Use specific GET API instead of list for labels

timofurrer:
Removing some tech debt to use specific resources GET API instead of the list API...

Also implementing the missing `gitlab_label` importer (#36) and removing some unnecessary large testing configurations.

### DIFF
--- a/internal/provider/resource_gitlab_group_label.go
+++ b/internal/provider/resource_gitlab_group_label.go
@@ -72,7 +72,6 @@ func resourceGitlabGroupLabelCreate(ctx context.Context, d *schema.ResourceData,
 	}
 
 	d.SetId(label.Name)
-
 	return resourceGitlabGroupLabelRead(ctx, d, meta)
 }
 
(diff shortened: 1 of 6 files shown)


internal/provider/resource_gitlab_group_label.go
```diff
@@ -72,7 +72,6 @@ func resourceGitlabGroupLabelCreate(ctx context.Context, d *schema.ResourceData,
 	}
 
 	d.SetId(label.Name)
```
jtymes, line 74: Not quite related to this PR, but I wonder if this could be problematic using the `Name` instead of the `ID` if someone were to rename the label.

internal/provider/resource_gitlab_label.go
```diff
@@ -20,6 +22,12 @@ var _ = registerResource("gitlab_label", func() *schema.Resource {
 		ReadContext:   resourceGitlabLabelRead,
 		UpdateContext: resourceGitlabLabelUpdate,
 		DeleteContext: resourceGitlabLabelDelete,
+		// FIXME: this importer sucks a little, but we can't use a passthrough importer, because
+		//        the resource id is flawed and we don't want to break backwards-compatibility.
```
jtymes, line 26: Looks like my earlier comment is something you're already aware of 😄

timofurrer: Yeah, exactly - it's actually not the only resource -.^ ... It's something for v4 of this provider to change ..

jtymes (approved):

armsnyder (approved):

PR #960, opened 2022-03-14
Implement all missing importers

timofurrer:
This change set together with (https://github.com/gitlabhq/terraform-provider-gitlab/pull/954) implements all missing importers and therefore closes the long open #36 .

I've focused on the importer themselves and not so much about the underlying ids. 
I'll follow up with an issue describing the current state of resource ids currently in use. I hope that we can streamline those for v4 of the provider ...

### DIFF
--- a/internal/provider/resource_gitlab_deploy_token.go
+++ b/internal/provider/resource_gitlab_deploy_token.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"net/http"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
@@ -15,7 +16,6 @@ import (
 )
 
 var _ = registerResource("gitlab_deploy_token", func() *schema.Resource {
-	// lintignore: XR002 // TODO: Resolve this tfproviderlint issue
 	return &schema.Resource{
 		Description: `The ` + "`gitlab_deploy_token`" + ` resource allows to manage the lifecycle of group and project deploy tokens.
 
@@ -24,6 +24,9 @@ var _ = registerResource("gitlab_deploy_token", func() *schema.Resource {
 		CreateContext: resourceGitlabDeployTokenCreate,
 		ReadContext:   resourceGitlabDeployTokenRead,
 		DeleteContext: resourceGitlabDeployTokenDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceGitlabDeployTokenStateImporter,
+		},
 
 		Schema: map[string]*schema.Schema{
 			"project": {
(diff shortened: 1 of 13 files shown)


internal/provider/resource_gitlab_deploy_token.go
```diff
@@ -257,3 +260,25 @@ func resourceGitlabDeployTokenDelete(ctx context.Context, d *schema.ResourceData
 
 	return nil
 }
+
+func resourceGitlabDeployTokenStateImporter(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	s := strings.Split(d.Id(), ":")
+	if len(s) != 3 {
+		d.SetId("")
+		return nil, fmt.Errorf("Invalid Deploy Token import format; expected '{type}:{type_id}:{deploy_key_id}' whereas the type can either be a group or project")
```
jtymes, line 268:
```suggestion
		return nil, fmt.Errorf("Invalid Deploy Token import format; expected '{type}:{type_id}:{deploy_token_id}' where the type can either be a group or project")
```

internal/provider/resource_gitlab_deploy_token.go
```diff
@@ -257,3 +260,25 @@ func resourceGitlabDeployTokenDelete(ctx context.Context, d *schema.ResourceData
...
+
+	d.SetId(id)
+	switch typ {
+	case "project":
+		d.Set("project", type_id)
+	case "group":
+		d.Set("group", type_id)
+	default:
+		d.SetId("")
+		return nil, fmt.Errorf("Invalid Deploy Token import format; expected '{type}:{type_id}:{deploy_key_id}' whereas the type can either be a group or project")
```
jtymes, line 281:
```suggestion
		return nil, fmt.Errorf("Invalid Deploy Token import format; expected '{type}:{type_id}:{deploy_token_id}' where the type can either be a group or project")
```

internal/provider/resource_gitlab_deploy_token.go
```diff
@@ -257,3 +260,25 @@ func resourceGitlabDeployTokenDelete(ctx context.Context, d *schema.ResourceData
...
 }
+
+func resourceGitlabDeployTokenStateImporter(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	s := strings.Split(d.Id(), ":")
+	if len(s) != 3 {
+		d.SetId("")
+		return nil, fmt.Errorf("Invalid Deploy Token import format; expected '{type}:{type_id}:{deploy_key_id}' whereas the type can either be a group or project")
+	}
+
+	typ, type_id, id := s[0], s[1], s[2]
```
jtymes, line 271:
```suggestion
	token_type, type_id, id := s[0], s[1], s[2]
```

jtymes (changes requested):

internal/provider/resource_gitlab_deploy_token.go
jtymes: Thinking about this more, it seems like it should be `tokenType` and `typeID`.

timofurrer:
@jtymes good catches, thanks for the review!

armsnyder (approved):

PR #964, opened 2022-03-16
resource/gitlab_repository_file: Mitigate parallelism limitation 

### LINKED ISSUE #940: Remove parallelism limitation in `gitlab_repository_file` resource
At the moment we have the limitation that if a terraform module / run contains more than one `gitlab_repository_file` resource the entire terraform `apply` / `destroy` has to be run with `-parallelism=1` for reliable success:

https://github.com/gitlabhq/terraform-provider-gitlab/blob/b1521811a1e99391801e69b19287565a4918257d/internal/provider/resource_gitlab_repository_file.go#L21

If it's run in parallel you'll experience something like this:

```
Error: somefile.yaml failed to delete repository file: 
(400 Bad Request) DELETE https://gitlab.com/api/v4/projects/30716/repository/files/vsomefile.yaml: 400 
{message: 9:Could not update refs/heads/master. Please refresh and try again..}
```

The root cause cannot be fixed in the provider, because the GitLab upstream API has this limitation that parallel changes to the repository are not allowed.

Any clever ideas on how to mitigate / remove this limitations? Maybe retries with a timeout and random wait time in between for that particular error? 

/cc @armsnyder

timofurrer:
This change set implements the following two measures to mitigate / remove the parallelism issues we were having as described in #940:

1. only allow a single `gitlab_repository_file` resource to make an API call (no matter to which project or file)
1. retry API calls in case of a refresh error

The first mitigates that resources within an apply / destroy are queued and the second mitigates the case when something else outside of terraform changes the repository state. Both combined are good mitigations for the issue at hand.

The drawback of this change is that it'll slow down certain terraform configurations - necessarily and unnecessarily in cases were a files in different repositories / projects are changed, but still queued. I think that's a limitation I'm willing to take for the benefits it provides.

Closes #940

### DIFF
--- a/docs/resources/repository_file.md
+++ b/docs/resources/repository_file.md
@@ -4,12 +4,24 @@ page_title: "gitlab_repository_file Resource - terraform-provider-gitlab"
 subcategory: ""
 description: |-
   The gitlab_repository_file resource allows to manage the lifecycle of a file within a repository.
+  -> Timeouts Default timeout for Create, Update and Delete is one minute and can be configured in the timeouts block.
+  -> Implementation Detail GitLab is unable to handle concurrent calls to the GitLab repository files API for the same project.
+     Therefore, this resource queues every call to the repository files API no matter of the project, which may slow down the terraform
+     execution time for some configurations. In addition, retries are performed in case a refresh is required because another application
+     changed the repository at the same time.
   Upstream API: GitLab REST API docs https://docs.gitlab.com/ee/api/repository_files.html
 ---
 
 # gitlab_repository_file (Resource)
 
 The `gitlab_repository_file` resource allows to manage the lifecycle of a file within a repository.
+
+-> **Timeouts** Default timeout for *Create*, *Update* and *Delete* is one minute and can be configured in the `timeouts` block.
+
+-> **Implementation Detail** GitLab is unable to handle concurrent calls to the GitLab repository files API for the same project.
+   Therefore, this resource queues every call to the repository files API no matter of the project, which may slow down the terraform
+   execution time for some configurations. In addition, retries are performed in case a refresh is required because another application
+   changed the repository at the same time.
 
 **Upstream API**: [GitLab REST API docs](https://docs.gitlab.com/ee/api/repository_files.html)
 
@@ -54,6 +66,7 @@ resource "gitlab_repository_file" "this" {
 - `author_name` (String) Name of the commit author.
 - `id` (String) The ID of this resource.
 - `start_branch` (String) Name of the branch to start the new commit from.
+- `timeouts` (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
 
 ### Read-Only
 
@@ -65,6 +78,15 @@ resource "gitlab_repository_file" "this" {
 - `last_commit_id` (String) The last known commit id.
 - `ref` (String) The name of branch, tag or commit.
 - `size` (Number) The file size.
+
+<a id="nestedblock--timeouts"></a>
+### Nested Schema for `timeouts`
+
+Optional:
+
+- `create` (String)
+- `delete` (String)
+- `update` (String)
 
 ## Import
 
--- a/docs/resources/repository_file.md
+++ b/docs/resources/repository_file.md
@@ -4,15 +4,12 @@ page_title: "gitlab_repository_file Resource - terraform-provider-gitlab"
 subcategory: ""
 description: |-
   The gitlab_repository_file resource allows to manage the lifecycle of a file within a repository.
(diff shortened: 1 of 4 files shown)


jtymes (changes requested):

jtymes (dismissed):

internal/provider/resource_gitlab_repository_file.go
```diff
@@ -89,10 +117,16 @@ var _ = registerResource("gitlab_repository_file", func() *schema.Resource {
 })
 
 func resourceGitlabRepositoryFileCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	client := meta.(*gitlab.Client)
 	project := d.Get("project").(string)
 	filePath := d.Get("file_path").(string)
 
+	log.Printf("[DEBUG] gitlab_repository_file: waiting for lock to create %s/%s", project, filePath)
+	apiSync.Lock()
```
armsnyder, line 124: One limitation of using `sync.Mutex` here is that it will block "forever". Since the caller passes a `context.Context`, it expects that you'll return when the context is cancelled.

One idea would be to use a channel as a lock.

```go

type lock chan struct{}

func newLock() lock {
	// The buffer size of one allows us to use it as a lock.
	return make(lock, 1)
}

func (c lock) lock(ctx context.Context) error {
	select {
	case c <- struct{}{}:
		// Lock acquired.
		return nil
	case <-ctx.Done():
		// Timeout.
		return ctx.Err()
	}
}

func (c lock) unlock() {
	<-c
}

var resourceGitlabRepositoryFileLock = newLock()

func resourceGitlabRepositoryFileCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
	if err := resourceGitlabRepositoryFileLock.lock(ctx); err != nil {
		return diag.FromErr(err)
	}
	defer resourceGitlabRepositoryFileLock.unlock()
}
```

armsnyder (changes requested):

internal/provider/resource_gitlab_repository_file.go
timofurrer: I actually was thinking of something like this ...

> One limitation of using sync.Mutex here is that it will block "forever"

... however, in which case could that possibly happen for us? Given that the GitLab calls (which may block) are using the passed ctx?

I can implement it as kind of a "defensive" thing if you think it makes sense.

armsnyder: You are correct that there cannot be deadlocks. By ""forever"" I mean that it will block for longer than the caller might expect, because we are not respecting the context.

As an exaggerated example, suppose there are two `gitlab_repository_file` resources with different timeouts.

```terraform
resource "gitlab_repository_file" "foo" {
  timeouts {
    create = "60m"
  }
}

resource "gitlab_repository_file" "bar" {
  timeouts {
    create = "5m"
  }
}
```

If `"foo"` gets the lock first and blocks, then `"bar"` will also timeout after `60m` instead of the intended `5m`.

I would be ok with keeping the `sync.Mutex` if it as a temporary hack. Otherwise I prefer the channel lock.

For me, generally I only use `sync.Mutex` to protect data structures like maps when I am not doing I/O while holding the lock.

docs/resources/repository_file.md
```diff
@@ -4,15 +4,24 @@ page_title: "gitlab_repository_file Resource - terraform-provider-gitlab"
...
+-> Timeouts Default timeouts for *Create*, *Update* and *Delete* is one minute and can be configured in the `timeouts` block.
+
+-> Implementation Detail GitLab is unable to handle concurrent calls to the GitLab repository files API for the same project.
+   Therefore, this resource queues every call to the repository files API no matter of the project, which may slow down the terraform
+   execution time for some configurations. In addition, retries are performed in case a refresh is required because another application
+   changed the repository at the same time.
```
armsnyder, line 24: These callouts can be formatted better.

Try them out inhttps://registry.terraform.io/tools/doc-preview

They should be like this:

```md
-> **Implementation Detail** GitLab is unable...
```

internal/provider/resource_gitlab_repository_file.go
timofurrer: Ah, yes makes sense. I'll adapt it tomorrow :)

docs/resources/repository_file.md
timofurrer: omg, how did I just learn about that tool :O 🎉

internal/provider/resource_gitlab_repository_file.go
timofurrer: Alright, looks good now :)

docs/resources/repository_file.md
armsnyder: This doesn't look resolved

timofurrer: 😕  don't know what happened here ... rebase gone wrong or something. I've pushed the update :)

armsnyder (approved):

Issue #966, opened 2022-03-19
Support true topic deletion


jtymes:
### Feature Description

Topic deletion landed in [14.9](https://docs.gitlab.com/ee/api/topics.html#delete-a-project-topic).

@timofurrer I'm assuming you'll want this one since you implemented it all the way so far, but I'm also willing to give it a try. 

### Do you want to implement this?

- [ ] I would like to implement this myself 👷

timofurrer:
I actually have that already implemented on a branch 😅

PR #967, opened 2022-03-19
resource/gitlab_topic: support proper deletion of topics

timofurrer:
This is blocked by the GitLab 14.9 release :)

Closes: #966


internal/provider/resource_gitlab_topic.go
```diff
@@ -125,27 +126,43 @@ func resourceGitlabTopicUpdate(ctx context.Context, d *schema.ResourceData, meta
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	if !softDestroy && deleteNotSupported {
+		return diag.Errorf("GitLab 14.9 introduced the proper deletion of topics. Set `soft_destroy = true` to empty out a topic instead of deleting it.")
 	}
 
-	log.Printf("[WARN] Not deleting gitlab topic %s. Instead emptying its description", d.Id())
+	// NOTE: the `soft_destroy` field is deprecated and will be removed in a future version.
+	//       It was only introduced because GitLab prio to 14.9 didn't support topic deletion.
```
jtymes, line 145: nit that doesn't need to block:
```suggestion
	//       It was only introduced because GitLab prior to 14.9 didn't support topic deletion.
```

jtymes (approved):

internal/provider/resource_gitlab_topic.go
timofurrer: good catch, thanks!

jtymes (dismissed):

beekeep (approved):
Changes look good!

PR #968, opened 2022-03-19
resource/gitlab_topic: Support avatars

timofurrer:
This change set introduces avatar support for topics.

To reliably detect changes in the avatar the avatar is hashed into the internal `_avatar_hash` attribute and a custom diff function is used to detect changes in the avatar if the filename stays the same.

I thought about comparing against the image returned from `avatar_url`, however I doubt that this would be reliable enough because the image might be changed slightly - e.g. cropped or compressed - especially when gravatar is used.

@armsnyder wdyt about this solution?


timofurrer:
@PatrickRice-KSC I have to dig into how that really works for the `s3_bucket_object` resource. To me, it was always cumbersome to provide the hash myself - "thinking why doesn't it do that its own?". However, I agree that there is some little maintenance overhead and the `_avatar_hash` might be confusing, but it's documented.

PatrickRice-KSC:
I think the benefit on top of the maintenance is that we can really let the end users control the tradeoffs they want to make though, instead of codifying it directly into the provider. Have only one file? Eh, sha256 is good. Have 10,000,000 files (at which point, you're probably doing something horribly wrong)? Maybe they either don't want to hash it and they're OK with updating it each time as a result or they want to use md5 instead because it's only being used for change data control as opposed to encryption/security. 

I think there is also the benefit of it being a "known pattern" too - both azure and aws handle it this way (GCP doesn't, because they explicitly store the md5 in the cloud storage bucket so it compares remotely). AWS also handles lambda ZIP files this way as well (haven't looked at functions as much in azure/GCP)

timofurrer:
Yes, the `known pattern` kinda got me :D I'm going to address it tomorrow - now it's too late! Thanks for the valuable feedback 🎉

timofurrer:
@PatrickRice-KSC please have a look on my code changes :)

internal/provider/resource_gitlab_topic.go
```diff
@@ -32,21 +33,48 @@ var _ = registerResource("gitlab_topic", func() *schema.Resource {
...
 				Deprecated:  "GitLab 14.9 introduced the proper deletion of topics. This field is no longer needed.",
 			},
 			"description": {
-				Description: "A text describing the topic",
+				Description: "A text describing the topic.",
 				Type:        schema.TypeString,
 				Optional:    true,
 			},
+			"avatar": {
+				Description: "A local path to the avatar image to upload. **Note**: not available for imported resources",
```
jtymes, line 52:
```suggestion
				Description: "A local path to the avatar image to upload. **Note**: not available for imported resources.",
```
I see you're adding periods to others, but don't have one at the end here 😉

jtymes (commented):
Minor nit, otherwise LGTM.

internal/provider/resource_gitlab_topic.go
timofurrer: thanks!

PatrickRice-KSC:
@timofurrer  - Thanks for the updates, I agree with @jtymes , it looks good to me too. Let's get it merged and launched! 🚀

PatrickRice-KSC (approved):

nagyv (approved):
Approving after @jtymes and @PatrickRice-KSC

PR #969, opened 2022-03-19
Use context for all GitLab API calls

timofurrer:
This change set uses the context for all GitLab API calls.

Required:

* https://github.com/xanzy/go-gitlab/pull/1419


armsnyder (approved):
lgtm

PR #971, opened 2022-03-20
Add new resource for linking Project to Group as file_project_template

PatrickRice-KSC:
## Description

Adds support for linking a template project to groups so that projects within that group can use templates from the template project.

Partially implements #736

    - Create the resource
    - Update the attributes
    - Import the resource


PatrickRice-KSC:
This PR is blocked for now because I haven't implemented a basic test for it yet (I'm going to plan to work on this over the next day or two to get that implemented). I wanted to get this in here, since this is kind-of an odd resource since it doesn't have a traditional "create" method or "delete" method and I wanted to get thoughts on the coding patterns.

Odd things:

- I'm explicitly not implementing an Import because I don't think it makes sense for this resource (and left a comment as such in the code)
- The `create` and `update` methods are the same, since they both call update behind the scenes
- The `delete` resource doesn't have anything to delete, so I marked the group_id as force new and have delete set the project value to 0 (which removes the project from the group as a template).

Very interested in thoughts 😄

PatrickRice-KSC:
Dropping another commit here for now, the ee tests will fail on this commit because the delete test is going to fail.

The reason for this is because to "delete" the linkage here requires passing in a PUT request with a literally empty value (I.e., localhost:8080/api/v4/groups/1234?file_template_project_id) and the current go-gitlab library doesn't support that as far as I can tell. If you pass in 0, you get ?file_template_project_id=0, which makes sense. If you pass in nil, it's omitted completely from the marshal.

I'm working through this still, but let me know if I'm just missing something simple. I think that's the last step (other than linting/documentation) for this to be ready.

PatrickRice-KSC:
Ok, I've applied a.... let's call it a "workaround", and you guys can read the block comment that's left in the code to explain why it works 😋 . I'm not particularly proud of it, but I can't figure out a way to make it work otherwise without removing the `omitempty` in go-gitlab, which would absolutely be a breaking change for everyone who consumes that library (ourselves included), or without manually creating a PUT request outside of the library. I seriously debated the second option, as it seems potentially easier to read/understand in the future.

@timofurrer  | @armsnyder  - this is ready for review.

internal/provider/resource_gitlab_group_project_file_template.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	return resourceGitLabGroupProjectFileTemplateRead(ctx, d, meta)
+}
+
+func resourceGitLabGroupProjectFileTemplateDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*gitlab.Client)
+	groupId := d.Get("group_id").(int)
+	options := &gitlab.UpdateGroupOptions{}
+
+	/*
```
timofurrer, line 100: @PatrickRice-KSC  That's indeed a clever hack 😅 

However, I think we should "workaround" the go-gitlab API by using a request func to alter the body:

```go
_, _, err := client.Groups.UpdateGroup(groupId, nil, gitlab.WithContext(ctx), func(req *retryablehttp.Request) error {
	        hackedOptions := struct {
		        FileTemplateProjectID    *string    `url:"file_template_project_id" json:"file_template_project_id"`
	        }{
		        FileTemplateProjectID:  gitlab.String(""),
	        }
	        body, err := json.Marshal(hackedOptions)
	        if err != nil {
		        return err
	        }
	        if err := req.SetBody(body); err != nil {
		        return err
	        }
		return nil
	})
```

This would intercept the request and set the body with our "hacked" options ... 
Consider it pseudo-code as I haven't tried it ...

internal/provider/resource_gitlab_group_project_file_template_test.go
```diff
@@ -0,0 +1,139 @@
...
+		Steps: []resource.TestStep{
+			{
+				SkipFunc: isRunningInCE,
+				Config:   testAccGroupProjectFileTemplateConfig(rInt, "firstProject"),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGitlabGroupExists("gitlab_group.parent_group", &parentGroup),
+					testAccCheckGitlabProjectExists("gitlab_project.template_project", &templateProject),
+					// Note - we can't use the testAccCheckGitlabGroupAttributes, because that checks the TF
+					// state attributes, and file project template explicitly doesn't exist there.
+					testAccCheckGitlabGroupFileTemplateValue(&parentGroup, &templateProject),
```
timofurrer, line 30: I think it also makes sense to verify the state.

internal/provider/resource_gitlab_group_project_file_template_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/xanzy/go-gitlab"
+	"testing"
+)
+
+func TestAccGitlabGroupProjectFileTemplate_basic(t *testing.T) {
+	rInt := acctest.RandInt()
+	var parentGroup gitlab.Group
+	var templateProject gitlab.Project
```
timofurrer, line 15: Let's create the test group and project outside of TF and focus in the TF config on "what is under test" ...

internal/provider/resource_gitlab_group_project_file_template.go
```diff
@@ -0,0 +1,116 @@
...
+
+-> This resource requires a GitLab Enterprise instance with a Premium license.
+
+**Upstream API**: [GitLab REST API docs](https://docs.gitlab.com/ee/api/groups.html#update-group)`,
+
+		// Since this resource updates an in-place resource, the update method is the same as the create method
+		CreateContext: resourceGitLabGroupProjectFileTemplateCreateOrUpdate,
+		UpdateContext: resourceGitLabGroupProjectFileTemplateCreateOrUpdate,
+		ReadContext:   resourceGitLabGroupProjectFileTemplateRead,
+		DeleteContext: resourceGitLabGroupProjectFileTemplateDelete,
```
timofurrer, line 27: What about the importer 🤔  ?

internal/provider/resource_gitlab_group_project_file_template.go
```diff
@@ -0,0 +1,116 @@
...
+	d.SetId(fmt.Sprintf("%d", group.ID))
+	d.Set("file_template_project_id", group.FileTemplateProjectID)
+
+	return nil
+}
+
+func resourceGitLabGroupProjectFileTemplateCreateOrUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*gitlab.Client)
+
+	groupId := d.Get("group_id").(int)
```
timofurrer, line 80:
```suggestion
	groupID := d.Get("group_id").(int)
```

internal/provider/resource_gitlab_group_project_file_template_test.go
```diff
@@ -0,0 +1,139 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGitlabGroupExists("gitlab_group.parent_group", &parentGroup),
+					testAccCheckGitlabProjectExists("gitlab_project.template_project", &templateProject),
+					testAccCheckGitlabGroupFileTemplateValue(&parentGroup, &templateProject),
+				),
+			},
+			{
+				//Test that when we remove the resource, the group gets it's value set to 0
+				SkipFunc: isRunningInCE,
+				Config:   testAccGroupProjectFileTemplateRemovedConfig(rInt),
```
timofurrer, line 46: When the test project and group is outside of TF as per my other comment we can just rely on the normal post-destroy checks ...

examples/resources/gitlab_group_project_file_template/resource.tf
```diff
@@ -0,0 +1,16 @@
+resource "gitlab_group" "foo" {
+  name        = "group"
+  path        = "group"
+  description = "An example group"
+}
+
+resource "gitlab_project" "bar" {
+  name             = "template project"
+  description      = "contains file templates"
+  visibility_level = "public"
```
timofurrer, line 10: It's missing the very crucial `namespace_id` here, right?

```suggestion
  visibility_level = "public"
  
  namespace_id = gitlab_group.foo.id
```

timofurrer (changes requested):

internal/provider/resource_gitlab_group_project_file_template.go
PatrickRice-KSC: @timofurrer  - I considered this, but importing doesn't make much sense in this particular case. Importing tends to be used to manage a pre-existing resource, but there isn't really a resource being managed here, so there isn't much to import. If you wanted to overwrite what was there, you'd just... include the resource block, and the first time it ran it would overwrite the existing value.

We could certainly add one, it just seems like it's code we need to maintain that doesn't actually provide value :-/

internal/provider/resource_gitlab_group_project_file_template.go
PatrickRice-KSC: huh, ok. I'll go ahead and do some testing with this and see if I can get it to work. Thanks!

examples/resources/gitlab_group_project_file_template/resource.tf
PatrickRice-KSC: well now I'm curious how all the tests pass without that... I'll do some digging here.

timofurrer: It's only missing in the example :)

PatrickRice-KSC: Whew. little bit of panic there for a second 😂

internal/provider/resource_gitlab_group_project_file_template.go
PatrickRice-KSC: @timofurrer  - Nice code! This worked exactly right the first time 😄 . I refactored the tests using the old method and made sure they passed, then updated this, and tests still pass.

internal/provider/resource_gitlab_group_project_file_template_test.go
PatrickRice-KSC: Added a state verify check function in addition to the API checking.

internal/provider/resource_gitlab_group_project_file_template.go
PatrickRice-KSC: refactored throughout the file. Sorry, habit due to naming standards we use at my job 😄

PatrickRice-KSC:
Squashed and rebased since I think we're close to done here

internal/provider/resource_gitlab_group_project_file_template.go
```diff
@@ -0,0 +1,133 @@
...
+})
+
+func resourceGitLabGroupProjectFileTemplateRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*gitlab.Client)
+
+	groupID := d.Get("group_id").(int)
+	group, _, err := client.Groups.GetGroup(groupID, nil, gitlab.WithContext(ctx))
+	if err != nil {
+		if is404(err) {
+			log.Printf("[DEBUG] gitlab group %d not found so removing from state", groupID)
```
timofurrer, line 61:
```suggestion
			log.Printf("[DEBUG] gitlab group %d not found, removing from state", groupID)
```

internal/provider/resource_gitlab_group_project_file_template.go
```diff
@@ -0,0 +1,133 @@
...
+			d.SetId("")
+			return nil
+		}
+		return diag.FromErr(err)
+	}
+	if group.MarkedForDeletionOn != nil {
+		log.Printf("[DEBUG] gitlab group %s is marked for deletion", d.Id())
```
timofurrer, line 68:
```suggestion
		log.Printf("[DEBUG] gitlab group %s is marked for deletion, removing from state", d.Id())
```

internal/provider/resource_gitlab_group_project_file_template.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// Creating the resource means updating the existing group to link the project to the group.
+	options := &gitlab.UpdateGroupOptions{}
+	if d.HasChanges("file_template_project_id") {
+		options.FileTemplateProjectID = gitlab.Int(d.Get("file_template_project_id").(int))
+	}
+
+	_, _, err := client.Groups.UpdateGroup(groupID, options)
+	if err != nil {
+		return diag.Errorf("could not update group after creation %d: %s", groupID, err)
```
timofurrer, line 92:
```suggestion
		return diag.Errorf("unable to update group %d with `file_template_project_id` set to %d: %s", groupID, projectID, err)
```

This also requires to set the `projectID` to `d.Get(...)` ...

internal/provider/resource_gitlab_group_project_file_template.go
```diff
@@ -0,0 +1,133 @@
...
+
+	_, _, err := updateGroupWithOverwrittenFileTemplateOption(client, groupID, options)
+	if err != nil {
+		return diag.Errorf("could not update group %d to remove file template ID: %s", groupID, err)
+	}
+	return resourceGitLabGroupProjectFileTemplateRead(ctx, d, meta)
+}
+
+func updateGroupWithOverwrittenFileTemplateOption(client *gitlab.Client, groupID int, options *gitlab.UpdateGroupOptions) (*gitlab.Group, *gitlab.Response, error) {
+	return client.Groups.UpdateGroup(groupID, options, func(request *retryablehttp.Request) error {
```
timofurrer, line 110: Out of curiosity, what happens if `options` is `nil`? If it works I'd say way pass `nil` instead of an empty options object. WDYT? 🏓

internal/provider/resource_gitlab_group_project_file_template_test.go
```diff
@@ -0,0 +1,149 @@
...
+	"os"
+	"strconv"
+	"testing"
+)
+
+func TestAccGitlabGroupProjectFileTemplate_basic(t *testing.T) {
+	// Since we do some manual setup in this test, we need to handle the test skip first.
+	if os.Getenv(resource.EnvTfAcc) == "" {
+		t.Skip(fmt.Sprintf("Acceptance tests skipped unless env '%s' set", resource.EnvTfAcc))
+	}
```
timofurrer, line 18: We have a test helper for that:

```suggestion
	testAccCheck(t)
```

internal/provider/resource_gitlab_group_project_file_template_test.go
```diff
@@ -0,0 +1,149 @@
...
+					testAccCheckGitlabGroupFileTemplateValue(baseGroup, secondProject),
+					testAccCheckGitlabGroupFileTemplateStateValue(baseGroup, secondProject),
+				),
+			},
+			{
+				//Test that when we remove the resource, the group gets its value set to 0
+				SkipFunc: isRunningInCE,
+				Config:   testAccGroupProjectFileTemplateConfig(baseGroup.ID, secondProject.ID),
+				Destroy:  true,
+			},
```
timofurrer, line 50: Having a proper `CheckDestroy` function is actually enough to check this destroy behavior. I don't think we need to have an explicit test step for that.

```suggestion
```

internal/provider/resource_gitlab_group_project_file_template_test.go
```diff
@@ -0,0 +1,149 @@
...
+		CheckDestroy:      testAccCheckProjectFileTemplateDestroy,
+		Steps: []resource.TestStep{
+			{
+				SkipFunc: isRunningInCE,
+				Config:   testAccGroupProjectFileTemplateConfig(baseGroup.ID, firstProject.ID),
+				Check: resource.ComposeTestCheckFunc(
+					// Note - we can't use the testAccCheckGitlabGroupAttributes, because that checks the TF
+					// state attributes, and file project template explicitly doesn't exist there.
+					testAccCheckGitlabGroupFileTemplateValue(baseGroup, firstProject),
+					testAccCheckGitlabGroupFileTemplateStateValue(baseGroup, firstProject),
```
timofurrer, line 33: I think it makes sense to use the off the shelve functions when ever possibl
```suggestion
					resource.TestCheckResourceAttr("gitlab_group_project_file_template.linking_template", "group_id", baseGroup.ID),
					resource.TestCheckResourceAttr("gitlab_group_project_file_template.linking_template", "file_template_project_id", firstProject.ID),
```

internal/provider/resource_gitlab_group_project_file_template_test.go
```diff
@@ -0,0 +1,149 @@
...
+		}
+
+		if projectID != strconv.Itoa(project.ID) {
+			return fmt.Errorf("Project didn't match with the state file. State had %s and expected %d", projectID, project.ID)
+		}
+		return nil
+	}
+}
+
+func setupTestData(t *testing.T) (*gitlab.Group, *gitlab.Project, *gitlab.Project) {
```
timofurrer, line 72: Why not use the test helpers from `helper_test.go`? You may need to create a new one to create a project within a group.

timofurrer (changes requested):

internal/provider/resource_gitlab_group_project_file_template_test.go
PatrickRice-KSC: I updated this along with refactoring the resource_gitlab_project_test file, which is where I pulled the original logic from.

internal/provider/resource_gitlab_group_project_file_template.go
PatrickRice-KSC: `nil` works exactly the same, so I've gone ahead and updated it. I don't really have an opinion in this case about which one is better, I think they're equally descriptive. 😄

internal/provider/resource_gitlab_group_project_file_template_test.go
PatrickRice-KSC: Updated - I created a new method to create a project with a namespace when the input namespace != 0, and refactored the existing createProject to call that with a 0 namespace so we have less duplicated code.

internal/provider/resource_gitlab_group_project_file_template_test.go
timofurrer: great! Thanks for going the extra mile 🎉

internal/provider/resource_gitlab_group_project_file_template.go
```diff
@@ -0,0 +1,134 @@
...
+func resourceGitLabGroupProjectFileTemplateCreateOrUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*gitlab.Client)
+
+	groupID := d.Get("group_id").(int)
+	projectID := gitlab.Int(d.Get("file_template_project_id").(int))
+
+	// Creating the resource means updating the existing group to link the project to the group.
+	options := &gitlab.UpdateGroupOptions{}
+	if d.HasChanges("file_template_project_id") {
+		options.FileTemplateProjectID = gitlab.Int(d.Get("file_template_project_id").(int))
```
timofurrer, line 88:
```suggestion
		options.FileTemplateProjectID = projectID
```

timofurrer (changes requested):
I've just stumbled over one minor thing. I think afterwards, we are ready to merge :tada:

internal/provider/resource_gitlab_group_project_file_template.go
PatrickRice-KSC: Committed - So I think we'll be all set once checks run!

internal/provider/helper_test.go
```diff
@@ -118,16 +118,29 @@ func testAccCurrentUser(t *testing.T) *gitlab.User {
 
 // testAccCreateProject is a test helper for creating a project.
 func testAccCreateProject(t *testing.T) *gitlab.Project {
+	return testAccCreateProjectWithNamespace(t, 0)
+}
+
+// testAccCreateProject is a test helper for creating a project. This method accepts a namespace to great a project
```
armsnyder, line 124: lying comment!

```suggestion
// testAccCreateProjectWithNamespace is a test helper for creating a project. This method accepts a namespace to great a project
```

PatrickRice-KSC: Lol, nice catch!

Issue #956, opened 2022-03-14
Display gitlab_repository_file diff as decoded base64


mkjmdski:
### Feature Description

Is it possible in terms of terraform to use some kind of "diff plugin" to display base64 encoded files in decoded format? At this point any diff on gitlab_repository_file is unreadable for human. Once again thanks for implementing that :).

timofurrer:
@mkjmdski I've addressed in #972. I think the only way to achieve this is to allow to specify plan text as `content` and automatically encode it to base64. I've implemented it in a backwards-compatible way.

I'm open to feedback if you want to try it out :)

PR #972, opened 2022-03-20
resource/gitlab_repository_file: Support auto base64 encoding for content

timofurrer:
This change sets allows to specify not-yet base64 encoded `content` and the resource will automatically base64 encode it for the API.

This will make the plan match more readable and will solve the limitations described in #956 

Initially I thought it's best to use a `DiffSuppressFunc` and always store the `content` as base64. However, I think it's best to store it in the same format (text vs. plan) as given in the config. This ensures that access the resource content in the config produces deterministic results.

Closes: #956


PatrickRice-KSC (dismissed):
This looks good to me @timofurrer  - great work!

nagyv (approved):
Approving after @PatrickRice-KSC

PR #983, opened 2022-03-22
Fix gitlab_project datasource tests

timofurrer:


docker-compose.yml
```diff
@@ -4,7 +4,7 @@ version: '3'
 # Only one of these services should be run at a time.
 services:
   gitlab-ce:
-    image: gitlab/gitlab-ce
+    image: gitlab/gitlab-ce:14.9.0-ce.0
```
jtymes, line 7: Will dependabot pick up the patch releases?

jtymes (approved):

docker-compose.yml
nagyv: @jtymes Do you know a way to find it out?

jtymes: Sure, just was curious. :shipit:

armsnyder (approved):

Issue #85, opened 2019-02-06
gitlab_project: Support creating project for specific user


eddiemcbride:
### Terraform Version
Terraform v0.11.7
+ provider.gitlab v1.1.0

### Affected Resource(s)
Please list the resources as a list, for example:
- gitlab_project
- gitlab_user

### Terraform Configuration Files
```
# Define variables
variable "gitlab_token" {}
variable "gitlab_url" {}
variable "gitlab_username" {}
variable "gitlab_usertitle" {}
variable "gitlab_userp" {}
variable "gitlab_useremail" {}
variable "gitlab_projectname" {}
variable "gitlab_projectdesc" {}

# Set provider
provider "gitlab" {
    token = "${var.gitlab_token}"
    base_url = "${var.gitlab_url}"
}

# Add a user
resource "gitlab_user" "new_gitlab_user" {
  name             = "${var.gitlab_usertitle}"
  username         = "${var.gitlab_username}"
  password         = "${var.gitlab_userp}"
  email            = "${var.gitlab_useremail}"
  is_admin         = false
  can_create_group = true
  is_external      = false
}

# Create a project under the new user
resource "gitlab_project" "new_gitlab_project" {
  name        = "${var.gitlab_projectname}"
  description = "${var.gitlab_projectdesc}"
  visibility_level = "private"
  namespace_id = "${gitlab_user.new_gitlab_user.id}"
}
```

### Expected Behavior
It should create the new GitLab project under the new user namespace.

### Actual Behavior
It creates the project under a random user's namespace

### Comments
This is my first time posting an issue here so if I've done this incorrectly, I apologize.  We have had multiple people try this internally and we can easily recreate the scenario.   Also, we are using GitLab Enterprise Edition 11.7.4-ee.

armsnyder:
I was able to reproduce this. Based on my reading of the [create-project API](https://docs.gitlab.com/ee/api/projects.html#create-project) and [create-project-for-user API](https://docs.gitlab.com/ee/api/projects.html#create-project-for-user), the `namespace_id` argument is meant to be used to specify a group id, not a user id.

My proposal is we add a `user_id` argument to the `gitlab_project` resource, and if it is set, then use the create-project-for-user API to create the project.

tyriis:
As this is still an issue (at least for me), I have had a closer look.
related to: https://docs.gitlab.com/ee/api/namespaces.html namesaces, user_ids and group ids are not the same.
[...]

PR #987, opened 2022-03-26
Officially support creating project in user namespace using `namespace_id` from user resources / data sources

timofurrer:
This change set addresses #85. It implements the `namespace_id` attribute on the `gitlab_user` and `gitlab_users` resource and data sources respectively. 

It also adds a test and example for the `gitlab_project` resource for creating a project in a users namespace given the `namespace_id` from the former mentioned resources / data sources.

I've implemented this upstream in https://gitlab.com/gitlab-org/gitlab/-/merge_requests/82045 and will be available in GitLab 14.10, thus this PR is blocked until:

* GitLab 14.10 is out on 2022-04-22
* https://github.com/xanzy/go-gitlab/pull/1430 is merged

Closes: #85


internal/provider/data_source_gitlab_user.go
```diff
@@ -171,6 +171,12 @@ var _ = registerDataSource("gitlab_user", func() *schema.Resource {
 				Type:        schema.TypeString,
 				Computed:    true,
 			},
+			"namespace_id": {
+				Description: "The ID of the users namespace. Requires admin token to access this field. Available since GitLab 14.10.",
```
jtymes, line 175: Assuming this is possessive and not a plural?
```suggestion
				Description: "The ID of the user's namespace. Requires admin token to access this field. Available since GitLab 14.10.",
```

internal/provider/data_source_gitlab_users.go
```diff
@@ -212,6 +212,12 @@ var _ = registerDataSource("gitlab_users", func() *schema.Resource {
 							Type:        schema.TypeString,
 							Computed:    true,
 						},
+						"namespace_id": {
+							Description: "The ID of the users namespace. Requires admin token to access this field. Available since GitLab 14.10.",
```
jtymes, line 216: If my other comment stands, then same thing here.

internal/provider/resource_gitlab_user.go
```diff
@@ -110,6 +110,12 @@ var _ = registerResource("gitlab_user", func() *schema.Resource {
 				Default:          "active",
 				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice(validUserStateValues, false)),
 			},
+			"namespace_id": {
+				Description: "The ID of the users namespace. Available since GitLab 14.10.",
```
jtymes, line 114: Should the note about admin only be here too?

jtymes (changes requested):

internal/provider/resource_gitlab_user.go
timofurrer: well, the entire resource can only be used by admins, thus I thought it would be redundant and only mentioned it for the data sources.

internal/provider/data_source_gitlab_users.go
timofurrer: It mentions the admin limitation already or what are you referring to?

jtymes: Sorry, I meant the `users -> user's` suggestion.

jtymes (approved):

armsnyder (approved):

PR #988, opened 2022-03-26
resource/gitlab_project_mirror: Support deletion on destroy

timofurrer:
This change set add support for deletion on a destroy. It also switched to the new `GET` endpoint to retrieve a single project mirror instead of finding the proper one in the list `GET` endpoint.

It's currently blocked by:

* GitLab 14.10 release
* https://github.com/xanzy/go-gitlab/pull/1432 is merged

### DIFF
--- a/internal/provider/resource_gitlab_project_mirror.go
+++ b/internal/provider/resource_gitlab_project_mirror.go
@@ -19,6 +19,10 @@ var _ = registerResource("gitlab_project_mirror", func() *schema.Resource {
 This is for *pushing* changes to a remote repository. *Pull Mirroring* can be configured using a combination of the
 import_url, mirror, and mirror_trigger_builds properties on the gitlab_project resource.
 
+-> **Destroy Behavior** GitLab 14.10 introduced an API endpoint to delete a project mirror.
+   Therefore, for GitLab 14.10 and newer the project mirror will be destroyed when the resource is destroyed.
+   For older versions, the mirror will be disabled and the resource will be destroyed.
+
 **Upstream API**: [GitLab REST API docs](https://docs.gitlab.com/ee/api/remote_mirrors.html)`,
 
 		CreateContext: resourceGitlabProjectMirrorCreate,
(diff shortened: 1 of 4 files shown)


internal/provider/resource_gitlab_project_mirror.go
```diff
@@ -83,6 +87,12 @@ import_url, mirror, and mirror_trigger_builds properties on the gitlab_project r
 				Optional:    true,
 				Default:     true,
 			},
+			"disable_on_destroy": {
+				Description: "Disables the mirror on a destroy instead of deleting it. **Note**: project mirror deletion has only available from GitLab 14.10 onwards. In v4 of the provider the default will be switched from `true` to `false`.",
```
jtymes, line 91:
```suggestion
				Description: "Disables the mirror on a destroy instead of deleting it. **Note**: project mirror deletion is only available from GitLab 14.10 onwards. In v4 of the provider the default will be switched from `true` to `false`.",
```

jtymes (changes requested):

jtymes (dismissed):

armsnyder:
IMO it's ok to change the behavior to always delete on destroy.

timofurrer:
> IMO it's ok to change the behavior to always delete on destroy.

The reason for it is backwards-compatibility with GitLab 14.9 and earlier releases.

armsnyder:
Could you detect the version in the delete function to avoid needing the attribute?

timofurrer:
> Could you detect the version in the delete function to avoid needing the attribute?

Yes, we could. I was hesitant with that because I'm not convinced yet that a silent different behavior for one than the other GitLab version is a thing we want.

timofurrer:
hmmm, on second thought ... maybe it's okay 😅  - because it's just what GitLab for that specific version supports, so 🤷 

what do the others think? @PatrickRice-KSC @jtymes ?

armsnyder:
Here's why I don't see it as a breaking change:

1. Behavior changes are not categorically breaking.
2. When a user destroys a resource they expect it to be destroyed.
3. The thing being destroyed is a mirror configuration. No actual data is being destroyed as a result of the change.

I would like to avoid adding an attribute if possible, unless this is a degree of control that users need.

timofurrer:
@armsnyder alright, good arguments - you've convinced me 😅 

I've pushed another change set :)

jtymes (approved):

armsnyder (approved):
🚢

Issue #518, opened 2021-01-07
Feature Request: ability to set "Enforce SSO-only authentication for this group"


cweidinger:
We're using terraform to ensure certain values are set so that no one can accidentally undo the setting. This is one of those critical security settings.

teekhandekar:
@cweidinger Can I work on this ?

timofurrer:
Checking the API docs it seems like this isn't supported in the API yet :'( 

@cweidinger did you have any specific API in mind which I'm missing here?

jtymes:
> Checking the API docs it seems like this isn't supported in the API yet :'( 

Correct. [Related issue](https://gitlab.com/gitlab-org/gitlab/-/issues/297006).

PR #989, opened 2022-03-26
Use single `GET` endpoints to retrieve GAT and PAT

timofurrer:
We should be able to merge this :)

### DIFF
--- a/internal/provider/resource_gitlab_group_access_token.go
+++ b/internal/provider/resource_gitlab_group_access_token.go
@@ -166,50 +166,31 @@ func resourceGitlabGroupAccessTokenRead(ctx context.Context, d *schema.ResourceD
 	}
 
 	log.Printf("[DEBUG] read gitlab GroupAccessToken %d, group ID %s", groupAccessTokenId, group)
-
-	//there is a slight possibility to not find an existing item, for example
-	// 1. item is #101 (ie, in the 2nd page)
-	// 2. I load first page (ie. I don't find my target item)
-	// 3. A concurrent operation remove item 99 (ie, my target item shift to 1st page)
-	// 4. a concurrent operation add an item
-	// 5: I load 2nd page  (ie. I don't find my target item)
-	// 6. Total pages and total items properties are unchanged (from the perspective of the reader)
-
-	page := 1
-	for page != 0 {
-		groupAccessTokens, response, err := client.GroupAccessTokens.ListGroupAccessTokens(group, &gitlab.ListGroupAccessTokensOptions{Page: page, PerPage: 100}, gitlab.WithContext(ctx))
-		if err != nil {
-			return diag.FromErr(err)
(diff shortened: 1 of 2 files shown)


Shocktrooper:
@timofurrer Is there any GitHub action we can use that is similar to the !RemindMe 1 week/1 month/etc. of reddit. Where there can be a post/notification to look at a PR/issue when a certain period of time is reached? This would be useful for when functionality in the provider is blocked by a functionality that is scheduled to be released in a future GitLab version so as not to forget about something like this PR

timofurrer:
@Shocktrooper a quick search revealed at least one: https://github.com/agrc/reminder-action

Though I'm not really sure if it's worth the maintenance - GitLab releases are always at the same day of the month, every month, thus it's kinda hard to miss - and I've created a milestone where to assign these issues / PRs. I think that gives us an additional benefit over the reminder, because we can actually filter them :)

Shocktrooper:
@timofurrer The milestones seem like a good way to go.  I was just thinking for some issues where maybe the functionality is scheduled a long way down the line/not at all vs within a few months. They would be reminders to check if the upstream API has been released such as for an issue like [this](https://github.com/gitlabhq/terraform-provider-gitlab/issues/518) where GitLab has not added the functionality at the time and we might want to check in say like 6 months

timofurrer:
This would break compatibility to GitLab < 14.10

Shocktrooper:
@timofurrer Is this unblocked now that 14.10 has been released and the prerequisite PR has been merged?

internal/provider/resource_gitlab_group_access_token.go
```diff
@@ -166,50 +166,31 @@ func resourceGitlabGroupAccessTokenRead(ctx context.Context, d *schema.ResourceD
...
-		if err != nil {
-			return diag.FromErr(err)
+	groupAccessToken, _, err := client.GroupAccessTokens.GetGroupAccessToken(group, groupAccessTokenId, gitlab.WithContext(ctx))
+	if err != nil {
+		if is404(err) {
+			log.Printf("[DEBUG] GitLab GroupAccessToken %d, group ID %s not found, removing from state", groupAccessTokenId, group)
+			d.SetId("")
+			return nil
 		}
+	}
```
PatrickRice-KSC, line 176:
```suggestion
	if err != nil && is404(err) {
		log.Printf("[DEBUG] GitLab GroupAccessToken %d, group ID %s not found, removing from state", groupAccessTokenId, group)
		d.SetId("")
		return nil
	}
```

Collapse nested IF statements :)

PatrickRice-KSC (changes requested):
Losing internet on the flight since I'm landing, but one comment so far 😄

internal/provider/resource_gitlab_group_access_token.go
timofurrer: This was actually a bug :D It should return an error if `err and not a 404` ...
I've updated the change set :)

PatrickRice-KSC:
No other comments from reviews, thanks @timofurrer  and @Shocktrooper !

PatrickRice-KSC (approved):

Issue #994, opened 2022-03-30
gitlab_project_access_token: Support different access_level


MrLuje:
### Feature Description

When creating gitlab_project_access_token through the provider, the generated user associated to the token is added with "maintainer" role to the project and there is no way to specify the role.

This parameter is supported by the gitlab api ([gitlab_project_access_token](https://docs.gitlab.com/ee/api/project_access_tokens.html#create-a-project-access-token))

<html>
<body>

Attribute | Type | required | Description
-- | -- | -- | --
access_level | Integer | no | A valid access level. Default value is 40 (Maintainer). Other allowed values are 10 (Guest), 20 (Reporter), and 30 (Developer).

</body>
</html>

Example :
```hcl
resource "gitlab_project_access_token" "bar" {
  name = "my project token"
  project = "project"  
  expires_at = "2022-04-01"
  scopes = ["read_repository" , "api", "write_repository", "read_api"]

  access_level = "reporter"   #<--- new 
}
```

### Do you want to implement this?

- [X] I would like to implement this myself 👷

PatrickRice-KSC:
Thanks for the issue @MrLuje ! It looks like the access_level attribute is supported by the go-gitlab library that we use for interacting with the GitLab APIs as well, so I think you're good to start whenever you're ready.

Let us know when you've got a PR you'd like us to review, and make sure you read through the [CONTRIBUTING.md](https://github.com/gitlabhq/terraform-provider-gitlab/blob/main/CONTRIBUTING.md) and let us know if you have any questions!

PR #997, opened 2022-03-30
Add access_level to gitlab_project_access_token resource

MrLuje:
## Description

Add `access_level` to `gitlab_project_access_token` resource #994 

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/internal/provider/resource_gitlab_project_access_token.go
+++ b/internal/provider/resource_gitlab_project_access_token.go
@@ -91,7 +91,7 @@ var _ = registerResource("gitlab_project_access_token", func() *schema.Resource 
 				Computed:    true,
 			},
 			"access_level": {
-				Description:      fmt.Sprintf("The access level of the associated user_id membership (cf resource_gitlab_project_membership). Valid values are: %s", renderValueListForDocs(validProjectAccessLevelNames)),
+				Description:      fmt.Sprintf("The access level of the associated user_id project membership (cf resource_gitlab_project_membership). Valid values are: %s. Default is `%s`", renderValueListForDocs(validProjectAccessLevelNames), accessLevelValueToName[gitlab.MaintainerPermissions]),
 				Type:             schema.TypeString,
 				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice(validProjectAccessLevelNames, false)),
 				Optional:         true,
--- a/internal/provider/resource_gitlab_project_access_token.go
+++ b/internal/provider/resource_gitlab_project_access_token.go
@@ -91,7 +91,7 @@ var _ = registerResource("gitlab_project_access_token", func() *schema.Resource 
 				Computed:    true,
 			},
 			"access_level": {
-				Description:      fmt.Sprintf("The access level of the associated user_id project membership (cf resource_gitlab_project_membership). Valid values are: %s. Default is `%s`", renderValueListForDocs(validProjectAccessLevelNames), accessLevelValueToName[gitlab.MaintainerPermissions]),
+				Description:      fmt.Sprintf("The access level for the project access token. Valid values are: %s. Default is `%s`.", renderValueListForDocs(validProjectAccessLevelNames), accessLevelValueToName[gitlab.MaintainerPermissions]),
 				Type:             schema.TypeString,
 				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice(validProjectAccessLevelNames, false)),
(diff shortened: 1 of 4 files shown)


internal/provider/resource_gitlab_project_access_token.go
```diff
@@ -90,17 +90,27 @@ var _ = registerResource("gitlab_project_access_token", func() *schema.Resource
 				Type:        schema.TypeInt,
 				Computed:    true,
 			},
+			"access_level": {
+				Description:      fmt.Sprintf("The access level of the associated user_id project membership (cf resource_gitlab_project_membership). Valid values are: %s. Default is `%s`", renderValueListForDocs(validProjectAccessLevelNames), accessLevelValueToName[gitlab.MaintainerPermissions]),
```
PatrickRice-KSC, line 94:
```suggestion
				Description:      fmt.Sprintf("The access level of the associated user_id project membership (cf resource_gitlab_project_membership). Valid values are: %s. Default is `%s`.", renderValueListForDocs(validProjectAccessLevelNames), accessLevelValueToName[gitlab.MaintainerPermissions]),
```

Nitpick, but our standard has been to end bullet points with a full stop 😄

internal/provider/resource_gitlab_project_access_token.go
```diff
@@ -90,17 +90,27 @@ var _ = registerResource("gitlab_project_access_token", func() *schema.Resource
...
+				Type:             schema.TypeString,
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice(validProjectAccessLevelNames, false)),
```
PatrickRice-KSC, line 96: @MrLuje  - We have examples of both in the code base, but is there a reason in this specific case why we're ignoring casing on the input string?

PatrickRice-KSC:
@MrLuje - Let us know if you're done with commits, overall what you've pushed is looking good! 🎉

internal/provider/resource_gitlab_project_access_token.go
timofurrer:
```suggestion
				Description:      fmt.Sprintf("The access level for the project access token. Valid values are: %s. Default is `%s`", renderValueListForDocs(validProjectAccessLevelNames), accessLevelValueToName[gitlab.MaintainerPermissions]),
```

I think that most use cases don't even require nor need to know about the `user_id` generated for the access token (in fact not even required to know that a "user" is even created ...). 
Since *Access Levels* for tokens in GitLab is a common thing I think we can be quite brief on the docs here, too (like the upstream docs are for these cases).

MrLuje:
@PatrickRice-KSC I'm done

timofurrer (approved):
LGTM 🎉

timofurrer:
Thanks @MrLuje for the contribution and the fast turn arounds, very much appreciated 🎉

internal/provider/resource_gitlab_project_access_token.go
```diff
@@ -90,17 +90,27 @@ var _ = registerResource("gitlab_project_access_token", func() *schema.Resource
 				Type:        schema.TypeInt,
 				Computed:    true,
 			},
+			"access_level": {
+				Description:      fmt.Sprintf("The access level for the project access token. Valid values are: %s. Default is `%s`.", renderValueListForDocs(validProjectAccessLevelNames), accessLevelValueToName[gitlab.MaintainerPermissions]),
+				Type:             schema.TypeString,
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice(validProjectAccessLevelNames, false)),
+				Optional:         true,
+				Default:          accessLevelValueToName[gitlab.MaintainerPermissions],
```
armsnyder, line 98: @timofurrer @MrLuje This seems to be causing errors in our CI pipeline. Is there a reason to default to maintainer access, or can this default be removed?

```
Error: POST http://127.0.0.1:8080/api/v4/projects/42/access_tokens: 400 {message: 400 Bad request - Could not provision maintainer access to project access token}
```

armsnyder: Ah, I saw your comment below

> `// default permission on gitlab side when unspecified`

In this case I think the attribute should be Optional+Computed instead of having a Default

armsnyder: Ah, disregard. The CI failure was caused by the expiration date, not the access level.

PR #998, opened 2022-03-31
Lint Issues- Testing out addition of Check Destroy in some Acceptance Tests + other refactoring

### LINKED ISSUE #433: Fix AT001: missing CheckDestroy
Check: [AT001](https://github.com/bflad/tfproviderlint/tree/master/passes/AT001)

```
./gitlab/resource_gitlab_project_share_group_test.go:17:28: AT001: missing CheckDestroy
./gitlab/resource_gitlab_project_test.go:511:28: AT001: missing CheckDestroy
```

We should fix these reported issues, and then remove this check from the list of ignored checks in the gitlabproviderlint make target.

To reproduce, run `go run ./gitlabproviderlint ./gitlab` (pending #432)

Shocktrooper:
Resolves #433


internal/provider/resource_gitlab_project_test.go
```diff
@@ -898,8 +894,7 @@ func testAccCheckGitlabProjectMirroredAttributes(project *gitlab.Project, want *
 	}
 }
 
-// lintignore: AT002 // not a Terraform import test
-func TestAccGitlabProject_importURLMirrored(t *testing.T) {
+func TestAccGitlabProject_ImprtURLMirrored(t *testing.T) {
 	// Since we do some manual setup in this test, we need to handle the test skip first.
 	if os.Getenv(resource.EnvTfAcc) == "" {
 		t.Skip(fmt.Sprintf("Acceptance tests skipped unless env '%s' set", resource.EnvTfAcc))
```
timofurrer, line 900: while you are at it, can you replace this with a `testAccCheck(t)` ?

internal/provider/resource_gitlab_project_test.go
```diff
@@ -770,8 +767,7 @@ func TestAccGitlabProject_transfer(t *testing.T) {
 	})
 }
 
-// lintignore: AT002 // not a Terraform import test
-func TestAccGitlabProject_importURL(t *testing.T) {
+func TestAccGitlabProject_imprtURL(t *testing.T) {
 	// Since we do some manual setup in this test, we need to handle the test skip first.
 	if os.Getenv(resource.EnvTfAcc) == "" {
```
timofurrer, line 772: same here, with `testAccCheck(t)`

timofurrer:
Can you remove the AT001 check from the provider lint check ignores:

https://github.com/gitlabhq/terraform-provider-gitlab/blob/9c09272feaac4be52233e98df54ecf4b61a35476/GNUmakefile#L23

internal/provider/resource_gitlab_project_share_group_test.go
```diff
@@ -109,6 +109,34 @@ func testAccCheckGitlabProjectIsNotShared(projectName string) resource.TestCheck
 	}
 }
 
+func testAccCheckGitlabProjectShareGroupDestroy(s *terraform.State) error {
+	var projectId string
+	var groupId int
+	var err error
+
+	for _, rs := range s.RootModule().Resources {
```
timofurrer, line 117: In case there is no resource, the subsequent `GetProject` will fail ...

internal/provider/resource_gitlab_group_share_group_test.go
```diff
@@ -125,46 +104,50 @@ func testAccCheckGitlabGroupIsNotShared(groupName string) resource.TestCheckFunc
...
+			}
 		}
+	}
 
-		resource "gitlab_group" "test_share" {
-		  name = "%[1]s_share"
-		  path = "%[1]s_share"
-		}
+	// Get Main Group
+	group, _, err := testGitlabClient.Groups.GetGroup(groupId, nil)
```
timofurrer, line 122: This will fail if the `for` doesn't find the resource ...

internal/provider/resource_gitlab_project_test.go
Shocktrooper: Done

internal/provider/resource_gitlab_project_test.go
Shocktrooper: Done

internal/provider/resource_gitlab_group_share_group_test.go
Shocktrooper: Done

internal/provider/resource_gitlab_project_share_group_test.go
Shocktrooper: Fixed

Shocktrooper:
@armsnyder If you have a chance as well to take a look the issue is not apparent to me why it is failing.

armsnyder:
Hi @Shocktrooper -- I couldn't tell what the issue was either. I assume it has to do with the new function you added to render out the Config field in each TestStep. Maybe try removing the TestSteps one by one, and running a targeted test to catch where the problem lies.

I also opened this PR to make the error more helpful in the future, not that it will help us now.  😄 https://github.com/hashicorp/terraform-plugin-sdk/pull/925

Shocktrooper:
@armsnyder I figured it out. Apparently having a completely blank resource config was the issue from what I can tell. It was messing somehow with the test step rendering.  After looking at the test again I didn't think it was needed either because we have a check destroy function so it should all be working now :)

timofurrer:
@Shocktrooper @armsnyder I've actually noticed that in other providers test where the have something like `" "` as a config to prevent this ...

Shocktrooper:
@timofurrer interesting do you have an example of this in action as I am unsure how that would be different than what I replaced in my last commit

timofurrer:
@Shocktrooper yes, here:

https://github.com/hashicorp/terraform-provider-google/blob/b5b0438689c8a10fad804bdbc59ff45afa8aeb33/google/resource_cloud_run_service_test.go#L63

internal/provider/resource_gitlab_project_test.go
```diff
@@ -689,7 +686,7 @@ func TestAccGitlabProject_import(t *testing.T) {
 	})
 }
 
-func TestAccGitlabProject_nestedImport(t *testing.T) {
+func TestAccGitlabProject_nestedImprt(t *testing.T) {
```
timofurrer, line 689:
```suggestion
func TestAccGitlabProject_nestedImport(t *testing.T) {
```

internal/provider/resource_gitlab_project_test.go
```diff
@@ -769,9 +766,7 @@ func TestAccGitlabProject_transfer(t *testing.T) {
 	})
 }
 
-// lintignore: AT002 // not a Terraform import test
-func TestAccGitlabProject_importURL(t *testing.T) {
-	// Since we do some manual setup in this test, we need to handle the test skip first.
+func TestAccGitlabProject_imprtURL(t *testing.T) {
```
timofurrer, line 769:
```suggestion
func TestAccGitlabProject_importURL(t *testing.T) {
```

did a search and replace go wrong with this `imprt` ?

Shocktrooper: I just renamed it because of `lintignore: AT002` which triggers off of the word import.  Looking at it again I wonder if the tests need to be refactored a tad to just do the state verifies after each test step in other tests and we can remove the import tests unless you think there is value in the 4 that are there

internal/provider/resource_gitlab_project_test.go
Shocktrooper: See below comment about imports

internal/provider/resource_gitlab_project_test.go
timofurrer: Yes, the proper solution would be to add the import verification test steps to the other tests and if there is an *import* name in the test otherwise, we may rename (only if it makes sense) or ignore the error.

internal/provider/resource_gitlab_project_test.go
Shocktrooper: fixed

internal/provider/resource_gitlab_project_test.go
Shocktrooper: Left some of the import tests as they seemed a bit more intricate than a basic import test and deleted others that looked like a basic import test

timofurrer (approved):
LGTM 🎉

PR #1001, opened 2022-04-02
Fix tfplugindocs generate error on M1 macs

armsnyder:
## Description

Fixes this error since upgrading to tfproviderdocs 0.7.0, which changed how the terraform binary is loaded. (I have an M1 mac, which I assume is the cause. It's always the cause. 😓 I tried upgrading to the latest Terraform for my local version, and that didn't help.)

```
make generate                                                                                                protected-environment-multi
GITLAB_TOKEN="" /Users/adam/repos/terraform-provider-gitlab/bin/tfplugindocs generate
rendering website for provider "terraform-provider-gitlab"
copying any existing content to tmp dir
exporting schema from Terraform
compiling provider "gitlab"
using Terraform CLI binary from PATH if available, otherwise downloading latest Terraform CLI binary
running terraform init
Error executing command: unable to generate website: exit status 1

Error: Incompatible provider version

Provider registry.terraform.io/hashicorp/gitlab v0.0.1 does not have a
package available for your current platform, darwin_amd64.

Provider releases are separate from Terraform CLI releases, so not all
providers are available for all platforms. Other versions of this provider
may have different platforms supported.



make: *** [generate] Error 1
```

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -11,7 +11,8 @@ build: ## Build the provider binary.
 
 generate: tool-tfplugindocs ## Generate files to be checked in.
 	@# Setting empty environment variables to work around issue: https://github.com/hashicorp/terraform-plugin-docs/issues/12
-	GITLAB_TOKEN="" $(GOBIN)/tfplugindocs generate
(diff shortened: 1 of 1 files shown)


nagyv (approved):
LGTM. Thanks

PR #1003, opened 2022-04-02
Added external Wiki service

### LINKED ISSUE #996: Feature Request: gitlab_service_external_wiki
### Feature Description

Hello,

we use confluence as wiki for our projects. It would be awesome if the integration could be configured via terraform. At the moment we use null resources & curl. Ressource `gitlab_service_external_wiki` is needed :)

https://docs.gitlab.com/ee/api/integrations.html#external-wiki

Best regards

SeameX

- [ ] I would like to implement this myself 👷

Shocktrooper:
Resolves #996


internal/provider/resource_gitlab_service_external_wiki_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+		return nil
+	}
+}
+
+func testAccCheckGitlabServiceExternalWikiDestroy(s *terraform.State) error {
+	var project string
+
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type == "gitlab_project_environment" {
```
armsnyder, line 100: This looks like a copy-paste-bug

armsnyder (changes requested):
Thanks. Just found a test bug.

internal/provider/resource_gitlab_service_external_wiki_test.go
Shocktrooper: Gosh darned it. I copied part of the logic and forgot to change the resource 🥇

Shocktrooper: All fixed

armsnyder (approved):
lgtm!

armsnyder (changes requested):
Test failed

```
Error running post-test destroy, there may be dangling resources: [ERROR] External Wiki Service 89 still exists
```

Shocktrooper:
@armsnyder I wonder if this is the downstream API as looking at the code nothing is apparent.  I will check the GitLab API and see what it returns after some testing.
I found this in the docs
https://docs.gitlab.com/ee/api/integrations.html#:~:text=Disable%20the%20External%20wiki%20integration%20for%20a%20project.%20Integration%20settings%20are%20preserved.

Shocktrooper:
Yup it appears there is an active flag that is set to false on delete. I will refactor the resource to include this computed field

Shocktrooper:
@armsnyder After looking at the other integrations code I believe most if not all of them need to have their resource + unit tests updated to include the active flag + some computed fields. The unit tests for destruction verification seem to rely on the project being marked for deletion vs the integration being deactivated on top of a 404 being impossible to get after an integration is activated the first time.

armsnyder:
Thanks for updating the code to match the other services. There's a new test failure now:

```
resource_gitlab_service_external_wiki_test.go:24: Step 3/6 error: Check failed: Check 4/4 error: gitlab_service_external_wiki.this: Attribute 'updated_at' expected "2022-04-02T05:12:58Z", got "2022-04-02T05:13:01Z"
[277](https://github.com/gitlabhq/terraform-provider-gitlab/runs/5797570330?check_suite_focus=true#step:8:277)
```

armsnyder (approved):
lgtm thanks!

PR #1005, opened 2022-04-02
gitlab_project_access_token: Test bugfix and destroy improvement

### LINKED ISSUE #1002: gitlab_project_access_token - Could not provision maintainer access to project access token
### GitLab Provider version

_No response_

### GitLab version

_No response_

### Terraform version

_No response_

```hcl
resource "gitlab_project" "foo" {
  name = "foo-123"
  description = "Terraform acceptance tests"

  # So that acceptance tests can be run in a gitlab organization
  # with no billing
  visibility_level = "public"
}

resource "gitlab_project_access_token" "bar" {
  name = "my project token"
  project = gitlab_project.foo.id
  expires_at = "2022-04-01"
  scopes = ["read_repository" , "api", "write_repository", "read_api"]
}
```

### Relevant log output

```shell
=== RUN   TestAccGitlabProjectAccessToken_basic
166
    resource_gitlab_project_access_token_test.go:22: Step 1/6 error: Error running apply: exit status 1
167
        
168
        Error: POST http://127.0.0.1:8080/api/v4/projects/42/access_tokens: 400 {message: 400 Bad request - Could not provision maintainer access to project access token}
169
        
170
          with gitlab_project_access_token.bar,
171
          on terraform_plugin_test.tf line 11, in resource "gitlab_project_access_token" "bar":
172
          11: resource "gitlab_project_access_token" "bar" {
173
        
174
--- FAIL: TestAccGitlabProjectAccessToken_basic (6.45s)
175
=== RUN   TestAccGitlabProjectAccessToken_accessLevel
176
    resource_gitlab_project_access_token_test.go:102: Step 1/3 error: Error running apply: exit status 1
177
        
178
        Error: POST http://127.0.0.1:8080/api/v4/projects/44/access_tokens: 400 {message: 400 Bad request - Could not provision maintainer access to project access token}
179
        
180
          with gitlab_project_access_token.bar,
181
          on terraform_plugin_test.tf line 11, in resource "gitlab_project_access_token" "bar":
182
          11: resource "gitlab_project_access_token" "bar" {
183
        
184
--- FAIL: TestAccGitlabProjectAccessToken_accessLevel (6.48s)
```

### Description

The pipelines have been failing since merging #997 

See: https://github.com/gitlabhq/terraform-provider-gitlab/runs/5796647075?check_suite_focus=true

armsnyder:
## Description

Fixes #1002 

- Fixed a test bug where a hardcoded expiration date that passed caused failures
- Improved the destroy function to wait for the token to finish deleting

Note the diff to the `resourceGitlabProjectAccessTokenRead` looks large, but actually it is a straight move. I moved the list API calls into a `resourceGitlabProjectAccessTokenFind` function because I had to use it in the destroy logic also.

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/internal/provider/resource_gitlab_project_access_token_test.go
+++ b/internal/provider/resource_gitlab_project_access_token_test.go
@@ -2,378 +2,103 @@ package provider
 
 import (
 	"fmt"
-	"strconv"
 	"testing"
 	"time"
 
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
-	"github.com/onsi/gomega"
-	gitlab "github.com/xanzy/go-gitlab"
 )
 
 func TestAccGitlabProjectAccessToken_basic(t *testing.T) {
-	var pat testAccGitlabProjectAccessTokenWrapper
-	rInt := acctest.RandInt()
+	testAccCheck(t)
 
-	testAccGitlabProjectStart(t)
+	project := testAccCreateProject(t)
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:          func() { testAccPreCheck(t) },
 		ProviderFactories: providerFactories,
 		CheckDestroy:      testAccCheckGitlabProjectAccessTokenDestroy,
 		Steps: []resource.TestStep{
-			// Create a project and a Project Access Token
+			// Create a basic access token.
 			{
-				Config: testAccGitlabProjectAccessTokenConfig(rInt),
+				Config: fmt.Sprintf(`
+				resource "gitlab_project_access_token" "foo" {
+					project = %d
+					name    = "foo"
+					scopes  = ["api"]
+				}
+				`, project.ID),
+				// Check computed and default attributes.
 				Check: resource.ComposeTestCheckFunc(
-					testAccCheckGitlabProjectAccessTokenExists("gitlab_project_access_token.bar", &pat),
-					testAccCheckGitlabProjectAccessTokenAttributes(&pat, &testAccGitlabProjectAccessTokenExpectedAttributes{
-						name:        "my project token",
-						scopes:      map[string]bool{"read_repository": true, "api": true, "write_repository": true, "read_api": true},
-						expiresAt:   "2022-04-01",
-						accessLevel: accessLevelValueToName[gitlab.MaintainerPermissions], // default permission on gitlab side when unspecified
-					}),
+					resource.TestCheckResourceAttr("gitlab_project_access_token.foo", "active", "true"),
+					resource.TestCheckResourceAttr("gitlab_project_access_token.foo", "revoked", "false"),
+					resource.TestCheckResourceAttr("gitlab_project_access_token.foo", "access_level", "maintainer"),
+					resource.TestCheckResourceAttrSet("gitlab_project_access_token.foo", "token"),
+					resource.TestCheckResourceAttrSet("gitlab_project_access_token.foo", "created_at"),
+					resource.TestCheckResourceAttrSet("gitlab_project_access_token.foo", "user_id"),
+					resource.TestCheckNoResourceAttr("gitlab_project_access_token.foo", "expires_at"),
 				),
 			},
-			// Update the Project Access Token to change the parameters
+			// Verify upstream resource with an import.
 			{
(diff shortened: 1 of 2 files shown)


internal/provider/resource_gitlab_project_access_token_test.go
```diff
@@ -2,378 +2,103 @@ package provider
...
-						name:        "my project token",
-						scopes:      map[string]bool{"read_repository": true, "api": true, "write_repository": true, "read_api": true},
-						expiresAt:   "2022-04-01",
-						accessLevel: accessLevelValueToName[gitlab.DeveloperPermissions],
-					}),
+					resource.TestCheckResourceAttr("gitlab_project_access_token.foo", "active", "true"),
+					resource.TestCheckResourceAttr("gitlab_project_access_token.foo", "revoked", "false"),
+					resource.TestCheckResourceAttrSet("gitlab_project_access_token.foo", "token"),
+					resource.TestCheckResourceAttrSet("gitlab_project_access_token.foo", "created_at"),
+					resource.TestCheckResourceAttrSet("gitlab_project_access_token.foo", "user_id"),
```
PatrickRice-KSC, line 67: @armsnyder  - Given we're setting a "expires_at" with a specific value (and this is the only test that does so), it seems like we may want to test that the value matches what we pass in. I know there is an import test right after, but it seems an explicit check here may be good.

PatrickRice-KSC: Same, likely, with the access level (since those explicit tests were removed)

PatrickRice-KSC:
Very minor comment, otherwise LGTM. I think we should prioritize getting this merged so other PRs can rebase and ensure their checks pass, so let me know if there is anything I can do to help!

Comment for others - It makes sense to me that the "update" tests were largely removed, given pretty much every attribute of the access_token is a "ForceNew" attribute, and some of the newer tests were just wrapped into the second step of the "Basic" test. So even though the code looks like a bunch of lines were removed, I think this simplification covers the same general test cases. Given it's simpler, to me that means it's a good change!

internal/provider/resource_gitlab_project_access_token_test.go
armsnyder: After each TestStep, the framework re-runs `terraform plan` and fails the test if the plan is not empty, meaning that the values in state match what you specified in the config. Please correct me if I'm wrong, but I don't think there is a point to using `resource.TestCheckResourceAttr` functions to check what you set in the config.

armsnyder:
> It makes sense to me that the "update" tests were largely removed, given pretty much every attribute of the access_token is a "ForceNew" attribute

Yeah, and to add more context, since this resource has no "update" function at all, the Terraform SDK enforces that all attributes have `ForceNew: true`.

internal/provider/resource_gitlab_project_access_token_test.go
PatrickRice-KSC: @armsnyder - Ah, I knew it did that in the import steps, I didn't realize it did that automatically after every step, but I suppose that's why the `ExpectNonEmptyPlan` exists logically. 

Seems reasonable, I'm good.

PatrickRice-KSC:
> > It makes sense to me that the "update" tests were largely removed, given pretty much every attribute of the access_token is a "ForceNew" attribute
> 
> Yeah, and to add more context, since this resource has no "update" function at all, the Terraform SDK enforces that all attributes have `ForceNew: true`.

Learn new things every day, thanks for the info!

PatrickRice-KSC (approved):

PatrickRice-KSC:
@nagyv  - This is one that would be a great priority to merge, if you wouldn't mind, so we can have other PRs rebase and get their checks passing!

internal/provider/resource_gitlab_project_access_token_test.go
armsnyder: Oddly, in the link I shared, their examples do use `resource.TestCheckResourceAttr` to check the attributes set in the config. This is probably because there are edge cases where you might have a `DiffSuppressFunc` that renders the empty-plan check not useful.

My feeling is still that the combination of the test framework's features + using `ImportStateVerify` means we don't need to check these, since `ImportStateVerify` uses strict equality without respecting `DiffSuppressFunc` or `CustomizeDiff`, but I plan on opening an issue so we can discuss more.

Resolving this for now through.

timofurrer (approved):

timofurrer:
oh boy, what a fail 🤦 Thanks for fixing this @armsnyder 🎉

PR #1006, opened 2022-04-02
Add vscode settings for gitpod

armsnyder:
## Description

I found these settings useful while trying out our gitpod.io integration. It allows you to run or debug any test by clicking the play button, rather than requiring you to use the terminal.

Screen Shot 2022-04-02 at 1 23 37 AM

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,8 @@
+{
+    "go.testEnvVars": {
+        "TF_ACC": "1",
+        "GITLAB_TOKEN": "ACCTEST1234567890123",
+        "GITLAB_BASE_URL": "http://127.0.0.1:8080"
+    },
+    "go.testFlags": ["-count=1", "-v"]
+}


jtymes (dismissed):

PatrickRice-KSC:
@nagyv - would you be so kind as to merge with timo on vacation? 😄

PatrickRice-KSC (dismissed):

PatrickRice-KSC:
Actually, @armsnyder  - We may want to include a `.vscode/launch.json` file as well with the following contents:
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Run Tests",
            "type": "go",
            "request": "launch",
            "mode": "test",
            "remotePath": "",
            "program": "${workspaceFolder}/internal/provider",
            "env": {
                "TF_ACC": "1",
                "GITLAB_TOKEN": "ACCTEST1234567890123",
                "GITLAB_BASE_URL": "http://127.0.0.1:8080"
            },
            "showLog": true
        },
        
    ]
}
```

This will allow users to not only run individual tests, but to _debug_ them as well using delve.

armsnyder:
Hmm I was able to debug them without a launch.json by clicking the "debug" button in the screenshot. How would the launch.json be different?

PatrickRice-KSC:
Hm, let me test again. I was able to run them that way, but I wasn't able to set any breakpoints. Were breakpoints working for you with the base config?

armsnyder:
Yeah breakpoints worked when I tried.

PatrickRice-KSC:
Ok, testing with a fresh workspace just to be sure, but that's what I was targeting, so if that's working OK then I'm good without.

PatrickRice-KSC:
Hm, I still can't get debug breakpoints to run without the launch.json (they just sit and spin forever, which was the issue I was seeing last time I was trying). I'm good with merging this while we figure this out, as this doesn't break anything and it's clearly working ideally in some setups.

Just out of curiosity - what browser are you using? This seems like functionality that shouldn't at all be browser dependent, but just one more variable to remove.

armsnyder:
I remember it spinning for a while the first time I ran it. I'm using Safari. I'll try doing some more testing in a new workspace too.

armsnyder:
The first time I hit "debug", it took about 10-15 seconds to start executing, but it worked and hit the breakpoint. Logs from the debug tab:

```
Starting: [/home/gitpod/go-packages/bin/dlv]() dap --check-go-version=false --listen=127.0.0.1:32905 --log-dest=3 from [/workspace/terraform-provider-gitlab/internal/provider]()
DAP server listening at: 127.0.0.1:32905
2022-04-03T02:09:15Z error layer=debugger can't find build-id note on binary
Type 'dlv help' for list of commands.
```

PatrickRice-KSC:
I'm trying on chrome just for fun. I don't have a mac on me so it's difficult to try Safari; I waited for ~5 _minutes_ though, and got nothing - a matter of seconds wouldn't have be a problem for sure 😓.

I'd say we merge it the way you have it for now, and let's collect more information before we add more code/configs to maintain than necessary. I'm all for simplicity, and if I'm the only one having issues, then I normally use IntelliJ when I'm debugging anyway so I'm not one of the core target audience 😄

PatrickRice-KSC:
oh my. It works on chrome. So there's a browser quirk there. That's... not at all what I'd expect.

PatrickRice-KSC:
Just a comment for posterity - @armsnyder  and I spent some time walking through this together on Discord. Initially it seemed like some plugins may have been causing an issue, but after we both dumped our GitPod workspaces and started from scratch, and I did a completely fresh install of Firefox, it's working for both of us in Firefox as well. So off we go!

timofurrer (approved):

PR #1007, opened 2022-04-03
Initial Addition of Personal Access Token Resource + some refactors

### LINKED ISSUE #622: add personal_access_token resource
### Affected Resource(s)
- new: personal_access_token

With the GitLab Users API supporting personal access tokens for a user it would be helpful to be able to manage some of these via terraform.  This is especially helpful with service accounts as this token can then be used to configure a service in another application that is not GitLab

### Important Factoids

- go-gitlab already has the capability to create personal access tokens. 
- Revoking & Listing tokens for a user is a separate API call not apart of the users api. go-gitlab work would need to be done
- Before this can happen as we would want to be able to query the tokens that are created after the initial deployment
- This is applicable to all license tiers

Relevant Links

- https://github.com/xanzy/go-gitlab/blob/master/users.go#L838
- https://docs.gitlab.com/ee/api/personal_access_tokens.html
- https://docs.gitlab.com/ee/api/users.html#create-a-personal-access-token

Shocktrooper:
Resolves #622


Shocktrooper:
@armsnyder If you want the honors for review since you and I were talking about this on the attached issue.

armsnyder:
@Shocktrooper I'm just waiting for the pipeline to turn green before reviewing. Let me know if you need help before then. 🙆‍♂️

internal/provider/resource_gitlab_personal_access_tokens.go
```diff
@@ -0,0 +1,265 @@
...
+
+var validPersonalAccessTokenScopes = []string{
+	"api",
+	"read_user",
+	"read_api",
+	"read_repository",
+	"write_repository",
+	"read_registry",
+	"write_registry",
+	"sudo",
```
armsnyder, line 26: Where do you see "read_registry", "write_registry", and "sudo" documented? They aren't listed here: https://docs.gitlab.com/ee/api/users.html#create-a-personal-access-token

internal/provider/resource_gitlab_personal_access_tokens.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+var _ = registerResource("gitlab_personal_access_token", func() *schema.Resource {
+	return &schema.Resource{
+		Description: `The ` + "`gitlab_personal_access_token`" + ` resource allows to manage the lifecycle of a personal access token for a specified user.
+
+-> **Note**: In order to utilize this resource without failure the executing user must be an admin. 
+
+**Upstream API**: [GitLab REST API docs](https://docs.gitlab.com/ee/api/personal_access_tokens.html)
+**Upstream API**: [GitLab REST API docs](https://docs.gitlab.com/ee/api/users.html#create-a-personal-access-token)`,
```
armsnyder, line 36: It would be ok to only include the first link (https://docs.gitlab.com/ee/api/personal_access_tokens.html) here, since that page links to the other one.

By the way, try out the preview took here: https://registry.terraform.io/tools/doc-preview . Without a new line between each paragraph, you get the text on the same line.

internal/provider/resource_gitlab_personal_access_tokens.go
```diff
@@ -0,0 +1,265 @@
...
+		options.ExpiresAt = &parsedExpiresAtISOTime
+		log.Printf("[DEBUG] create gitlab PersonalAccessToken %s with expires_at %s for user ID %d", *options.Name, *options.ExpiresAt, userId)
+	}
+
+	personalAccessToken, _, err := client.Users.CreatePersonalAccessToken(userId, options, gitlab.WithContext(ctx))
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	log.Printf("[DEBUG] created gitlab PersonalAccessToken %d - %s for user ID %d", personalAccessToken.ID, *options.Name, userId)
```
armsnyder, line 136: nit: we don't usually log _after_ creating the resource

internal/provider/resource_gitlab_personal_access_tokens.go
```diff
@@ -0,0 +1,265 @@
...
+
+	if v, ok := d.GetOk("expires_at"); ok {
+		parsedExpiresAt, err := time.Parse("2006-01-02", v.(string))
+		if err != nil {
+			return diag.Errorf("Invalid expires_at date: %v", err)
+		}
+
+		parsedExpiresAtISOTime := gitlab.ISOTime(parsedExpiresAt)
+		options.ExpiresAt = &parsedExpiresAtISOTime
+		log.Printf("[DEBUG] create gitlab PersonalAccessToken %s with expires_at %s for user ID %d", *options.Name, *options.ExpiresAt, userId)
```
armsnyder, line 128: This will cause two "create"s to be logged. We should only log once before creating a resource.

internal/provider/resource_gitlab_personal_access_tokens.go
```diff
@@ -0,0 +1,265 @@
...
+
+	log.Printf("[DEBUG] Delete gitlab PersonalAccessToken %s", d.Id())
+	_, err = client.PersonalAccessTokens.RevokePersonalAccessToken(personalAccessTokenId, gitlab.WithContext(ctx))
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	log.Printf("[DEBUG] Waiting for PersonalAccessToken %s to finish deleting", d.Id())
+
+	err = resource.RetryContext(ctx, 5*time.Minute, func() *resource.RetryError {
```
armsnyder, line 221: Did you try it without this retry first? As long as the tests pass, we do not need it.

armsnyder (changes requested):

internal/provider/resource_gitlab_personal_access_tokens_test.go
```diff
@@ -0,0 +1,109 @@
...
+				// The token is only known during creating. We explicitly mention this limitation in the docs.
+				ImportStateVerifyIgnore: []string{"token"},
+			},
+			// Recreate the access token with updated attributes.
+			{
+				Config: fmt.Sprintf(`
+				resource "gitlab_personal_access_token" "foo" {
+					user_id    = %d
+					name       = "foo"
+					scopes     = ["api", "read_api", "read_repository", "write_repository"]
```
armsnyder, line 57: Related to my other comment, can you make sure you include all valid scopes in this line?

internal/provider/resource_gitlab_personal_access_tokens.go
Shocktrooper: Sudo only appears when you are an admin on a self-hosted GitLab instance and read/write registry are there for all users on SAAS and self-hosted.  It appears that GitLab's documentation is out of date for their API as the UI docs are up to date.
https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html#personal-access-token-scopes

internal/provider/resource_gitlab_personal_access_tokens.go
Shocktrooper: Will do! Thank you for that I had no idea that tool existed.  I usually use the right click preview in VS code but for some reason it is broken in chrome on a GitPod and results in a 401.

internal/provider/resource_gitlab_personal_access_tokens.go
Shocktrooper: I was borrowing from the project access token when doing this.  I shall remove it from there as well if there are no objections

internal/provider/resource_gitlab_personal_access_tokens.go
Shocktrooper: I will remove this one and also from the project access tokens if there are no objections.

internal/provider/resource_gitlab_personal_access_tokens.go
Shocktrooper: I thought there might be an underlying issue with responses in GitLab itself for tokens.  I can remove this and test it out, are we sure that that this isn't needed on a real GitLab instance?

internal/provider/resource_gitlab_personal_access_tokens_test.go
Shocktrooper: done also during this testing I found out something that is really off with the API. I will denote this in the thread above about scopes

internal/provider/resource_gitlab_personal_access_tokens.go
Shocktrooper: During my GitPod acceptance testing it appears that the retry is not needed

internal/provider/resource_gitlab_personal_access_tokens.go
Shocktrooper: So I found something really off with the underlying GitLab API. In the API docs it leaves off a few scopes (read_registry, write_registry, sudo). Of these missing from the API documentation only `sudo` can be set out of those scopes via the API. In the UI the end user is fully capable of adding `read_registry` and `write_registry` but errors out with `400 {error: scopes does not have a valid value}` if done via the provider/API for . I will try to find a ticket in GitLab to see if anyone else has noticed this inconsistency or if the API update is planned for a future release but is only UI settable for now

Shocktrooper: @nagyv or @DonNicoJs I was unable to find anything in the GitLab issue tracker can you see on your end if anyone is aware of the API inconsistency for user PAT scopes?

internal/provider/resource_gitlab_personal_access_tokens.go
armsnyder: Thanks!

internal/provider/resource_gitlab_personal_access_tokens.go
armsnyder: Thanks!

internal/provider/resource_gitlab_personal_access_tokens.go
armsnyder: I think we should remove it. I had to add it in the project_access_token resource because the `CheckDestroy` test failed consistently without it.

I added a comment to this GitLab issue here so we can get more info: https://gitlab.com/gitlab-org/gitlab/-/issues/352299#note_898951315

internal/provider/resource_gitlab_personal_access_tokens.go
armsnyder: Ok cool. This is the last remaining thing before I can approve. 😄

armsnyder (dismissed):

internal/provider/resource_gitlab_group_access_token.go
```diff
@@ -236,6 +227,6 @@ func resourceGitlabGroupAccessTokenDelete(ctx context.Context, d *schema.Resourc
 	}
 
 	log.Printf("[DEBUG] Delete gitlab GroupAccessToken %s", d.Id())
-	_, err = client.GroupAccessTokens.DeleteGroupAccessToken(group, groupAccessTokenId, gitlab.WithContext(ctx))
+	_, err = client.GroupAccessTokens.RevokeGroupAccessToken(group, groupAccessTokenId, gitlab.WithContext(ctx))
```
timofurrer, line 230: What's the reason for changing the behavior here? When we changed something kike that in other resources we considered jt a breaking change and added a flag to control the destroy behavior ... Wdyt?

internal/provider/resource_gitlab_project_access_token.go
```diff
@@ -197,7 +186,7 @@ func resourceGitlabProjectAccessTokenDelete(ctx context.Context, d *schema.Resou
 	}
 
 	log.Printf("[DEBUG] Delete gitlab ProjectAccessToken %s", d.Id())
-	_, err = client.ProjectAccessTokens.DeleteProjectAccessToken(project, projectAccessTokenID, gitlab.WithContext(ctx))
+	_, err = client.ProjectAccessTokens.RevokeProjectAccessToken(project, projectAccessTokenID, gitlab.WithContext(ctx))
```
timofurrer, line 189: Same thing here ...

internal/provider/resource_gitlab_group_access_token.go
Shocktrooper: This is just an upstream name change in go-gitlab. No functionality has changed as far as I can tell.

internal/provider/resource_gitlab_project_access_token.go
Shocktrooper: This is just an upstream name change in go-gitlab. No functionality has changed as far as I can tell.

internal/provider/resource_gitlab_group_access_token.go
armsnyder: Timo this was you! You renamed it! 😆

Shocktrooper: O.o I thought @svanharmelen did that as a tweak to my PR for Personal Access Tokens.
https://github.com/xanzy/go-gitlab/commit/4c72f522ec4f6a37769d8fbb6e942f2e5afd15fd

internal/provider/resource_gitlab_personal_access_tokens.go
Shocktrooper: As to not lose sight of this difference in the API vs UI scopes for the future how do you want to capture the difference?

internal/provider/resource_gitlab_project_access_token.go
svanharmelen: Yeah this was me, sorry for the trouble! Wanted to make the naming more inline with how the rest of the methods is used (the use the naming used in the docs)...

internal/provider/resource_gitlab_group_access_token.go
svanharmelen: It was me, see the other comment... Sorry for the inconvenience!

armsnyder: Oh gosh, sorry for my slander 😭

internal/provider/resource_gitlab_personal_access_tokens.go
Brcrwilliams: @Shocktrooper Hello, GitLab team member here. I didn't work on this API, but I took a look at the code and it validates the scopes using [`Gitlab::Auth#all_available_scopes`](https://gitlab.com/gitlab-org/gitlab/-/blob/25be580b697e5697df4288e5b5064bc1a3e3a480/lib/gitlab/auth.rb#L386). The `read_registry` and `write_registry` scopes are only available if the container registry is enabled on the GitLab instance. Since this API requires administrator, I assume you are testing this with a self-hosted instance. Does it have container registry enabled?

I've also opened https://gitlab.com/gitlab-org/gitlab/-/merge_requests/84272 to add the missing scopes to the documentation. Thanks for your feedback!

internal/provider/resource_gitlab_group_access_token.go
timofurrer: Hahah, was just going crazy over here about this 😅 thanks @svanharmelen for the rescue 👋

PatrickRice-KSC:
> Grumble Grumble So it appears that the test works perfectly in a GitPod but fails in GitHub actions. @armsnyder Is this workflow using the `main` branches setup files or the ones I committed to this branch?

The GitHub Actions for the `-ee` check uses the main branch for running the `make testacc-up` script, so it wouldn't be using your setup script. This prevents an issue where someone edits the makefile in their branch and exposes the license :( 

I would suggest we separate your setup script change into a different PR that we can merge, then rebase this to get the changes in. Thoughts?

Shocktrooper:
@PatrickRice-KSC Sounds good to me! Maybe I can make an annotation in the CONTRIBUTING.md in the future to say this is how to handle changes to setup files for testing

PatrickRice-KSC:
> Maybe I can make an annotation in the CONTRIBUTING.md in the future to say this is how to handle changes to setup files for testing

Makes sense to me!

Shocktrooper:
@PatrickRice-KSC Not sure if there is some setting to use the current files or something for setup on the main branch for feature branches but even with a pull of the separated code into this branch the tests are still failing on GItHub but succeeding in GitPods.  Only other thing I can think of is there is a race condition with enabling the container registry in the startup script and the unit test running

PatrickRice-KSC:
> but even with a pull of the separated code into this branch the tests are still failing on GItHub but succeeding in GitPods. 

Hm.... we'll probably want to check out the makefile to see if there is something we can put in the await_healthy script to ensure it's enabled. Seems weird to think that it would have a race condition, given that the test that's failing would be running after a bunch of time had passed (due to other tests running first), so maybe there is some secondary step needed to ensure it's enabled on the running container?

I'll try to take a look at it here in a couple of hours.

PatrickRice-KSC:
@Shocktrooper  - something weird is going on when running the tests locally if they're passing, because the change to the startup script doesn't enable the registry. I spun up your branch in gitpod, and here is the admin area:

image

when attempting to create a token from the instance, no read_registry or write_registry is available:

image


My guess is that this is because we don't have HTTPS set up for the instance, and GitLab all but refuses to start the container registry without HTTPS set up: https://docs.gitlab.com/ee/administration/packages/container_registry.html#enable-the-container-registry

We could probably mess around with the settings a bit to get it to start up the instance without HTTPS, and that's probably what we need to do in order for this to work :-/

PatrickRice-KSC:
Just taking a note here from our call: Shock and I had a quick discord call - there was some carryover (probably in the volume mount in docker-compose) that were enabling his local tests to work, and when you start with a fresh GitPod they fail in GitPod too. He's checking out maybe mapping in a `gitlab.rb` file instead of using the gitlab-rails comment to see if he can get it to work fresh 🙂

internal/provider/resource_gitlab_personal_access_tokens.go
Shocktrooper: So turns out it did not work and I needed to map a gitlab.rb file in. I had a false positive because of the volume mounts persisting between tests

timofurrer:
@Shocktrooper please rebase to include the latest GitLab instance setup including the proper certs.

internal/provider/resource_gitlab_personal_access_tokens.go
```diff
@@ -0,0 +1,242 @@
...
+				Sensitive:   true,
+			},
+		},
+	}
+})
+
+func resourceGitlabPersonalAccessTokenCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*gitlab.Client)
+
+	userId := d.Get("user_id").(int)
```
timofurrer, line 100: NIT: I suggest to get this attribute as close to where it's first used as possible ...

internal/provider/resource_gitlab_personal_access_tokens.go
```diff
@@ -0,0 +1,242 @@
...
+	log.Printf("[DEBUG] create gitlab PersonalAccessToken %s (scopes: %s) for user ID %d", *options.Name, options.Scopes, userId)
+
+	if v, ok := d.GetOk("expires_at"); ok {
+		parsedExpiresAt, err := time.Parse("2006-01-02", v.(string))
+		if err != nil {
+			return diag.Errorf("Invalid expires_at date: %v", err)
+		}
+
+		parsedExpiresAtISOTime := gitlab.ISOTime(parsedExpiresAt)
+		options.ExpiresAt = &parsedExpiresAtISOTime
```
timofurrer, line 125: we have a helper function for that:

```suggestion
		```suggestion
		parsedExpiresAt, err := parseISO8601Date(v.(string))
		if err != nil {
			return diag.Errorf("failed to parse expires_at '%s' as ISO8601 formatted date: %v", v.(string), err)
		}
		
		options.ExpiresAt = parsedExpiresAt
```
```

File: internal/provider/resource_gitlab_personal_access_tokens.go (hunk 9 above)
Comment (timofurrer) [reply internal/provider/resource_gitlab_personal_access_tokens.go:100]: Also, let's use `userID` instead of `userId` ...

File: internal/provider/resource_gitlab_personal_access_tokens.go [hunk 11]
```diff
@@ -0,0 +1,242 @@
...
+
+	personalAccessToken, _, err := client.Users.CreatePersonalAccessToken(userId, options, gitlab.WithContext(ctx))
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	tokenId := strconv.Itoa(personalAccessToken.ID)
+	userIdString := strconv.Itoa(userId)
+
+	d.SetId(buildTwoPartID(&userIdString, &tokenId))
```
timofurrer, line 136: tbh. I'd favor:

```suggestion
	d.SetId(fmt.Sprintf("%d:%d", userID, personalAccessToken.ID))
```

internal/provider/resource_gitlab_personal_access_tokens.go
```diff
@@ -0,0 +1,242 @@
...
+	if err != nil {
+		return diag.Errorf("Error converting user ID to string: %v", userId)
+	}
+
+	client := meta.(*gitlab.Client)
+
+	personalAccessTokenId, err := strconv.Atoi(tokenId)
+	if err != nil {
+		return diag.Errorf("%s cannot be converted to int", tokenId)
+	}
```
timofurrer, line 159: Can we refactor this into it's own function, like here:

https://github.com/gitlabhq/terraform-provider-gitlab/blob/4337c7437f809c2190d3e07e94137fa58b734729/internal/provider/resource_gitlab_project_issue.go#L240

internal/provider/resource_gitlab_personal_access_tokens.go
timofurrer: or refactor into own function like here:

https://github.com/gitlabhq/terraform-provider-gitlab/blob/4337c7437f809c2190d3e07e94137fa58b734729/internal/provider/resource_gitlab_project_issue.go#L254

internal/provider/resource_gitlab_personal_access_tokens.go
```diff
@@ -0,0 +1,242 @@
...
+	d.Set("name", personalAccessToken.Name)
+	if personalAccessToken.ExpiresAt != nil {
+		d.Set("expires_at", personalAccessToken.ExpiresAt.String())
+	}
+	d.Set("active", personalAccessToken.Active)
+	d.Set("created_at", personalAccessToken.CreatedAt.Format(time.RFC3339))
+	d.Set("revoked", personalAccessToken.Revoked)
+
+	err = d.Set("scopes", personalAccessToken.Scopes)
+	if err != nil {
```
timofurrer, line 183:
```suggestion
	if err = d.Set("scopes", personalAccessToken.Scopes); err != nil {
```

internal/provider/resource_gitlab_personal_access_tokens.go
```diff
@@ -0,0 +1,242 @@
...
+
+	personalAccessToken, err := resourceGitlabPersonalAccessTokenFind(ctx, client, userIdInt, personalAccessTokenId)
+	if errors.Is(err, errResourceGitlabPersonalAccessTokenNotFound) {
+		log.Printf("[DEBUG] failed to read gitlab PersonalAccessToken %d, user ID %s", personalAccessTokenId, userId)
+		d.SetId("")
+	}
+
+	if err != nil {
+		return diag.FromErr(err)
+	}
```
timofurrer, line 171: NIT: not a big fan of the readability here. I suggest to avoid fallthroughs like that and early exit:

```suggestion
	if errors.Is(err, errResourceGitlabPersonalAccessTokenNotFound) {
		log.Printf("[DEBUG] failed to read gitlab PersonalAccessToken %d, user ID %s", personalAccessTokenId, userId)
		d.SetId("")
		return nil
	}

	if err != nil {
		return diag.FromErr(err)
	}
```

timofurrer: ... it's even an issue, right?!

internal/provider/resource_gitlab_personal_access_tokens.go
```diff
@@ -0,0 +1,242 @@
...
+
+	return nil
+}
+
+func resourceGitlabPersonalAccessTokenDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	_, tokenId, err := parseTwoPartID(d.Id())
+	if err != nil {
+		return diag.Errorf("Error parsing ID: %s", d.Id())
```
timofurrer, line 193: Let's also give the user hints about the error at hand here:

```suggestion
		return diag.Errorf("Error parsing ID: %s: %v", d.Id(), err)
```

internal/provider/resource_gitlab_personal_access_tokens.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	client := meta.(*gitlab.Client)
+
+	personalAccessTokenId, err := strconv.Atoi(tokenId)
+	if err != nil {
+		return diag.Errorf("%s cannot be converted to int", tokenId)
```
timofurrer, line 200:
```suggestion
		return diag.Errorf("token id '%s' cannot be converted to int", tokenId)
```

internal/provider/resource_gitlab_personal_access_tokens.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	log.Printf("[DEBUG] Delete gitlab PersonalAccessToken %s", d.Id())
+	_, err = client.PersonalAccessTokens.RevokePersonalAccessToken(personalAccessTokenId, gitlab.WithContext(ctx))
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	return diag.FromErr(err)
```
timofurrer, line 209:
```suggestion
	return nil
```

internal/provider/resource_gitlab_personal_access_tokens_test.go
```diff
@@ -0,0 +1,109 @@
...
+					scopes  = ["api"]
+				}
+				`, user.ID),
+				// Check computed and default attributes.
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("gitlab_personal_access_token.foo", "active", "true"),
+					resource.TestCheckResourceAttr("gitlab_personal_access_token.foo", "revoked", "false"),
+					resource.TestCheckResourceAttrSet("gitlab_personal_access_token.foo", "token"),
+					resource.TestCheckResourceAttrSet("gitlab_personal_access_token.foo", "created_at"),
+					resource.TestCheckResourceAttrSet("gitlab_personal_access_token.foo", "user_id"),
```
timofurrer, line 39:
```suggestion
					resource.TestCheckResourceAttr("gitlab_personal_access_token.foo", "user_id", fmt.Sprintf("%d", user.ID)),
```

timofurrer (changes requested):

internal/provider/resource_gitlab_personal_access_tokens.go
```diff
@@ -0,0 +1,242 @@
...
+	"read_registry",
+	"write_registry",
+	"sudo",
+}
+
+var _ = registerResource("gitlab_personal_access_token", func() *schema.Resource {
+	return &schema.Resource{
+		Description: `The ` + "`gitlab_personal_access_token`" + ` resource allows to manage the lifecycle of a personal access token for a specified user.
+
+-> **Note** In order to utilize this resource without failure the executing user must be an admin.
```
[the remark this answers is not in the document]
Shocktrooper: Actually this is really weird. The doc preview automatically adds `Note` to the highlighted textbox if you don't specify it yourself but also omits an extra `Note` word if you do add it. This was using the docs preview tool

https://registry.terraform.io/tools/doc-preview

internal/provider/resource_gitlab_personal_access_tokens.go
Shocktrooper: Do we want to remove the helper function from `util.go` then and change all instances to the Sptrintf option?

internal/provider/resource_gitlab_personal_access_tokens.go
Shocktrooper: Should I also remove `[DEBUG]` like I did with `[ERROR]` or do debug statements print differently?

internal/provider/resource_gitlab_personal_access_tokens.go
Shocktrooper: This was changed in a previous review comment if you want a change to the abstracted function to do something like this please let me know

internal/provider/resource_gitlab_personal_access_tokens.go
Shocktrooper: Same as above comment with the abstracted function

internal/provider/resource_gitlab_personal_access_tokens.go
timofurrer: AFAIK: debug statements are different ...

internal/provider/resource_gitlab_personal_access_tokens.go
timofurrer: I'm not 100% sure yet, the function usually gives some more meaning to it and also consistently uses `:` as the delimiter. I just don't like the current verbosity ...

I suggest to use a separate function and there you may or may not use `buildTwoPartID`, but at least the resource CRUD function is concise. WDYT?

internal/provider/resource_gitlab_personal_access_tokens_test.go
Shocktrooper: Implemented

internal/provider/resource_gitlab_personal_access_tokens.go
Shocktrooper: Implemented

internal/provider/resource_gitlab_personal_access_tokens.go
Shocktrooper: Implemented

internal/provider/resource_gitlab_personal_access_tokens.go
Shocktrooper: Implemented

internal/provider/resource_gitlab_personal_access_tokens.go
Shocktrooper: Implemented

internal/provider/resource_gitlab_personal_access_tokens.go
```diff
@@ -0,0 +1,242 @@
...
+
+	userId := d.Get("user_id").(int)
+
+	currentUserAdmin, err := isCurrentUserAdmin(client)
+	if err != nil {
+		return diag.Errorf("[ERROR] cannot query the user API for current user: %v", err)
+	}
+
+	if !currentUserAdmin {
+		return diag.Errorf("[ERROR] Current user needs to be admin when creating a personal access token: %v", err)
```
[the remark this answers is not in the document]
Shocktrooper: Implemented

internal/provider/resource_gitlab_personal_access_tokens.go
Shocktrooper: Sounds good this is good to go then

internal/provider/resource_gitlab_personal_access_tokens.go
Shocktrooper: Mmmm agreed on the verbosity. I was looking at golang and couldn't find anything like overloading functions or anonymous type inputs.  I'm for more concise with the Sprintf and then during reviews we can catch people doing non-standard ID's and such.  We could also potentially do a custom linting rule for this as well

internal/provider/resource_gitlab_personal_access_tokens.go
timofurrer: the way you've changed it to is okay for me.

internal/provider/resource_gitlab_personal_access_tokens.go
timofurrer: the way you've changed it to is okay for me.

internal/provider/resource_gitlab_personal_access_tokens.go
timofurrer: the current version is fine for me for now :)

timofurrer (approved):

PR #608, opened 2021-04-22
support for can_push on gitlab_deploy_key and gitlab_deploy_key_enable

seboudry:
Hi!

Here's some changes to have `can_push` implemented on `gitlab_deploy_key_enable`.
Moreover, `can_push` actually works for `gitlab_deploy_key` so the documentation is updated.

Fix #294
Fix #577

Issue #1008, opened 2022-04-03
RFC - Testing alignment


armsnyder:
Looking at our test code, we have a number of different testing approaches that vary by resource. 😩 As long as new PRs are looking to existing testing patterns in the codebase, I think we should work to model good testing, so I'm hoping to get some feedback before doing some refactoring. 💃 

@timofurrer and I have discussed what we feel is a good approach for new tests. (Please jump in if I misrepresent anything Timo!) The theme is to make them simpler to write and review. We can delete hundreds of lines of test code per resource.

1. The test config should only include the resource/s being tested.

   Undesirable:
   ```terraform
   resource "gitlab_project" "foo" {
     name = "foo-%d"
     description = "Terraform acceptance tests"

     # So that acceptance tests can be run in a gitlab organization
     # with no billing
     visibility_level = "public"
   }

   resource "gitlab_managed_license" "foo" {
     project = "${gitlab_project.foo.id}"
     name = "MIT"
     approval_status = "%s"
   }
   ```

   Desirable:
   ```terraform
   resource "gitlab_managed_license" "foo" {
     project = "${gitlab_project.foo.id}"
     name = "MIT"
     approval_status = "%s"
   }
   ```

   Pros:
   * Allows us to use [`CheckDestroy`](https://www.terraform.io/plugin/sdkv2/testing/acceptance-tests/testcase#checkdestroy) correctly (see #205)
   * Less code to review
   * Isolates issues
   * Faster runtime (fewer API calls)

   Cons:
   * Requires a single call to `testAccCheck(t)` at the top of the test function when using helpers to set up the test using the GitLab client.

   Unknowns:
   * I haven't considered whether we should do this for data sources as well. With data sources there's high variability in the set-up API calls you would need to make (one per data source type).

1. Use [`ImportStateVerify`](https://github.com/hashicorp/terraform-plugin-sdk/blob/2c03a32a9d1be63a12eb18aaf12d2c5270c42346/helper/resource/testing.go#L482) TestSteps as a replacement for writing custom "CheckExists" and "CheckAttributes" check funcs.

   Undesirable:

   ```go
   {
   	Config: testAccGitlabFooConfig(rInt),
   	Check: resource.ComposeTestCheckFunc(
   		testAccCheckGitlabFooExists("gitlab_foo.foo", &foo),
   		testAccCheckGitlabFooAttributes(&foo, &testAccGitlabFooExpectedAttributes{
   			Foo: "bar",
[...]

PR #1009, opened 2022-04-04
Can push deploy keys

Shocktrooper:
Continuation of #608 that got stale


Shocktrooper:
@timofurrer After looking at the initial proposed changes the only thing I can think that might need changing from a cursory check is to implement some of the things from #1008 but that is up to you if you want to proceed with the current code or have me do those changes.

Shocktrooper:
@timofurrer I might have to rewrite part of this as the docs seem off to what the capabilities of the resource are on top of I had import failures when trying to implement things from #1008 . Also is there a reason to use a custom importer function for the deploy_key/deploy_key_enable resource vs just using `schema.ImportStatePassthroughContext` with a 2 part ID like in `resource_gitlab_project_environment.go`?

timofurrer:
@Shocktrooper I don't know what the initial reason to implement it like that was. What I can say is that we should whenever possible use the passthru importer - which would have been possible in this case. The reason for not changing it is to preserve backwards compatibility. (Which apparently has gone unnoticed for the slack resource, see #1012).

We could implement some logic to handle the break "gracefully", but since it's "only" a tech debt in this case I'd leave it as is and concentrate on other things and change it in the provider v4 :)

Shocktrooper:
@timofurrer & @pdecat & @armsnyder Is there a preferred ID we are going with going forward for integrations or even generally? Are we going straight for the project Id as also the integration ID or a 2 part ID like `projectId:integrationId`

timofurrer:
Well, since we want to use the passthru importer the id **must** contain all necessary information to read the resource at minimum. Sometimes more information is valuable (can't point to any example since my connectivity is very limited ...)

Shocktrooper:
@timofurrer and @pdecat I was looking at the resource and it looks a bit janky to me as I do not see a need for a custom import at all.  Form what I can tell we can use the  passthru importer without any changes to the ID because it contains all the necessary information already and does not need to be changed. Can I remove the custom importer functions or am I missing something?

Shocktrooper:
Also if it seems appropriate whomever reviews this please add it to the [v3.14.0](https://github.com/gitlabhq/terraform-provider-gitlab/milestone/12) milestone

timofurrer:
> @timofurrer and @pdecat I was looking at the resource and it looks a bit janky to me as I do not see a need for a custom import at all. Form what I can tell we can use the passthru importer without any changes to the ID because it contains all the necessary information already and does not need to be changed. Can I remove the custom importer functions or am I missing something?

The currently used resource id only contains the deploy key id, but you need the project id, too. Thus, without breaking the id (or using some migration strategy) it's not possible to use the passthru importer ...

Shocktrooper:
@timofurrer I made some refactors to move the deploy token to a 2 part ID which is consistent with how the resource is imported.  One thing I want to test and not sure how to do it so I am open to suggestions is to make an acceptance test for testing the ID migration path where someone might have had a 1 part legacy ID, then the next terraform run migrates that to the 2 part ID.

timofurrer:
I've played around a bit and turns out it's actually possible to use state migrations for resource ids 😮  I've submitted a draft PR of my scratch:

*  https://github.com/gitlabhq/terraform-provider-gitlab/pull/1033

Unless I'm missing something IMHO this should be the way to go - also for the slack integration resource we've done this ...

WDYT? 

cc @armsnyder

timofurrer:
However, I'm still not a fan of changing the id without releasing a new major.

It's also against the terraform provider development best practices, see: https://www.terraform.io/plugin/sdkv2/best-practices/versioning#:~:text=Changing%20resource%20import,resource%20ID%20format

IMHO: this should be part of the v4 milestone ...

Shocktrooper:
@timofurrer Alright I can revert it and we can wait till v4

timofurrer:
Yes, I suggest to focus here on the `can_push` attribute and do the id change later.

Shocktrooper:
@timofurrer I reverted the `deploy key` id changes but left the changes I made to `deploy key enable` id because that already had a 2 part id but instead of using the passthrough importer it had a custom importer that I believe was not necessary. Let me know what you think

internal/provider/resource_gitlab_deploy_key_enable.go
```diff
@@ -127,17 +142,21 @@ func resourceGitlabDeployKeyEnableDelete(ctx context.Context, d *schema.Resource
 	return nil
 }
 
-func resourceGitlabDeployKeyEnableStateImporter(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
-	s := strings.Split(d.Id(), ":")
-	if len(s) != 2 {
-		d.SetId("")
-		return nil, fmt.Errorf("Invalid Deploy Key import format; expected '{project_id}:{deploy_key_id}'")
+func resourceGitLabDeployKeyEnableParseId(id string) (int, int, error) {
```
timofurrer, line 145: The project id can actually be the `path_with_namespace` as string, too. Thus, this wouldn't work. Also the `d.Set("project", strconv.Itoa(project))` situations, won't work in those cases.

internal/provider/resource_gitlab_deploy_key_enable_test.go
```diff
@@ -2,145 +2,131 @@ package provider
...
 		PreCheck:          func() { testAccPreCheck(t) },
 		ProviderFactories: providerFactories,
 		CheckDestroy:      testAccCheckGitlabDeployKeyEnableDestroy,
 		Steps: []resource.TestStep{
-			// Create a project and deployKey with default options
+			// Enable a deployKey on project with default options
+			{
+				Config: testAccGitlabDeployKeyEnableConfig(testProjectKeyShared.ID, parentProjectDeployKey.ID),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet("gitlab_deploy_key_enable.foo", "project"),
```
timofurrer, line 37: since this is not a computed field and there is a subsequent `Verify import` test step it's superfluous ...

timofurrer: ... same for the other test steps ..

internal/provider/resource_gitlab_deploy_key_test.go
```diff
@@ -22,134 +23,86 @@ func TestAccGitlabDeployKey_basic(t *testing.T) {
...
+				ImportState:       true,
+				ImportStateVerify: true,
 			},
 		},
 	})
 }
 
-// lintignore: AT002 // TODO: Resolve this tfproviderlint issue
-func TestAccGitlabDeployKey_import(t *testing.T) {
+func TestAccGitlabDeployKey_suppresstrailingspace(t *testing.T) {
```
timofurrer, line 74:
```suggestion
func TestAccGitlabDeployKey_suppressTrailingSpace(t *testing.T) {
```

timofurrer (changes requested):

internal/provider/resource_gitlab_deploy_key_enable_test.go
Shocktrooper: Would you recommend we just remove the one check for project then across the test steps then is all?

timofurrer: Yes, but for all non-computed fields - thus, `can_push` is the same or the `title` in the `gitlab_deploy_key` resource ...

internal/provider/resource_gitlab_deploy_key_enable.go
Shocktrooper: I was unaware it was not converted to an int internally.  Do you have an example/suggestion of how to handle the case where a user would do an import with a `path_with_namespace ` and also I think the docs might need to be updated as well if that is allowable.  The import docs that I started working off of only has the projectID as an option for import.

timofurrer: You don't need to do anything special. Just use a string - `go-gitlab` will do the correct thing for you.

And indeed, the documentation can be improved there and mention that either a project path or id can be used. (it's correctly documented for the `project` argument already though.)

internal/provider/resource_gitlab_deploy_key_enable_test.go
Shocktrooper: Check if the commit I just pushed is what you were thinking when you have time

internal/provider/resource_gitlab_deploy_key_test.go
Shocktrooper: Fixed

internal/provider/resource_gitlab_deploy_key_enable_test.go
timofurrer: ... yes, right direction - however, the `can_push` and `title` etc. is still there, which can also be removed :)

Shocktrooper: Since there are no computed fields for this resource should I do just configs with no attribute checks and then just have import verifies in-between the test steps?

internal/provider/resource_gitlab_deploy_key_enable.go
Shocktrooper: I updated the import documentation and I believe strings should work now for the project attribute

internal/provider/resource_gitlab_deploy_key_enable_test.go
Shocktrooper: I have removed the non-computed fields from attr checks in the acceptance tests and added the title and key to the enable resource tests

internal/provider/resource_gitlab_deploy_key_enable_test.go
```diff
@@ -2,145 +2,129 @@ package provider
...
 		ProviderFactories: providerFactories,
 		CheckDestroy:      testAccCheckGitlabDeployKeyEnableDestroy,
 		Steps: []resource.TestStep{
-			// Create a project and deployKey with default options
+			// Enable a deployKey on project with default options
+			{
+				Config: testAccGitlabDeployKeyEnableConfig(testProjectKeyShared.ID, parentProjectDeployKey.ID),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet("gitlab_deploy_key_enable.foo", "key"),
+					resource.TestCheckResourceAttrSet("gitlab_deploy_key_enable.foo", "title"),
```
timofurrer, line 38: Those are also non-computed fields. The thing is that the verify import test step will actually check that the imported state will match the config, which implies that these attributes are correctly set into state.

timofurrer: e.g. for a little experiment: try to remove these checks here and also remove the line `105` in `internal/provider/resource_gitlab_deploy_key.go` where the `title` is set to state. The subsequent `Verify import` step at line 42 here should catch that :)

timofurrer (changes requested):

internal/provider/resource_gitlab_deploy_key_enable_test.go
timofurrer: also don't hesitate to ping me on Discord if you want to discuss this in detail :)

timofurrer (approved):
LGTM 🎉

PR #1023, opened 2022-04-12
Adding registry enablement fixes

Shocktrooper:
New Prerequisite to #1007

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -65,15 +65,19 @@ SERVICE ?= gitlab-ce
 GITLAB_TOKEN ?= ACCTEST1234567890123
 GITLAB_BASE_URL ?= http://127.0.0.1:8080/api/v4
 
-testacc-up: ## Launch a GitLab instance.
+testacc-up: | certs ## Launch a GitLab instance.
 	docker-compose up -d $(SERVICE)
 	./scripts/await-healthy.sh
 
 testacc-down: ## Teardown a GitLab instance.
-	docker-compose down
+	docker-compose down --volumes
 
 testacc: ## Run acceptance tests against a GitLab instance.
 	TF_ACC=1 GITLAB_TOKEN=$(GITLAB_TOKEN) GITLAB_BASE_URL=$(GITLAB_BASE_URL) go test -v $(PROVIDER_SRC_DIR) $(TESTARGS) -timeout 40m
+
+certs: ## Generate certs for the GitLab container registry
+	mkdir -p certs
+	openssl req -x509 -newkey rsa:4096 -sha256 -days 3650 -nodes -keyout certs/gitlab-registry.key -out certs/gitlab-registry.crt -subj "/CN=gitlab-registry.com" -addext "subjectAltName=DNS:IP:127.0.0.1"
 
 # TOOLS
 # Tool dependencies are installed into a project-local /bin folder.
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -108,3 +108,6 @@ TERRAFORM_VERSION = v1.1.4
 tool-terraform:
 	@# See https://github.com/hashicorp/terraform/issues/30356
 	@[ -f $(GOBIN)/terraform ] || { mkdir -p tmp; cd tmp; rm -rf terraform; git clone --branch $(TERRAFORM_VERSION) --depth 1 https://github.com/hashicorp/terraform.git; cd terraform; GOBIN=$(GOBIN) go install; cd ..; rm -rf terraform; }
+
+clean: testacc-down
+	@rm -rf certs/
--- a/scripts/gitlab.rb
+++ b/scripts/gitlab.rb
@@ -1,0 +1,4 @@
+registry_external_url 'http://127.0.0.1:5050'
+registry['enable']                    = true
+registry_nginx['ssl_certificate']     = "/etc/gitlab/ssl/gitlab-registry.pem"
+registry_nginx['ssl_certificate_key'] = "/etc/gitlab/ssl/gitlab-registry.key"
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@ website/build
 website/node_modules
 website/vendor
 /tmp
+certs/gitlab-registry.crt
+certs/gitlab-registry.key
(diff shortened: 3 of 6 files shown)


Shocktrooper:
@PatrickRice-KSC If you want the honors

PatrickRice-KSC (dismissed):

PatrickRice-KSC:
Thanks @Shocktrooper  - This matches our conversation from earlier 🙂

GNUmakefile
```diff
@@ -66,11 +66,13 @@ GITLAB_TOKEN ?= ACCTEST1234567890123
 GITLAB_BASE_URL ?= http://127.0.0.1:8080/api/v4
 
 testacc-up: ## Launch a GitLab instance.
+	mkdir -p certs
+	openssl req -x509 -newkey rsa:4096 -sha256 -days 3650 -nodes -keyout certs/example.key -out certs/example.crt -subj "/CN=example.com" -addext "subjectAltName=DNS:IP:127.0.0.1"
```
timofurrer, line 70: Can we rename the certificate files (and probably also the CN) to something "more meaningful" ? It'snot really an example after all :)

timofurrer: maybe also put them into a hidden or temporary folder?

GNUmakefile
```diff
@@ -66,11 +66,13 @@ GITLAB_TOKEN ?= ACCTEST1234567890123
 GITLAB_BASE_URL ?= http://127.0.0.1:8080/api/v4
 
 testacc-up: ## Launch a GitLab instance.
```
timofurrer, line 68: Maybe add a specific target to generate the certificate files and have a proper make dependency on them to not generate them every time?

timofurrer (changes requested):
some minor requests, otherwise looks good 👍

GNUmakefile
Shocktrooper: for the certs they are in a temporary folder and the .gitignore already has those listed so they are not committed.  I can rename the certs to something better

GNUmakefile
Shocktrooper: What do you think the benefit might be to not generating them every time? I also tried looking at the openssl flags and couldn't find any options to prevent overwrite if there were pre-existing certs

GNUmakefile
Shocktrooper: changed it up a bit hopefully this is more of what you were thinking of

GNUmakefile
Shocktrooper: I ended up adding some file tests so hopefully that aligns more to what you were thinking

GNUmakefile
timofurrer: > for the certs they are in a temporary folder and the .gitignore already has those listed so they are not committed.  I can rename the certs to something better

I was more thinking about a real "temporary" one (e.g. inside `$TEMPDIR`) but it's ok for a first merge :)

GNUmakefile
timofurrer: > What do you think the benefit might be to not generating them every time? I also tried looking at the openssl flags and couldn't find any options to prevent overwrite if there were pre-existing certs

It would not hurt, but is just superfluous :) And I had real make goal dependencies in mind ... but again here - not a show stopper ...

timofurrer: Somethink li,e this: https://stackoverflow.com/a/21754930

Shocktrooper: Implemented

GNUmakefile
```diff
@@ -65,16 +65,20 @@ SERVICE ?= gitlab-ce
...
 	./scripts/await-healthy.sh
 
 testacc-down: ## Teardown a GitLab instance.
-	docker-compose down
+	docker-compose down --volumes
 
 testacc: ## Run acceptance tests against a GitLab instance.
 	TF_ACC=1 GITLAB_TOKEN=$(GITLAB_TOKEN) GITLAB_BASE_URL=$(GITLAB_BASE_URL) go test -v $(PROVIDER_SRC_DIR) $(TESTARGS) -timeout 40m
 
+certs: ## Generate certs for the GiitLab container registry
```
timofurrer, line 78:
```suggestion
certs: ## Generate certs for the GitLab container registry
```

scripts/gitlab.rb
```diff
@@ -0,0 +1,4 @@
+registry_external_url 'http://127.0.0.1:5050'
+registry['enable']                    = true
+registry_nginx['ssl_certificate']     = "/etc/gitlab/ssl/example.pem"
```
timofurrer, line 3: The files need to be renamed here, too.

On that note: should we have a test for the correct setup of the instance? Instead of having test failures later on? We can also do this outside of this PR to unblock things ...

GNUmakefile
```diff
@@ -65,16 +65,20 @@ SERVICE ?= gitlab-ce
 GITLAB_TOKEN ?= ACCTEST1234567890123
 GITLAB_BASE_URL ?= http://127.0.0.1:8080/api/v4
 
-testacc-up: ## Launch a GitLab instance.
+testacc-up: | certs ## Launch a GitLab instance.
```
timofurrer, line 68: It might be better to depend on the files instead of the folder ...

timofurrer (changes requested):

scripts/gitlab.rb
Shocktrooper: I'm thinking that might not be a bad idea.  Would we do it in the `await-healthy.sh` or the `healthcheck-and-setup.sh`

GNUmakefile
Shocktrooper: So my only thought on leaving as the folder is because the order-only dependency seems to be linked to individual files or individual directories vs a group of files.  So for instance if I were to have `| certs/gitlab-registry.crt certs/gitlab-registry.key` I would have to have 2 makefile targets but underneath them I would still be using the openssl command to regenerate both certs.

scripts/gitlab.rb
timofurrer: Maybe there or in an acceptance test. Let's do this out of scope of this PR to unblock the other things. I'll open an issue.

timofurrer (approved):
I took the liberty to fix the typo in the `gitlab.rb` in the cert filenames and added some note about openssl in the contributing guide and added a `clean` make target.

LGTM 🎉

PR #1026, opened 2022-04-17
Provide `gitlab_pages_domain` resource

### LINKED ISSUE #498: Feature request - Expose DNS records used in Pages configuration to rest of Terraform
Currently configuring Pages with a custom domain requires a manual intervention.

### Terraform Version
```
Terraform v0.13.5
```

### Affected Resource(s)
If applicable: 
- gitlab_project
- gitlab_project_pages (if it will exist in next release)

### Actual Behavior
Jane is working on a static site with GitLab Pages and already has a DNS provider that supports Terraform. She has to manually copy and paste DNS records for her custom domain from the GitLab console into her Terraform code, which is **troublesome and error prone**.

### Expected Behavior
Jane wishes to get the following from the GitLab provider:
- CNAME DNS record that directs traffic from her domain to Pages;
- TXT DNS record used for verification;

by ex. using a **data source**. That way she could **pass those records as variables** to the provider of her DNS company. This scales much better and is much less error prone than semi-automatic configuration.

### Important Factoids
GitLab provider version:
- 3.1.0
- 3.2.0

### References

- Similar case, inproving functionality of Pages for Terraform users: https://github.com/gitlabhq/terraform-provider-gitlab/issues/471
- Documentation: https://docs.gitlab.com/ee/user/project/pages/custom_domains_ssl_tls_certification/index.html#4-verify-the-domains-ownership

nagyv:
## Description

Create a `gitlab_pages_domain` resource

Closes #498

Blocked by https://github.com/xanzy/go-gitlab/pull/1452

  - Resource attributes match 1:1 the names and structure of the API resource in [the GitLab API documentation](https://docs.gitlab.com/ee/api/). ✅ 
    - Create the resource ✅ 
    - Update the attributes ✅ 
    - Import the resource
  - No new `//lintignore` comments were copied from existing code. (Linter rules are meant to be enforced on new code.) ✅


nagyv:
I expect the tests to fail for now. I run into two issues with the tests:

1. Running `make testacc RUN=TestAccGitlabPagesDomain_basic` returns error 404
2. Running `make testacc RUN=TestAccGitlabPagesDomain_basic GITLAB_TOKEN=glpat-xxx GITLAB_BASE_URL=https://gitlab.com` returns error 400. 

The latter led to https://github.com/xanzy/go-gitlab/pull/1452

Why was there a different error in GitPod? Are GitLab pages enabled in GitPod?

PatrickRice-KSC:
Hey @nagyv , yeah, pages isn't configured in the docker-compose/makefile yet. We could potentially add a new wildcard cert to the ones generated by openssl (should probably wait until we merge #1023 so we can build on top of the work shock is doing), which seems to be a pre-req for enabling pages based on: https://docs.gitlab.com/ee/administration/pages/index.html#configuration

image

timofurrer:
FYI @PatrickRice-KSC @nagyv I've merged #1023 🎉

internal/provider/resource_gitlab_pages_domain.go
```diff
@@ -0,0 +1,220 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	gitlab "github.com/xanzy/go-gitlab"
+)
+
+var _ = registerResource("gitlab_pages_domain", func() *schema.Resource {
+	return &schema.Resource{
+		Description: `The ` + "`gitlab_pages_domain`" + ` resource enables endpoints for connecting custom domain(s) and TLS certificates in GitLab Pages.
```
timofurrer, line 16:
```suggestion
		Description: `The ` + "`gitlab_pages_domain`" + ` resource allows to manage the lifecycle of a custom Pages domain including its TLS certificates.
```

internal/provider/resource_gitlab_pages_domain.go
```diff
@@ -0,0 +1,220 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"domain": {
+				Description: "The custom domain.",
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+			},
+			"project": {
+				Description: "The ID or full path of the project which the branch is created against.",
```
timofurrer, line 36:
```suggestion
				Description: "The ID or [URL-encoded path of the project](https://docs.gitlab.com/ee/api/index.html#namespaced-path-encoding) owned by the authenticated user.",
```

internal/provider/resource_gitlab_pages_domain.go
```diff
@@ -0,0 +1,220 @@
...
+				Type:        schema.TypeString,
+				ForceNew:    true,
+				Required:    true,
+			},
+			"auto_ssl_enabled": {
+				Description: "Enables automatic generation of SSL certificates issued by Let’s Encrypt for custom domains.",
+				Type:        schema.TypeBool,
+				Optional:    true,
+				ForceNew:    false,
+				Default:     nil,
```
timofurrer, line 46: NIT: we usually do not specify the defaults

```suggestion
```

internal/provider/resource_gitlab_pages_domain.go
```diff
@@ -0,0 +1,220 @@
...
+			},
+			"certificate": {
+				Description: "The certificate in PEM format with intermediates following in most specific to least specific order.",
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    false,
+				Default:     nil,
```
timofurrer, line 53: NIT: we usually do not specify the defaults

```suggestion
```

internal/provider/resource_gitlab_pages_domain.go
```diff
@@ -0,0 +1,220 @@
...
+			},
+			"key": {
+				Description: "The certificate key in PEM format.",
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    false,
+				Default:     nil,
```
timofurrer, line 60: NIT: we usually do not specify the defaults

```suggestion
```

internal/provider/resource_gitlab_pages_domain.go
```diff
@@ -0,0 +1,220 @@
...
+			},
+			"certificate_data": {
+				Description: "The certificate data.",
+				Type:        schema.TypeList,
+				MaxItems:    1,
+				Computed:    true,
+				Optional:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"expired": {
```
timofurrer, line 75: What about the `subject` ?

internal/provider/resource_gitlab_pages_domain.go
```diff
@@ -0,0 +1,220 @@
...
+				Description: "The certificate data.",
+				Type:        schema.TypeBool,
+				Computed:    true,
+			},
+			"verification_code": {
+				Description: "The verification code for the domain.",
+				Type:        schema.TypeString,
+				Computed:    true,
+				Sensitive:   true,
+			},
```
timofurrer, line 101: Are these undocumented fields?

internal/provider/resource_gitlab_pages_domain.go
```diff
@@ -0,0 +1,220 @@
...
+		},
+	}
+})
+
+func resourceGitlabPagesDomainCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*gitlab.Client)
+	projectId := d.Get("project").(string)
```
timofurrer, line 108: NIT:

```suggestion
	projectID := d.Get("project").(string)
```

internal/provider/resource_gitlab_pages_domain.go
```diff
@@ -0,0 +1,220 @@
...
+	options := &gitlab.CreatePagesDomainOptions{
+		Domain:         &domain,
+		AutoSslEnabled: &auto_ssl_enabled,
+		Certificate:    &certificate,
+		Key:            &key,
+	}
+
+	log.Printf("[DEBUG] create gitlab pages domain %s", domain)
+
+	_, _, err := client.PagesDomains.CreatePagesDomain(projectId, options, gitlab.WithContext(ctx))
```
timofurrer, line 123: NIT:

```suggestion
	_, _, err := client.PagesDomains.CreatePagesDomain(projectID, options, gitlab.WithContext(ctx))
```

internal/provider/resource_gitlab_pages_domain.go
```diff
@@ -0,0 +1,220 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(buildTwoPartID(&projectId, &domain))
```
timofurrer, line 128: NIT:

```suggestion
	d.SetId(buildTwoPartID(&projectID, &domain))
```

internal/provider/resource_gitlab_pages_domain.go
```diff
@@ -0,0 +1,220 @@
...
+	auto_ssl_enabled := d.Get("auto_ssl_enabled").(bool)
+	certificate := d.Get("certificate").(string)
+	key := d.Get("key").(string)
+
+	options := &gitlab.CreatePagesDomainOptions{
+		Domain:         &domain,
+		AutoSslEnabled: &auto_ssl_enabled,
+		Certificate:    &certificate,
+		Key:            &key,
+	}
```
timofurrer, line 119: NIT: to indicate that we only use these attributes for the GitLab options I think it makes sense to not have variables for them ...

```suggestion
	domain := d.Get("domain").(string)
	auto_ssl_enabled := d.Get("auto_ssl_enabled").(bool)
	certificate := d.Get("certificate").(string)
	key := d.Get("key").(string)

	options := &gitlab.CreatePagesDomainOptions{
		Domain:         gitlab.String(d.Get("domain").(string)),
		AutoSslEnabled: gitlab.Bool(d.Get("auto_ssl_enabled").(bool))
		Certificate:    gitlab.String(d.Get("certificate").(string)),
		Key:            gitlab.String(d.Get("key").(string)),
	}
```

Please verify the correct behavior during creation if these optional attributes are not given and that the GitLab API doesn't receive them ....

internal/provider/resource_gitlab_pages_domain.go
```diff
@@ -0,0 +1,220 @@
...
+
+	d.SetId(buildTwoPartID(&projectId, &domain))
+	return resourceGitlabPagesDomainRead(ctx, d, meta)
+}
+
+func resourceGitlabPagesDomainUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*gitlab.Client)
+
+	projectID := d.Get("project").(string)
```
timofurrer, line 135: It's a common practise to use the `id` here as well, like you did for the `Read()`

internal/provider/resource_gitlab_pages_domain.go
```diff
@@ -0,0 +1,220 @@
...
+	domain := d.Get("domain").(string)
+	auto_ssl_enabled := d.Get("auto_ssl_enabled").(bool)
+	certificate := d.Get("certificate").(string)
+	key := d.Get("key").(string)
+
+	options := &gitlab.UpdatePagesDomainOptions{
```
timofurrer, line 141: I'd only update the fields for which the attribute in the tf config actually changed.

You can use e.g. `d.HasChange("auto_ssl_enabled")` to check that and add it to the `options`.

internal/provider/resource_gitlab_pages_domain.go
```diff
@@ -0,0 +1,220 @@
...
+	return nil
+}
+
+func flattenCertificateData(pagesDomain *gitlab.PagesDomain) (certificate_data []map[string]interface{}) {
+	if pagesDomain == nil {
+		return
+	}
+
+	var certificate_expiration string
+	if pagesDomain.Certificate.Expiration == nil {
```
timofurrer, line 207: Can `pagesDomain.Certificate` be `nil` ?

internal/provider/resource_gitlab_pages_domain_test.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/xanzy/go-gitlab"
+)
+
+func TestAccGitlabPagesDomain_basic(t *testing.T) {
+	var pagesDomain gitlab.PagesDomain
+	rInt := acctest.RandInt()
+	testAccCheck(t)
```
timofurrer, line 16: NIT: early exit asap.

```suggestion
	testAccCheck(t)
	
	var pagesDomain gitlab.PagesDomain
	rInt := acctest.RandInt()
```

internal/provider/resource_gitlab_pages_domain_test.go
```diff
@@ -0,0 +1,124 @@
...
+	project := testAccCreateProject(t)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: providerFactories,
+		CheckDestroy:      testAccCheckGitlabPagesDestroy,
+		Steps: []resource.TestStep{
+			// Create a pages domain with all options
```
timofurrer, line 24: Let's add another test case to only supply the required data - there may be some issues in the current implementation when `pagesDomain.Certificate` is `nil` or even earlier when `go-gitlab` fails, because `nil` is passed to `gitlab.String()` - these are just suspicions though and everything might be okay as-is 👯

internal/provider/resource_gitlab_pages_domain_test.go
```diff
@@ -0,0 +1,124 @@
...
+			{
+				Config: testAccGitlabPagesDomainCreate(rInt, project.PathWithNamespace),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGitlabPagesDomainExists("gitlab_pages_domain.this", &pagesDomain),
+					resource.TestCheckResourceAttrSet("gitlab_pages_domain.this", "auto_ssl_enabled"),
```
timofurrer, line 29: Given the subsequent `Verify import` test step both of these checks are already done, because:

1. `import` does a `Read()` and therefore the Pages domain actually has to exist
2. it would result in a non-empty plan in case the `auto_ssl_enabled` is not properly set ...

internal/provider/resource_gitlab_pages_domain_test.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func testAccGitlabPagesDomainUpdate(rInt int, project string) string {
+	return fmt.Sprintf(`
+resource "gitlab_pages_domain" "this" {
+  project                     = "%[2]s"
+  domain                      = "page-%[1]d.example.com"
+  auto_ssl_enabled            = true
+}
+	`, rInt+1, project)
```
timofurrer, line 123: In case they don't grow after the feedback implementation I think we can inline them for better reason about the test cases.

timofurrer (changes requested):
A few minor feedbacks and nit picks no linter is able to catch yet ...

Looks solid otherwise 🎉

internal/provider/resource_gitlab_pages_domain_test.go
nagyv: Sorry, I don't understand your suggestion. :)

internal/provider/resource_gitlab_pages_domain_test.go
nagyv: What is your recommendation? I would prefer to have a "create" test case, beside the import.

internal/provider/resource_gitlab_pages_domain.go
nagyv: Even if we remove the default, I'd keep the `ForceNew`. The docs were not clear what the behaviour is if someone changes the `auto_ssl_enabled` flag. Re-creating the domain seems to be a safe bet.

internal/provider/resource_gitlab_pages_domain.go
nagyv: I'd leave the ForceNew here too. I might be wrong, but the `certificate_data` is a computed property based on the `certificate`. The plan might be misleading without ForceNew, as `certificate_data` will be known only after the apply step. I guess ForceNew might achieve this.

internal/provider/resource_gitlab_pages_domain.go
nagyv: I added ForceNew for the same reason as with the certificate. Let me know if my reasoning is wrong. I likely know less about terraform internals than you do.

internal/provider/resource_gitlab_pages_domain.go
nagyv: https://github.com/xanzy/go-gitlab/blob/master/pages_domains.go#L44 returns only the expired and expiration keys.

internal/provider/resource_gitlab_pages_domain.go
nagyv: Yes. https://github.com/gitlabhq/terraform-provider-gitlab/issues/498#issuecomment-734532593

internal/provider/resource_gitlab_pages_domain.go
nagyv: Do you mean separate testcases? The current test does not specify any of the optional attributes.

internal/provider/resource_gitlab_pages_domain.go
nagyv: No. I tried that first, but it's always initialized with `Expired: false, Expiration: nil`

internal/provider/resource_gitlab_pages_domain_test.go
nagyv: I removed the `Default` values and changed to testcase to include only the required ones.

internal/provider/resource_gitlab_pages_domain.go
timofurrer: hmm, I would assume that if it can be updated, it safe from the API side, now? 

But then we should set `ForceNew: true` anyways, and not `ForceNew: false`, right?

internal/provider/resource_gitlab_pages_domain.go
timofurrer: I don't get that 😅  ... I mean `ForceNew: false` is the default anyways, I'm just saying we should remove from the schema - so that it's non-explicit.

internal/provider/resource_gitlab_pages_domain.go
timofurrer: ... which is somewhat misaligned with the upstream API ... 

In case I have some time tomorrow I'll update the GitLab docs and `go-gitlab`. Otherwise, I think we go with what's there :)

internal/provider/resource_gitlab_pages_domain.go
timofurrer: True, noticed afterwards - let's make sure that both scenarios are tested: creation without optionals set, creation with optionals set ...

internal/provider/resource_gitlab_pages_domain_test.go
timofurrer: oh sorry, this might help: https://github.com/gitlabhq/terraform-provider-gitlab/issues/1008#:~:text=Prefer%20writing%20config%20in%2Dline%20to%20wrapping%20it%20in%20a%20function.

timofurrer:
@nagyv btw. I think something got messed up during the rebase ....

internal/provider/resource_gitlab_pages_domain_test.go
```diff
@@ -0,0 +1,123 @@
...
+					testAccCheckGitlabPagesDomainExists("gitlab_pages_domain.this", &pagesDomain),
+					resource.TestCheckResourceAttrSet("gitlab_pages_domain.this", "auto_ssl_enabled"),
+				),
+			},
+			// Verify import
+			{
+				ResourceName:            "gitlab_pages_domain.this",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"token"},
```
timofurrer, line 38:
```suggestion
```

There is not `token`

internal/provider/resource_gitlab_pages_domain_test.go
timofurrer: As per my other comment, let's have multiple test cases: testing w/ and w/o optionals set.

internal/provider/resource_gitlab_pages_domain_test.go
timofurrer: Maybe this helps: [Use ImportStateVerify TestSteps as a replacement for writing custom "CheckExists" and "CheckAttributes" check funcs.](https://github.com/gitlabhq/terraform-provider-gitlab/issues/1008#:~:text=and%20data%20sources.-,Use%20ImportStateVerify%20TestSteps%20as%20a%20replacement%20for%20writing%20custom%20%22CheckExists%22%20and%20%22CheckAttributes%22%20check%20funcs.,-Undesirable%3A)

Also see the best practice here: https://github.com/bflad/tfproviderlint/tree/main/passes/AT002#at002

internal/provider/resource_gitlab_pages_domain.go
```diff
@@ -0,0 +1,217 @@
...
+				Description: "The certificate in PEM format with intermediates following in most specific to least specific order.",
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    false,
+			},
+			"key": {
+				Description: "The certificate key in PEM format.",
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    false,
```
nagyv, line 57: Remove default

```suggestion
```

internal/provider/resource_gitlab_pages_domain.go
```diff
@@ -0,0 +1,217 @@
...
+				Description: "Enables automatic generation of SSL certificates issued by Let’s Encrypt for custom domains.",
+				Type:        schema.TypeBool,
+				Optional:    true,
+				ForceNew:    false,
+			},
+			"certificate": {
+				Description: "The certificate in PEM format with intermediates following in most specific to least specific order.",
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    false,
```
nagyv, line 51: Remove default

```suggestion
```

internal/provider/resource_gitlab_pages_domain.go
```diff
@@ -0,0 +1,217 @@
...
+				Description: "The ID or [URL-encoded path of the project](https://docs.gitlab.com/ee/api/index.html#namespaced-path-encoding) owned by the authenticated user.",
+				Type:        schema.TypeString,
+				ForceNew:    true,
+				Required:    true,
+			},
+			"auto_ssl_enabled": {
+				Description: "Enables automatic generation of SSL certificates issued by Let’s Encrypt for custom domains.",
+				Type:        schema.TypeBool,
+				Optional:    true,
+				ForceNew:    false,
```
nagyv, line 45: Remove default

```suggestion
```

PR #1034, opened 2022-04-20
resource/gitlab_pipeline_trigger: Mark `token` attribute as sensitive

timofurrer:
The `gitlab_pipeline_trigger` `token` attribute is actually a sensitive value. I think it's okay to mark it as sensitive in a minor release.


armsnyder (approved):
Sounds good, thanks!

PR #1041, opened 2022-04-24
Enable pages

nagyv:
## Description

Enables GitLab Pages. Required for https://github.com/gitlabhq/terraform-provider-gitlab/pull/1026

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/scripts/gitlab.rb
+++ b/scripts/gitlab.rb
@@ -1,3 +1,8 @@
+pages_external_url 'http://127.0.0.1:5051'
+pages_nginx['redirect_http_to_https'] = false
+pages_nginx['ssl_certificate'] = "/etc/gitlab/ssl/gitlab-registry.pem"
+pages_nginx['ssl_certificate_key'] = "/etc/gitlab/ssl/gitlab-registry.key"
+
 registry_external_url 'http://127.0.0.1:5050'
 registry['enable']                    = true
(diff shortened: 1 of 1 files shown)


timofurrer (approved):
LGTM 🎉

scripts/gitlab.rb
```diff
@@ -1,3 +1,8 @@
+pages_external_url 'http://127.0.0.1:5051'
+pages_nginx['redirect_http_to_https'] = false
+pages_nginx['ssl_certificate'] = "/etc/gitlab/ssl/gitlab-registry.pem"
```
timofurrer, line 3: I kinda dislike the naming of the re-used cert files here - since it's not only for the registry anymore. 

Also, is it configured properly for the pages to actually work? Regarding subdomains?

nagyv: @timofurrer I was thinking about these too but wanted to focus on the problem (the feature development). Tests in https://github.com/gitlabhq/terraform-provider-gitlab/pull/1026 pass with these settings in GitPod.

timofurrer: Sure - I've created a follow up issue here:

* https://github.com/gitlabhq/terraform-provider-gitlab/issues/1042

timofurrer: Feel free to merge once the tests passed 🎉

Issue #1035, opened 2022-04-20
Realization Milestone (project) object


Moon1706:
### Feature Description

I wanna develop [Release](https://docs.gitlab.com/ee/api/releases/) resource and data source. For this, I should create [Milestone](https://docs.gitlab.com/ee/api/milestones.html) and [Release links](https://docs.gitlab.com/ee/api/releases/links.html).

### Do you want to implement this?

- [X] I would like to implement this myself 👷

Moon1706:
Hi @timofurrer  !
I need advice from your side. I'll wanna develop Release resource, but I've hit with an interesting problem. [In official API documentation](https://docs.gitlab.com/ee/api/releases/) a release contains Milestones, but [official Golang API](https://github.com/xanzy/go-gitlab/blob/master/releases.go) not. That's not a trouble, just a fact. But, but as you can see Milestone in GitLab API also contains field `issue_stats` and it's super strange value, because we can't get it from [Milestone API](https://docs.gitlab.com/ee/api/milestones.html) and we'll must calculate it. Also, we can't just use a list of [Milestone structure in Golang API](https://github.com/xanzy/go-gitlab/blob/master/milestones.go) and have to develop a new structure with this field. As a result, the more I think about in the more I'm sure that we should skip this field. So, what do you think about it?

PR #1044, opened 2022-04-28
Added gitlab_project_milestone as a resource and data source 

Moon1706:
## Description

Add resource for creation and management Gitlab project milestones ([issue](https://github.com/gitlabhq/terraform-provider-gitlab/issues/1035)).

    - Create the resource
    - Update the attributes
    - Import the resource


Moon1706:
@timofurrer
Review please)

internal/provider/util.go
```diff
@@ -184,11 +184,29 @@ func parseTwoPartID(id string) (string, string, error) {
 	return parts[0], parts[1], nil
 }
 
+func parseTwoPartIDInt(id string) (string, int, error) {
```
timofurrer, line 187: I'm not sure about this util function here. 
Can you implement something like `resourceGitlabProjectMilestoneParseID` instead?

timofurrer: ... it's more consistent with the other resources. And I think eventually we may need some more generic concept to parse these ids (with whatever type combinations ...).

internal/provider/util.go
```diff
@@ -184,11 +184,29 @@ func parseTwoPartID(id string) (string, string, error) {
...
+
+	return project, intID, nil
+}
+
 // format the strings into an id `a:b`
 func buildTwoPartID(a, b *string) string {
 	return fmt.Sprintf("%s:%s", *a, *b)
 }
 
+func buildTwoPartIDInterface(a string, b interface{}) string {
```
timofurrer, line 206: Let's do the string formatting correctly in the individual resources.

Same reasons as stated for `parseTwoPartIDInt` ...

internal/provider/schema_gitlab_project_milestone.go
```diff
@@ -0,0 +1,121 @@
...
+			Description: "The ID of the milestone only in one project.",
+			Type:        schema.TypeInt,
+			Computed:    true,
+		},
+		"milestone_id": {
+			Description: "The global ID of the project’s milestone.",
+			Type:        schema.TypeInt,
+			Computed:    true,
+		},
+		"project_id": {
```
timofurrer, line 26: I think that this will eventually lead to problems, because it's an actual field from the GitLab API response which **always** is an integer id of the associated project. Thus, if someone passes in a path with namespace (e.g. in the data source) then there is a config - state conflict. 

I suggest to add an additional `project` argument with the same schema as this one here and use the `project_id` as a `Computed: true` field, only - with the data returned from the GitLab API ... WDYT?

internal/provider/schema_gitlab_project_milestone.go
```diff
@@ -0,0 +1,121 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/xanzy/go-gitlab"
+)
+
+func gitlabProjectMilestoneGetSchema() map[string]*schema.Schema {
+	validMilestoneStates := []string{"active", "closed"}
+
+	return map[string]*schema.Schema{
```
timofurrer, line 15: Let's put the arguments on top of the schema definition here for readability ...

internal/provider/schema_gitlab_project_milestone.go
```diff
@@ -0,0 +1,121 @@
...
+		"iid": {
+			Description: "The ID of the milestone only in one project.",
```
timofurrer, line 17:
```suggestion
			Description: "The ID of the project's milestone.",
```

(same as in the API documentation)

internal/provider/schema_gitlab_project_milestone.go
```diff
@@ -0,0 +1,121 @@
...
+			Type:        schema.TypeInt,
+			Computed:    true,
+		},
+		"milestone_id": {
+			Description: "The global ID of the project’s milestone.",
```
timofurrer, line 22:
```suggestion
			Description: "The instance-wide ID of the project’s milestone.",
```

internal/provider/resource_gitlab_project_milestone_test.go
```diff
@@ -0,0 +1,178 @@
...
+						Title:       fmt.Sprintf("test-%d", rInt2),
+						ProjectID:   project.ID,
+						Description: fmt.Sprintf("test-%d", rInt2),
+						StartDate:   gitlab.ISOTime(time.Date(2022, time.April, 10, 0, 0, 0, 0, time.UTC)),
+						DueDate:     gitlab.ISOTime(time.Date(2022, time.April, 15, 0, 0, 0, 0, time.UTC)),
+						State:       "closed",
+						Expired:     true,
+					}),
+				),
+			},
```
timofurrer, line 65: Add another import verification step here.

internal/provider/resource_gitlab_project_milestone_test.go
```diff
@@ -0,0 +1,178 @@
...
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: providerFactories,
+		CheckDestroy:      testAccCheckGitlabProjectMilestoneDestroy,
+		Steps: []resource.TestStep{
+			{
+				// create Milestone with required values only
+				Config: testAccGitlabProjectMilestoneConfigRequiredOnly(project.ID, rInt1, ""),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGitlabProjectMilestoneExists("this", &milestone),
+					testAccCheckGitlabProjectMilestoneAttributes("this", &milestone, &testAccGitlabProjectMilestoneExpectedAttributes{
```
timofurrer, line 33: Can you use the resource helpers from `resource` ? Like `resource.TestCheckResourceAttr` etc?

Also note the testing guidelines we are working on here: https://github.com/gitlabhq/terraform-provider-gitlab/issues/1008

e.g. non-computed field don't need to be asserted explicitly in case of a subsequent import verification.

internal/provider/resource_gitlab_project_milestone.go
```diff
@@ -0,0 +1,168 @@
...
+)
+
+var milestoneStateToStateEvent = map[string]string{
+	"active": "activate",
+	"closed": "close",
+}
+
+var _ = registerResource("gitlab_project_milestone", func() *schema.Resource {
+	return &schema.Resource{
+		Description: `The ` + "`gitlab_project_milestone`" + ` resource allows to manage the lifecycle of a milestone (project).
```
timofurrer, line 19:
```suggestion
		Description: `The ` + "`gitlab_project_milestone`" + ` resource allows to manage the lifecycle of a project milestone.
```

internal/provider/resource_gitlab_project_milestone.go
```diff
@@ -0,0 +1,168 @@
...
+		CreateContext: resourceGitlabProjectMilestoneCreate,
+		ReadContext:   resourceGitlabProjectMilestoneRead,
+		UpdateContext: resourceGitlabProjectMilestoneUpdate,
+		DeleteContext: resourceGitlabProjectMilestoneDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		Schema: constructSchema(
+			gitlabProjectMilestoneGetSchema(),
+		),
```
timofurrer, line 32:
```suggestion
		Schema: gitlabProjectMilestoneGetSchema(),
```

internal/provider/data_source_gitlab_project_milestones_test.go
```diff
@@ -0,0 +1,40 @@
...
+					resource.TestCheckResourceAttr("data.gitlab_project_milestones.this", "milestones.0.description", testMilestones[1].Description),
+					resource.TestCheckResourceAttr("data.gitlab_project_milestones.this", "milestones.1.title", testMilestones[0].Title),
+					resource.TestCheckResourceAttr("data.gitlab_project_milestones.this", "milestones.1.description", testMilestones[0].Description),
+				),
+			},
+		},
+	})
+}
+
+func testAccDataGitlabProjectMilestonesConfig(projectID int) string {
```
timofurrer, line 34: Move this inline to adhere to the upcoming testing guardrails: https://github.com/gitlabhq/terraform-provider-gitlab/issues/1008

internal/provider/data_source_gitlab_project_milestones.go
```diff
@@ -0,0 +1,129 @@
...
+	validMilestoneStates := []string{"active", "closed"}
+
+	return &schema.Resource{
+		Description: `The ` + "`gitlab_project_milestones`" + ` data source allows get details of a project milestones.
+
+**Upstream API**: [GitLab REST API docs](https://docs.gitlab.com/ee/api/milestones.html)`,
+
+		ReadContext: dataSourceGitlabProjectMilestonesRead,
+		Schema: map[string]*schema.Schema{
+			"project_id": {
```
timofurrer, line 24: Same reasoning as for the other schemas:

```suggestion
			"project": {
```

internal/provider/data_source_gitlab_project_milestones.go
```diff
@@ -0,0 +1,129 @@
...
+		if err != nil {
+			return diag.FromErr(err)
+		}
+
+		milestones = append(milestones, paginatedMilestones...)
+		options.Page = resp.NextPage
+	}
+
+	log.Printf("[DEBUG] get gitlab milestones from project: %s", project)
+	d.SetId(buildTwoPartIDInterface(project, optionsHash))
```
timofurrer, line 116:
```suggestion
	d.SetId(fmt.Sprintf("%s:%d", project, optionsHash))
```

even though I think that `project` isn't even required and `optionsHash` is enough ...

internal/provider/data_source_gitlab_project_milestone_test.go
```diff
@@ -0,0 +1,39 @@
...
+					resource.TestCheckResourceAttr("data.gitlab_project_milestone.this", "milestone_id", fmt.Sprintf("%v", testMilestone.ID)),
+					resource.TestCheckResourceAttr("data.gitlab_project_milestone.this", "title", testMilestone.Title),
+					resource.TestCheckResourceAttr("data.gitlab_project_milestone.this", "description", testMilestone.Description),
+				),
+			},
+		},
+	})
+}
+
+func testAccDataGitlabProjectMilestoneConfig(projectID int, milestoneID int) string {
```
timofurrer, line 32: Move this inline to adhere to the upcoming testing guardrails: https://github.com/gitlabhq/terraform-provider-gitlab/issues/1008

timofurrer (changes requested):
Thanks for your work @Moon1706 ! Already looks pretty good 🎉 

I've left some comments - just re-request a review when you are ready 🏓

internal/provider/util.go
Moon1706: Okey)

internal/provider/schema_gitlab_project_milestone.go
Moon1706: Yeap

internal/provider/schema_gitlab_project_milestone.go
Moon1706: @timofurrer 
Mmm, that's a tough question. This will not happen because in Milestone Golang cline the mechanism of work differs from Issues. For example, in `gitlabProjectIssueToStateMap` function we [add project value](https://github.com/gitlabhq/terraform-provider-gitlab/blob/main/internal/provider/schema_gitlab_project_issue.go#L284) and this value [is written in `project`](https://github.com/gitlabhq/terraform-provider-gitlab/blob/main/internal/provider/schema_gitlab_project_issue.go#L286). In my case, this value [is got from response of Golang client](https://github.com/Moon1706/terraform-provider-gitlab/blob/main/internal/provider/schema_gitlab_project_milestone.go#L97) and it's **always** string with project ID even if I contain in options path with namespace.
Probably, we'll have only one problem - person put and expect a path with namespace, but get project ID in state. But I reckon it's okay. If you don't agree, I change it.

internal/provider/schema_gitlab_project_milestone.go
Moon1706: I don't pretty good understand what you mean. Could you clarify that?

internal/provider/resource_gitlab_project_milestone_test.go
Moon1706: @timofurrer 
So, if I rightly understand [this](https://github.com/gitlabhq/terraform-provider-gitlab/issues/1008) and [this](https://github.com/hashicorp/terraform-plugin-sdk/blob/5adf5f1c4bf0aee7bb31d51cd7f016e81a39e3e5/helper/resource/testing.go) we should use the function
`resource.TestCheckResourceAttrSet("gitlab_foo.foo", "bar"),` for fields `Computed:    true`, aren't we?

internal/provider/data_source_gitlab_project_milestones.go
Moon1706: Connect with previous discussion.

internal/provider/schema_gitlab_project_milestone.go
timofurrer: Oh sorry, basically all non-computed arguments starting with the required ones, then optional, then computed :)

internal/provider/resource_gitlab_project_milestone_test.go
timofurrer: Yes, or `resource.TestCheckResourceAttrWith` if you want to check for the value which may be dynamic, see:

https://github.com/gitlabhq/terraform-provider-gitlab/blob/5c1b5250b38bf4e207ee4381b9c6e27e8e577f52/internal/provider/resource_gitlab_project_issue_test.go#L43

timofurrer: ... in case it's known before hand you could use `resource.TestCheckResourceAttr` instead.

internal/provider/schema_gitlab_project_milestone.go
timofurrer: To demonstrate the problem can you change [this](https://github.com/gitlabhq/terraform-provider-gitlab/pull/1044/files#diff-48bb913f0116bcb2402a3448e78d0167cde6401eb0c523d62b03d82349c4b93fR27) test config to:

```go
               Config: fmt.Sprintf(`
                resource "gitlab_project_milestone" "this" {
                  project_id = "%s"
                  title      = "test-%d"
                }`, project.PathWithNamespace, rInt1),
```

... which is a supported use case. 

When you run this you'll get this error:

```
=== RUN   TestAccGitlabProjectMilestone_basic
    resource_gitlab_project_milestone_test.go:19: Step 1/4 error: After applying this test step, the plan was not empty.
        stdout:
        
        
        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
        
          # gitlab_project_milestone.this must be replaced
        -/+ resource "gitlab_project_milestone" "this" {
              ~ created_at   = "2022-05-13T06:53:06Z" -> (known after apply)
              ~ expired      = false -> (known after apply)
              ~ id           = "root/acctest-5013295642796043147:5" -> (known after apply)
              ~ iid          = 1 -> (known after apply)
              ~ milestone_id = 5 -> (known after apply)
              ~ project_id   = "52" -> "root/acctest-5013295642796043147" # forces replacement
              ~ updated_at   = "2022-05-13T06:53:06Z" -> (known after apply)
              ~ web_url      = "http://fd87da6cdd76/root/acctest-5013295642796043147/-/milestones/1" -> (known after apply)
                # (2 unchanged attributes hidden)
            }
        
--- FAIL: TestAccGitlabProjectMilestone_basic (3.21s)
```

Note the line `~ project_id   = "52" -> "root/acctest-5013295642796043147" # forces replacement`. 

This is the problem I was referring to. What's going on here is that the API **always** returns an integer `id` no matter if you called the GitLab API using `/projects/52/milestones/1` or `/projects/root%2Facctest-5013295642796043147/milestones/1`. And then always writing this integer id into the `project_id` attribute will result in that state / plan mismatch.

I hope this helped, let me know :)

internal/provider/data_source_gitlab_project_milestone_test.go
```diff
@@ -0,0 +1,34 @@
+package provider
```
timofurrer, line 1:
```suggestion
//go:build acceptance
// +build acceptance

package provider
```

required after the rebase onto `main`, see #1045

internal/provider/data_source_gitlab_project_milestones_test.go
```diff
@@ -0,0 +1,35 @@
+package provider
```
timofurrer, line 1:
```suggestion
//go:build acceptance
// +build acceptance

package provider
```

required after the rebase onto `main`, see #1045

internal/provider/resource_gitlab_project_milestone_test.go
```diff
@@ -0,0 +1,96 @@
+package provider
```
timofurrer, line 1:
```suggestion
//go:build acceptance
// +build acceptance

package provider
```

required after the rebase onto `main`, see #1045

internal/provider/schema_gitlab_project_milestone.go
Moon1706: It's strange... Okey, in this case I agree with you and I've added `project` field.

internal/provider/schema_gitlab_project_milestone.go
```diff
@@ -0,0 +1,127 @@
...
+		stateMap["due_date"] = milestone.DueDate.String()
+	} else {
+		stateMap["due_date"] = nil
+	}
+	if milestone.StartDate != nil {
+		stateMap["start_date"] = milestone.StartDate.String()
+	} else {
+		stateMap["start_date"] = nil
+	}
+	stateMap["updated_at"] = milestone.UpdatedAt.Format(time.RFC3339)
```
timofurrer, line 116: I think that both the `milestones.UpdatedAt` and `milestone.CreatedAt` fields can technically be `nil` (because of their `*time.Time` type. 

However, probably only `milestone.UpdatedAt` ever will be, right?

internal/provider/schema_gitlab_project_milestone.go
```diff
@@ -0,0 +1,127 @@
...
+			ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice(validMilestoneStates, false)),
+		},
+		"created_at": {
+			Description: "The time of creation of the milestone. Date time string, ISO 8601 formatted, for example 2016-03-11T03:45:40Z.",
+			Type:        schema.TypeString,
+			Optional:    true,
+			Computed:    true,
+			// NOTE: since RFC3339 is pretty much a subset of ISO8601 and actually expected by GitLab,
+			//       we use it here to avoid having to parse the string ourselves.
+			ValidateDiagFunc: validation.ToDiagFunc(validation.IsRFC3339Time),
```
timofurrer, line 59: According to the API [docs](https://docs.gitlab.com/ee/api/milestones.html#edit-milestone), `created_at` cannot be set by the user.

```suggestion
			Computed:    true,
```

internal/provider/schema_gitlab_project_milestone.go
```diff
@@ -0,0 +1,127 @@
...
+		},
+		"updated_at": {
+			Description: "The last update time of the milestone. Date time string, ISO 8601 formatted, for example 2016-03-11T03:45:40Z.",
+			Type:        schema.TypeString,
+			Optional:    true,
+			Computed:    true,
+			// NOTE: since RFC3339 is pretty much a subset of ISO8601 and actually expected by GitLab,
+			//       we use it here to avoid having to parse the string ourselves.
+			ValidateDiagFunc: validation.ToDiagFunc(validation.IsRFC3339Time),
```
timofurrer, line 68: According to the API [docs](https://docs.gitlab.com/ee/api/milestones.html#edit-milestone), `updated_at` cannot be set by the user.

```suggestion
			Computed:    true,
```

internal/provider/resource_gitlab_project_milestone.go
```diff
@@ -0,0 +1,187 @@
...
+		return diag.FromErr(err)
+	}
+
+	log.Printf("[DEBUG] read gitlab milestone in project %s with ID %d", project, milestoneID)
+	milestone, resp, err := client.Milestones.GetMilestone(project, milestoneID, gitlab.WithContext(ctx))
+	if err != nil {
+		if is404(err) {
+			log.Printf("[WARN] recieved 404 for gitlab milestone ID %d in project %s, removing from state", milestoneID, project)
+			d.SetId("")
+			return diag.FromErr(err)
```
timofurrer, line 97: When reading a resource and it doesn't exist it's actually not an error per se, but we want to remove it from state instead. That's why we return `nil` instead of an error and just clear the id (`d.SetId("")`):

```suggestion
			return nil
```

internal/provider/resource_gitlab_project_milestone_test.go
```diff
@@ -0,0 +1,97 @@
...
+		ProviderFactories: providerFactories,
+		CheckDestroy:      testAccCheckGitlabProjectMilestoneDestroy,
+		Steps: []resource.TestStep{
+			{
+				// create Milestone with required values only
+				Config: fmt.Sprintf(`
+                resource "gitlab_project_milestone" "this" {
+                  project = "%v"
+                  title   = "test-%d"
+                }`, project.PathWithNamespace, rInt1),
```
timofurrer, line 31: NIT: for better readability:

```suggestion
				Config: fmt.Sprintf(`
				resource "gitlab_project_milestone" "this" {
                                   project = "%v"
                                   title   = "test-%d"
                                 }`, project.PathWithNamespace, rInt1),
```

internal/provider/resource_gitlab_project_milestone_test.go
```diff
@@ -0,0 +1,97 @@
...
+				// update some Milestone attributes
+				Config: fmt.Sprintf(`
+                resource "gitlab_project_milestone" "this" {
+                  project     = "%[1]d"
+                  title       = "test-%[2]d"
+                  description = "test-%[2]d"
+                  start_date  = "2022-04-10"
+                  due_date    = "2022-04-15"
+                  state       = "closed"
+                }`, project.ID, rInt2),
```
timofurrer, line 57: NIT: for better readability:

```suggestion
				Config: fmt.Sprintf(`
                                 resource "gitlab_project_milestone" "this" {
                                   project     = "%[1]d"
                                   title       = "test-%[2]d"
                                   description = "test-%[2]d"
                                   start_date  = "2022-04-10"
                                   due_date    = "2022-04-15"
                                   state       = "closed"
                                 }`, project.ID, rInt2),
```

timofurrer (changes requested):
Thanks for the update @Moon1706 :tada:

I've left a few more comments which required changes and some nitpicks.

Back to you 🏓

internal/provider/resource_gitlab_project_milestone_test.go
Moon1706: Hm,  I thought `make reviewable` replace this. Okay...

internal/provider/resource_gitlab_project_milestone.go
Moon1706: I'm agree with you. Stange that I added this error)

internal/provider/schema_gitlab_project_milestone.go
Moon1706: Mmm, I doubt. But yes, as "backup strategy" I can add if condition.

timofurrer: Yes, you are right. Strangely the API returns the same `updated_at` as the `created_at` right after creation 🤷 

So, either way is fine for me ;) 

Sorry for the confusion ✌️

timofurrer (approved):
Thank you @Moon1706 once again for a nice contribution 🤝 

Let's ship this 🚢

PR #1045, opened 2022-04-28
Remove test boilerplate by adding build tag

armsnyder:
## Description

Relates to #1008

I'm interested in feedback on this. Is this a good change, if contributors now have to be sure to add the `// +build acceptance` header to new tests, or should we instead keep our test-skipping logic as it is?

The goal of this PR is to remove the `testAccCheck(t)` and `PreCheck` boilerplate code from all tests.

I was able to do this by introducing an "acceptance" [build tag](https://mickey.dev/posts/go-build-tags-testing/).

The 4 commits are independent, so you can review them separately.

1. Run acceptance tests with acceptance build tag
    - I added the `// +build acceptance` build tag on all acceptance test files, but NOT unit test files. This allows us to run acceptance tests by specifying `go test -tags acceptance`, and keep the default `go test` command only running unit tests, without the need to check the value of `TF_ACC` in every test.

2. Run tfproviderlint as a golangci-lint plugin
    - I changed the way that we run tfproviderlint so that instead of running the binary, we run it as a plugin of golangci-lint. This was necessary because tfproviderlint does not allow you to specify build tags, so it was not linting acceptance tests anymore. Whereas, golangci-lint does let you specify build tags.

3. Allow linters to lint acceptance tests
    - Runs linters with `--build-tags acceptance` so that acceptance tests are included in linting.

4. Remove skip-test checks and rely on build tags instead
    - Deleted all uses of `testAccCheck` and `PreCheck`. Now that acceptance tests are enabled using an `acceptance` build tag, we do not need to do these runtime checks.

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,8 +1,8 @@
 {
-  "go.testEnvVars": {
-    "TF_ACC": "1",
-    "GITLAB_TOKEN": "ACCTEST1234567890123",
-    "GITLAB_BASE_URL": "http://127.0.0.1:8080"
-  },
-  "go.testFlags": ["-count=1", "-v", "-timeout=30m"]
+    "go.testEnvVars": {
+        "TF_ACC": "1",
+        "GITLAB_TOKEN": "ACCTEST1234567890123",
+        "GITLAB_BASE_URL": "http://127.0.0.1:8080"
+    },
+    "go.testFlags": ["-count=1", "-v", "-tags=acceptance", "-tags=acceptance"]
 }
--- a/.github/workflows/pr-lint.yml
+++ b/.github/workflows/pr-lint.yml
@@ -28,7 +28,6 @@ jobs:
       matrix:
         target:
           - lint-golangci
-          - lint-tfprovider
           - lint-examples-tf
           - lint-examples-sh
           - lint-generated
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -40,7 +40,6 @@ jobs:
       matrix:
         target:
           - lint-golangci
-          - lint-tfprovider
           - lint-examples-tf
           - lint-examples-sh
           - lint-generated
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,6 +4,11 @@ run:
   timeout: 3m
 
 linters-settings:
+  custom:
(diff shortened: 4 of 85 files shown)


.vscode/settings.json
```diff
@@ -4,5 +4,5 @@
         "GITLAB_TOKEN": "ACCTEST1234567890123",
         "GITLAB_BASE_URL": "http://127.0.0.1:8080"
     },
-    "go.testFlags": ["-count=1", "-v"]
```
PatrickRice-KSC, line 7: Do you think we want to add this to the settings.json? It seems like that may be confusing if someone clicks play on a unit test to run it, and it passes because it's skipped?

armsnyder: I didn't put build tags on the unit test files, so they still run when you use `acceptance`.

PatrickRice-KSC: Ah, nice. I assumed that it would only run files explicitly tagged.

timofurrer (approved):
@armsnyder It's an approve from me. 

I think it's the best and easiest option we have right now. The linting plugin option also seems fine to me.

The only thing which isn't perfect is that some "acceptance test" files also contain unit tests, which with this change are then only run for acceptance test builds, e.g. :

https://github.com/gitlabhq/terraform-provider-gitlab/blob/39bf02ff493539153950d6bd53226dca1a85aec8/internal/provider/resource_gitlab_deploy_token_test.go#L246

Nevertheless, I think they can easily split out later if it bothers ... I think 99% of testing focus is on acceptance tests anyways ...

timofurrer:
@armsnyder do you have time to get a little further with this one? I think it makes sense to merge asap so that other new resources can already make use of it ...

timofurrer (approved):
🚢

timofurrer:
I've just spotted that the acceptance tests are not properly executed: https://github.com/gitlabhq/terraform-provider-gitlab/runs/6419139956?check_suite_focus=true#step:6:1

Also I've added a commit to remove the last usage of `testAccPreCheck` :)

timofurrer:
I've added the `--tags` option to actually run the acceptance tests in `make testacc`.

timofurrer (approved):

PR #1049, opened 2022-04-30
Add gitlab_runner Resource

### LINKED ISSUE #49: Gitlab runner feature request
One of the obvious features to terraform is Gitlab runner (https://docs.gitlab.com/runner/)

What problem I would it solved?

Currently, I have the CI infrastructure terraformed on AWS and is very easy to register new runner via shell cli and remote-exec but impossible to deregister the runner.

```
 provisioner "remote-exec" {
    connection {
      type = "ssh"
      user = "ec2-user"
      private_key = "${file("PRIVATE_KEY")}"
    }

    inline = [
      "sudo gitlab-runner register -n -u https://gitlab.YOUR_DNS.com -r SECRET)_TOKEN -name your--manually-created-runner-waiting-to-be-terraformed --tag-list ci,test --executor shell",
    ]
  }
```

It has quite a disadvantage that the runners accumulates when I redeploy the infrastructure because I have no means how to delete them.

PatrickRice-KSC:
## Description

This PR adds support for registering runners using a `gitlab_runner` resource. 

Closes #49 

    - Create the resource
    - Update the attributes
    - Import the resource


PatrickRice-KSC:
unfortunately, there are a couple new `lintignore` issues since there are boolean values here which require the use of `GetOkExists`, which is a deprecated method ☹️

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/xanzy/go-gitlab"
+)
+
+var _ = registerResource("gitlab_runner", func() *schema.Resource {
+	return &schema.Resource{
+		Description: `The ` + "`gitlab_runner`" + ` resource allows registering a runner, either at an instance level
+or at a group level. The runner will be registered at a group level if the token used is from a group, or at an 
+instance level if the token used is for the instance.
```
timofurrer, line 19:
```suggestion
		Description: `The ` + "`gitlab_runner`" + ` resource allows to manage the lifecycle of a runner.
		
A runner can either be registered at an instance level or group level. 
The runner will be registered at a group level if the token used is from a group, or at an instance level if the token used is for the instance.
```

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+
+**Upstream API**: [GitLab REST API docs](https://docs.gitlab.com/ee/api/runners.html#register-a-new-runner)`,
+
+		// Since this resource updates an in-place resource, the update method is the same as the create method
```
timofurrer, line 23: wrong comment:

```suggestion
```

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"token": {
+				Description: `The registration token used to register the runner`,
+				Type:        schema.TypeString,
+				ForceNew:    true,
+				Required:    true,
+				Sensitive:   true,
+			},
+			// Even though the output variable is just called 'token', we need to differentiate between the regitration
```
timofurrer, line 40:
```suggestion
			// Even though the output variable is just called 'token', we need to differentiate between the registration
```

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+		UpdateContext: resourceGitLabRunnerUpdate,
+		ReadContext:   resourceGitLabRunnerRead,
+		DeleteContext: resourceGitLabRunnerDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"token": {
+				Description: `The registration token used to register the runner`,
```
timofurrer, line 34:
```suggestion
				Description: `The registration token used to register the runner.`,
```

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+			// and authentication token in terraform.
+			"authentication_token": {
+				Description: `The authentication token used for building a config.toml file. This value is not present when imported.`,
+				Type:        schema.TypeString,
+				Computed:    true,
+				Sensitive:   true,
+			},
+			"status": {
+				Description: `The status of runners to show, one of: online and offline. active and paused are also possible values
+				 which were deprecated in GitLab 14.8 and will be removed in GitLab 16.0`,
```
timofurrer, line 50: This indentation yields to a strange formatting (see the md) ...

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+			},
+			// Even though the output variable is just called 'token', we need to differentiate between the regitration
+			// and authentication token in terraform.
+			"authentication_token": {
+				Description: `The authentication token used for building a config.toml file. This value is not present when imported.`,
+				Type:        schema.TypeString,
+				Computed:    true,
+				Sensitive:   true,
+			},
+			"status": {
```
timofurrer, line 48: NIT: Let's have the computed-only fields at the end of the schema definition for more readability :)

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+				Description: `The status of runners to show, one of: online and offline. active and paused are also possible values
+				 which were deprecated in GitLab 14.8 and will be removed in GitLab 16.0`,
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"description": {
+				Description: `The runner's description'`,
```
timofurrer, line 55:
```suggestion
				Description: `The runner's description.`,
```

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			"paused": {
+				Description: `Whether the runner should ignore new jobs`,
```
timofurrer, line 60:
```suggestion
				Description: `Whether the runner should ignore new jobs.`,
```

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     gitlab.Bool(false),
```
timofurrer, line 63: hmmm, what's the point of `gitlab.Bool` here? I don't think that it changes anything in behavior.
Also, we usually don't explicitly set the values to the defaults - thus, you can omit specifying the `Default` at all here ...

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+			},
+			"locked": {
+				Description: `Whether the runner should be locked for current project`,
```
timofurrer, line 66:
```suggestion
				Description: `Whether the runner should be locked for current project.`,
```

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     gitlab.Bool(false),
+			},
+			"run_untagged": {
+				Description: `Whether the runner should handle untagged jobs`,
```
timofurrer, line 72:
```suggestion
				Description: `Whether the runner should handle untagged jobs.`,
```

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     gitlab.Bool(true),
+			},
+			"tag_list": {
+				Description: `List of runner’s tags`,
```
timofurrer, line 78:
```suggestion
				Description: `List of runner’s tags.`,
```

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+				Type:        schema.TypeList,
+				Optional:    true,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+			},
+			"access_level": {
+				Description: fmt.Sprintf(`The access_level of the runner. Valid values are: %s`,
+					renderValueListForDocs(runnerAccessLevelAllowedValues)),
```
timofurrer, line 85:
```suggestion
				Description: fmt.Sprintf(`The access_level of the runner. Valid values are: %s`, renderValueListForDocs(runnerAccessLevelAllowedValues)),
```

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.StringInSlice(runnerAccessLevelAllowedValues, true),
+				Default:      "not_protected",
```
timofurrer, line 89: I prefer to not duplicate the API defaults in the provider if not really necessary.

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+			},
+			"maximum_timeout": {
+				Description: `Maximum timeout set when this runner handles the job`,
```
timofurrer, line 92:
```suggestion
				Description: `Maximum timeout set when this runner handles the job.`,
```

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+				Type:        schema.TypeInt,
+				Optional:    true,
+			},
+			// While Maintenance Note is available during "create", it is not available during "update", so
+			// excluding it here for now.
+		},
+	}
+})
+
+func resourceGitLabRunnerDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
timofurrer, line 102: Any reason for the delete first?

NIT: We usually had it in `CRUD` order :)

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+	client := meta.(*gitlab.Client)
+	runnerId, err := strconv.Atoi(d.Id())
```
timofurrer, line 104: NIT:

```suggestion
	runnerID, err := strconv.Atoi(d.Id())
```

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	return nil
+}
+
+func resourceGitLabRunnerRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*gitlab.Client)
+	runnerId, err := strconv.Atoi(d.Id())
```
timofurrer, line 120:
```suggestion
	runnerID, err := strconv.Atoi(d.Id())
```

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+func resourceGitLabRunnerUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*gitlab.Client)
+	runnerId := d.Id()
+
+	options := &gitlab.UpdateRunnerDetailsOptions{}
+	if v, ok := d.GetOk("description"); ok {
+		options.Description = gitlab.String(v.(string))
+	}
+
+	//GetOK skips the block if the value is "false", so need to use GetOkExists even though it's deprecated.
```
timofurrer, line 155: NIT:

```suggestion
	// GetOK skips the block if the value is "false", so need to use GetOkExists even though it's deprecated.
```

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+	if v, ok := d.GetOk("access_level"); ok {
+		options.AccessLevel = gitlab.String(v.(string))
+	}
+
+	if v, ok := d.GetOk("maximum_timeout"); ok {
+		options.MaximumTimeout = gitlab.Int(v.(int))
+	}
+
+	// Explicitly not printing the registration token here, even though it may make debugging a bit trickier, since it's a
+	// secret.
```
timofurrer, line 239:
```suggestion
	// Explicitly not printing the registration token here, even though it may make debugging a bit trickier, since it's a secret.
```

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+	log.Printf("[DEBUG] Update GitLab Runner using registration token in configuration")
+	runner, _, err := client.Runners.RegisterNewRunner(options, gitlab.WithContext(ctx))
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(strconv.Itoa(runner.ID))
+
+	//The authentication_token will ONLY exist during creation, and will not return during "read", so we need to set it here.
```
timofurrer, line 248: NIT: in general I think it's nice to have a space between `//` and the first character .. just makes it more readable ... but major personal opinion, so 🤷  😅 

```suggestion
	// The authentication_token will ONLY exist during creation, and will not return during "read", so we need to set it here.
```

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+		CreateContext: resourceGitLabRunnerCreate,
+		UpdateContext: resourceGitLabRunnerUpdate,
+		ReadContext:   resourceGitLabRunnerRead,
+		DeleteContext: resourceGitLabRunnerDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"token": {
```
timofurrer, line 33: Should we call it `registration_token` explicitly? 

@armsnyder WDYT?

internal/provider/resource_gitlab_runner.go
```diff
@@ -0,0 +1,257 @@
...
+				Description: `The registration token used to register the runner`,
+				Type:        schema.TypeString,
+				ForceNew:    true,
+				Required:    true,
+				Sensitive:   true,
+			},
+			// Even though the output variable is just called 'token', we need to differentiate between the regitration
+			// and authentication token in terraform.
+			"authentication_token": {
```
timofurrer, line 42: There is also a `token_expires_at` field in the create response ... should we expose it, too?

internal/provider/resource_gitlab_runner_test.go
```diff
@@ -0,0 +1,121 @@
...
+	//The runner token is not populated on the return from the group create, so re-retrieve it to get the token.
+	group, _, err := testGitlabClient.Groups.GetGroup(group.ID, &gitlab.GetGroupOptions{})
+	if err != nil {
+		t.Fail()
+	}
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() {},
+		ProviderFactories: providerFactories,
+		CheckDestroy:      testAccCheckManagedLicenseDestroy,
```
timofurrer, line 24: The reference here is wrong ...

internal/provider/resource_gitlab_runner_test.go
```diff
@@ -0,0 +1,121 @@
...
+				}
+				`, group.RunnersToken),
+			},
+			{
+				ResourceName:      "gitlab_runner.this",
+				ImportState:       true,
+				ImportStateVerify: true,
+				//These need to be ignored since they don't come back in the "get" command
+				ImportStateVerifyIgnore: []string{"authentication_token", "token"},
+			},
```
timofurrer, line 55: Should we also test instance level runners?

internal/provider/resource_gitlab_runner_test.go
```diff
@@ -0,0 +1,121 @@
...
+					description = "Lorem Ipsum Dolor Sit Amet"
+				}
+				`, group.RunnersToken),
+			},
+			{
+				ResourceName:      "gitlab_runner.this",
+				ImportState:       true,
+				ImportStateVerify: true,
+				//These need to be ignored since they don't come back in the "get" command
+				ImportStateVerifyIgnore: []string{"authentication_token", "token"},
```
timofurrer, line 54: Not 100% sure, but since the `token` is part of the config you probably don't have to ignore it. Could you verify please?

timofurrer (changes requested):
@PatrickRice-KSC looks pretty solid already, thanks! 

There are a few copy & paste mistakes, nitpicks / styling stuff and other minor things.

Regarding the `GetOkExists()` - that's not a problem at all - IMHO it's still the best option we have with the current SDK and there will be a replacement before they delete it, thus all good.

Back to you 🏓

examples/resources/gitlab_runner/resource.tf
```diff
@@ -0,0 +1,21 @@
+# Basic GitLab Runner
```
timofurrer, line 1: I know that we don't do this a lot (we gotta start somewhere 😛 )  ...

Thus, what do you think about providing a more extensive example which would someone incorporate the `authentication_token` into a `config.toml` template? (I assume that would be an example workflow with this resource, right?)

internal/provider/resource_gitlab_runner_test.go
```diff
@@ -0,0 +1,121 @@
...
+		ProviderFactories: providerFactories,
+		CheckDestroy:      testAccCheckManagedLicenseDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: fmt.Sprintf(`
+				resource "gitlab_runner" "this" {
+					token = "%s"
+					description = "Lorem Ipsum"
+				}
+				`, group.RunnersToken),
```
timofurrer, line 32: Let's add a check if the `authentication_token` has been set using `resource.TestCheckResourceAttrSet()`. At the moment it's not verified, because it's not part of the import ...

internal/provider/resource_gitlab_runner_test.go
```diff
@@ -0,0 +1,121 @@
...
+	//The runner token is not populated on the return from the group create, so re-retrieve it to get the token.
+	group, _, err := testGitlabClient.Groups.GetGroup(group.ID, &gitlab.GetGroupOptions{})
+	if err != nil {
+		t.Fail()
+	}
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() {},
+		ProviderFactories: providerFactories,
+		CheckDestroy:      testAccCheckManagedLicenseDestroy,
```
timofurrer, line 73: also a wrong reference - you have to implement that check method in this test file yourself specific to the `gitlab_runner` resource :)

internal/provider/resource_gitlab_runner.go
PatrickRice-KSC: I debated this myself. I decided that I'd try to stick as closely to the API as I could, which is why I changed only the second one to authentication token. That having been said, changing it would be clearer to someone who is new to the provider which token it's expecting. I could absolutely go either way!

internal/provider/resource_gitlab_runner.go
PatrickRice-KSC: @timofurrer  - Good catch, I missed that. I'm not sure what someone would do with it, but no harm in including it!

PatrickRice-KSC: Hm..., it's not supported in the upstream go-gitlab. I'd say let's merge the PR without it and then we can add it later if someone has a use-case. Work for you?

internal/provider/resource_gitlab_runner_test.go
PatrickRice-KSC: @timofurrer  - I did verify, and it is required to ignore it. I suspect because it doesn't come back in the "Read" API (sensitive value below is obviously transient, so it's not a problem to post it): 

image

examples/resources/gitlab_runner/resource.tf
PatrickRice-KSC: Sure, I'll see what I can do.

internal/provider/resource_gitlab_runner_test.go
PatrickRice-KSC: Yeah... like you said, silly copy/paste mistake 😄

internal/provider/resource_gitlab_runner_test.go
PatrickRice-KSC: @timofurrer  - Good call! Added a test 😄

internal/provider/resource_gitlab_runner.go
PatrickRice-KSC: @timofurrer - true, but the markdown still renders correctly. Do you think we want to fix it? Since the documentation is generated, I'd personally favor the readability of the code in the func as opposed to the indentation of the markdown file:

image

internal/provider/resource_gitlab_runner.go
PatrickRice-KSC: @timofurrer  - Without the defaults set in the schema, the import step of the tests fail because it marks those values as "changed" (from "" -> "false")

internal/provider/resource_gitlab_runner.go
PatrickRice-KSC: Same comment as above about the import step of the test. Is there a better way to handle this?

internal/provider/resource_gitlab_runner.go
PatrickRice-KSC: No reason, just the order I created them in 😄 . Updated!

internal/provider/resource_gitlab_runner_test.go
PatrickRice-KSC: So this test is... pretty weird. I added a pretty lengthy comment at the top of it, but there seems to be a built-in delay in how often we can rotate the instance-level token. Since the CI should run against a fresh version of gitlab, it _shouldn't_ be a problem in the CI, but we'll see...

PatrickRice-KSC: OK, the test is failing in the CI/CD. I'm kinda feeling like it's flakey enough that we should remove the test until we have a better way of setting the instance runner token. Thoughts?

internal/provider/resource_gitlab_runner.go
timofurrer: Okay, but the alignment to me is still strange in the code with the one space indentation.
I'd rather align on the `The status ...` then ..

internal/provider/resource_gitlab_runner.go
timofurrer: Yes, usually it's best to stick to the API wording.

In this case I just think that the `token` term is kinda overloaded in this case: GitLab Token, Registration Token, Authentication Token ... So I thought why not explicitly name them.

But I'm also fine with just `token` :)

internal/provider/resource_gitlab_runner_test.go
timofurrer: interesting! Thanks for the verification 🎉

internal/provider/resource_gitlab_runner.go
timofurrer: Yes, sounds good. 
In the meantime I've created a PR in go-gitlab (let's not wait for it to be released. We can iterate :))

* https://github.com/xanzy/go-gitlab/pull/1459

internal/provider/resource_gitlab_runner.go
timofurrer: I've never seen this happening before 🤔 ,but okay :) The `gitlab.Bool` isn't necessary though, right?

timofurrer: aah just noticed, it's because it's not being marked as `Computed` !

internal/provider/resource_gitlab_runner.go
timofurrer: Yes, as the other attribute: mark it as `Computed` :)

internal/provider/resource_gitlab_runner_test.go
timofurrer: hmmm, so what's the problem here in the CI? Do you have the test somewhere?

examples/resources/gitlab_runner/resource.tf
```diff
@@ -0,0 +1,44 @@
...
+}
+
+# Generate a `config.toml` file that you can use to create a runner
+# This is the typical workflow for this resource, using it to create an authentication_token which can then be used
+# to generate the `config.toml` file to prevent re-registering the runner every time new hardware is created.
+resource "gitlab_runner" "my_runner" {
+  token = "12345"
+}
+
+# This creates a local "shell" runner, but can be changed to generate whatever is needed.
```
willianpaixao, line 30: Correct if I'm wrong but this won't really create a functional runner, unless you place that file in a cloud instance (for example).

I have a actual runner using the [official GitLab runner Helm chart](https://gitlab.com/gitlab-org/charts/gitlab-runner). Would you like me to add as an example?

examples/resources/gitlab_runner/resource.tf
```diff
@@ -0,0 +1,44 @@
+# Basic GitLab Runner
+resource "gitlab_runner" "this" {
+  token = "12345"
```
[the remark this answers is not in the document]
timofurrer: Yes, makes sense to add an additional example for these use cases I guess :)

examples/resources/gitlab_runner/resource.tf
timofurrer: No, I'd just registers one to obtain the auth token - nothing more :) 

I guess feel free to provider the example and see how it can fit the docs ...

internal/provider/resource_gitlab_runner.go
PatrickRice-KSC: Oh, sure. That's easy enough to fix!

internal/provider/resource_gitlab_runner.go
PatrickRice-KSC: Yknow, I didn't consider that `token` was in the provider block too. I'm going to change this to registration token just so we don't confuse new users.

internal/provider/resource_gitlab_runner_test.go
PatrickRice-KSC: @timofurrer  - You can see the test here: https://github.com/gitlabhq/terraform-provider-gitlab/pull/1049/files#diff-3bac8c8a2d8b8ebe1b74562e5f710b5fbc9d861fbfe8ac15d00d16721481d383R67

What's failing (assuming it's the same 403 I got when I was running it locally) requires a bit of explanation: There isn't a way to retrieve the instance-level runner token via the API that I can find, which means that instead of retrieving what's there, I call a _reset_ via the API. The reset API returns the _new_ registration token, however it's not immediately available for use. If you turn around and call the API using the new registration token immediately, you receive a 403 (which is what the CI/CD is receiving). However, what makes it janky is that the amount of time necessary to wait before the token can be used seems to be variable - I've seen everything from _instantly_ to _almost a minute_ in local testing. The time it takes for it to be usable seems to depend on how often you're calling the reset API, so I was hoping that since we only called it once during acceptance testing that it would work properly. That doesn't seem to be the case :-/

I don't really want to add a really long wait to the CI/CD to account for the case... I could add a new client-based call at the top of the test that calls "resource.Retry" until we get a non-403, but I feel like we still have a pseudo-race condition here.

examples/resources/gitlab_runner/resource.tf
PatrickRice-KSC: I wasn't actually trying to create a fully functioning runner in the example, honestly, because the terraform to do so in a secure fashion is... large. I was trying to provide juuuuust enough that someone could see "oh, this is how this resource is meant to be used".

Using your helm chart example, to create a functioning runner we'd have to pull in tf for creating a kubernetes cluster too, right? It feels like that's out of scope of the documentation personally.

examples/resources/gitlab_runner/resource.tf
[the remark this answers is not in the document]
PatrickRice-KSC: Good call, I'll update the examples to use those instead!

examples/resources/gitlab_runner/resource.tf
willianpaixao: > It feels like that's out of scope of the documentation personally

I agree it's out of scope, but in another hand I think we need to provide a more realistic example.
Would it look stupid to reference the chart but skip the Kubernetes cluster creation? Looks like a middle term.

examples/resources/gitlab_runner/resource.tf
PatrickRice-KSC: I did add an example that starts generating the config.toml, but with the way we have terraform-doc set up right now, the only thing we can "inject" is the example markdown. I could potentially create a new runner.md.tmpl file, and inject a new "Comprehensive Example" section, because it does feel like this is outside the scope of the basic example. I have to learn a bit more about the terraform-doc templating engine though to do that. I'd suggest that for the comprehensive example, I create a separate issue/PR for that 🙂

internal/provider/resource_gitlab_runner_test.go
timofurrer: Okay, yeah - I don't want to have such a delay in there either ...

Let's leave it out for now then. Do you mind adding this additional case to: https://gitlab.com/gitlab-org/gitlab/-/issues/352299

timofurrer: hmmm, is there a way to inject an instance token manually like we do with the PAT we use to authenticate to the instance?

PatrickRice-KSC: That's an interesting question - let me do some research. Since I use SaaS, I don't have a ton of experience with the ruby console commands, but I'll dig and see. What I'll do is remove the test for now, and we can re-add it with a PR once we can figure that out.

As to the gitlab issue, that seems to be related specifically to groups/subgroups (and is assigned as such) whereas this is related to application settings proper: @nagyv  - Do you think this should go in there, or should I create a new issue and let it get applied to a different gitlab team?

examples/resources/gitlab_runner/resource.tf
PatrickRice-KSC: @willianpaixao  - would you be willing to provide some TF that deploys that so I can include it in the PR? I think that makes sense, but I've not installed the runner via a helm chart (yet 😄 ), so I'd hate to include a bad example. It looks like the helm chart terraform provider accepts a kube config, so we can certainly leave setting up a k8s cluster as an 'exercise for the reader'.

internal/provider/resource_gitlab_runner.go
PatrickRice-KSC: Updated in last commit, just to be explicit 😄

internal/provider/resource_gitlab_runner.go
PatrickRice-KSC: Updated in last commit, just to be explicit 😄

internal/provider/resource_gitlab_runner_test.go
```diff
@@ -0,0 +1,150 @@
...
+	for _, rs := range state.RootModule().Resources {
+		if rs.Type != "gitlab_runner" {
+			continue
+		}
+
+		id, _ := strconv.Atoi(rs.Primary.ID)
+
+		runner, _, err := testGitlabClient.Runners.GetRunnerDetails(id)
+		if err == nil {
+			if runner != nil && runner.ID == id {
```
timofurrer, line 139: I know a lot of these destroy check functions have that check and I don't understand why, thus, I usually remove / don't copy them:

```suggestion
			if runner != nil {
```

examples/resources/gitlab_runner/resource.tf
```diff
@@ -0,0 +1,56 @@
...
+resource "local_file" "config" {
+  filename = "${path.module}/config.toml"
+  content  = <<CONTENT
+  concurrent = 1
+
+  [[runners]]
+    name = "Hello Terraform"
+    url = "https://example.gitlab.com/"
+    token = "${gitlab_runner.my_runner.authentication_token}"
+    executor="shell"
```
timofurrer, line 52:
```suggestion
    executor = "shell"
```

internal/provider/resource_gitlab_runner_test.go
PatrickRice-KSC: Works for me 😄  - I'm going to squash commits again here in a bit, but going to try and see if I can get a helm chart example and doing some research on the instance runner token set first.

internal/provider/resource_gitlab_runner_test.go
PatrickRice-KSC: So I _may_ have found a way to make this work. Digging into the code, it looks like this is likely a caching problem (which makes sense with the way the failures exhibit): https://gitlab.com/gitlab-org/gitlab/-/blob/master/ee/app/services/ee/ci/runners/reset_registration_token_service.rb#L22

However, digging into the specs that GitLab itself uses to test the API, there is a hint there. GitLab's tests stub in an environment variable that, based on its naming, would disable that in-memory caching: https://gitlab.com/gitlab-org/gitlab/-/blob/master/spec/requests/api/ci/runners_reset_registration_token_spec.rb#L77

I'm attempting to add that to the docker-compose and will see if that resolves the issue with repeated running of the spec. I did find out how to _read_ the value from the rails console, but not how to _set_ the value from the rails console. Some of that is likely my inexperience with ruby as a language. We _could_ read the value into a file, then parse the file from the test, but that seems like a really janky workaround. I hope the above works.

PatrickRice-KSC: Even better - we can set it in the gitlab.rb: https://gitlab.com/gitlab-org/omnibus-gitlab/blob/master/files/gitlab-config-template/gitlab.rb.template#L687

I'm going to open a new PR that adds that so we can merge that and then rebase for this PR.

timofurrer: Yes - that's exactly what I was hoping will work 🎉

PatrickRice-KSC: @timofurrer  - I've rebased to master (and squashed for a clean history) and re-added the test. I ran it in gitpod several times in a row with success. Looks like we're clean here!

examples/resources/gitlab_runner/resource.tf
PatrickRice-KSC: @willianpaixao  - I did some digging on this last night, and I don't think the helm chart would actually work with this resource, since there doesn't seem to be a way to short-circuit specifying the registration token and directly supply the authentication token. 

Reference on registration token being required: https://docs.gitlab.com/runner/install/kubernetes.html#required-configuration

You could potentially specify the token directly as it's injected into the runner here: https://docs.gitlab.com/runner/install/kubernetes.html#additional-configuration

I'd suggest though that until we can have someone test that, we should stick with the current documentation and add that additional piece later as needed.

internal/provider/resource_gitlab_runner_test.go
timofurrer: Great work @PatrickRice-KSC 🎉

examples/resources/gitlab_runner/resource.tf
timofurrer: > I'd suggest though that until we can have someone test that, we should stick with the current documentation and add that additional piece later as needed.

Yes, I agree. What we already have is plenty w.r.t to docs. Let's move forward and add more documentation in another iteration 🐾

internal/provider/resource_gitlab_runner.go
timofurrer: @PatrickRice-KSC this is still open. WDYT about trying to mark it as `Computed` ?

PatrickRice-KSC: @timofurrer  - whoops, sorry! Was hidden in a compressed conversation.  Yeah, I'm good with marking it as computed! I'll add that shortly so we can get this merged 🙂

timofurrer (approved):
Thanks for the good work @PatrickRice-KSC once again! 

LGTM 🎉

PR #1069, opened 2022-05-02
resource/gitlab_service_jira: several cleanup of tech debt

timofurrer:
This change set cleans up a few things in the `gitlab_service_jira` resource:

* properly use resource ID in read function
* use passthrough importer
* state in docs that `jira_issue_transition_id` attribute cannot be imported
* add import verification test steps to `_basic` test and remove dedicated `_import` test

### DIFF
--- a/internal/provider/resource_gitlab_service_jira.go
+++ b/internal/provider/resource_gitlab_service_jira.go
@@ -20,7 +20,7 @@ var _ = registerResource("gitlab_service_jira", func() *schema.Resource {
 		UpdateContext: resourceGitlabServiceJiraUpdate,
 		DeleteContext: resourceGitlabServiceJiraDelete,
 		Importer: &schema.ResourceImporter{
-			StateContext: resourceGitlabServiceJiraImportState,
+			StateContext: schema.ImportStatePassthroughContext,
 		},
 
 		Schema: map[string]*schema.Schema{
@@ -81,7 +81,7 @@ var _ = registerResource("gitlab_service_jira", func() *schema.Resource {
 				Sensitive:   true,
 			},
 			"jira_issue_transition_id": {
-				Description: "The ID of a transition that moves issues to a closed state. You can find this number under the JIRA workflow administration (Administration > Issues > Workflows) by selecting View under Operations of the desired workflow of your project. By default, this ID is set to 2.",
+				Description: "The ID of a transition that moves issues to a closed state. You can find this number under the JIRA workflow administration (Administration > Issues > Workflows) by selecting View under Operations of the desired workflow of your project. By default, this ID is set to 2. **Note**: importing this field is currently not supported.",
 				Type:        schema.TypeString,
 				Optional:    true,
 			},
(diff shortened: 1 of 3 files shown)


internal/provider/resource_gitlab_service_jira.go
```diff
@@ -166,46 +166,31 @@ func resourceGitlabServiceJiraCreate(ctx context.Context, d *schema.ResourceData
...
-	if v := jiraService.Properties.URL; v != "" {
-		d.Set("url", v)
+	d.Set("project", project)
+	d.Set("url", jiraService.Properties.URL)
+	d.Set("api_url", jiraService.Properties.APIURL)
+	d.Set("username", jiraService.Properties.Username)
+	d.Set("project_key", jiraService.Properties.ProjectKey)
+	// FIXME: The API or go-gitlab doesn't return `jira_issue_transition_id` properly,
+	//        therefore we don't overwrite the value set in the config, in case it's not set in the API
+	//        response. This currently impacts the import of this attribute.
```
PatrickRice-KSC, line 190: @timofurrer  - I assume we'll want to create an issue for this FIXME when we merge this so that we can investigate which of these two items it is 😄 .

I have a docker compose (and a jira license) that I can use to test this out, so I can take a look at it later.

PatrickRice-KSC (approved):
Overall, LGTM! 🎉

internal/provider/resource_gitlab_service_jira.go
timofurrer: @PatrickRice-KSC That would be great. 

However, I'm not sure how Jira would be involved here - I think it's a GitLab / go-gitlab issue here only, but please give it a try.
IMHO: we can also wait for that until a merge to hopefully avoid the FIXME entirely ...

timofurrer: @PatrickRice-KSC I didn't yet have time to dig into it, but just spotted this fishy part when unmarshalling the JSON from the API and fiddling with that very field: 

https://github.com/xanzy/go-gitlab/blob/65486a91e11bbb63cbf61bce23cfb48ee174b4d5/services.go#L673

PatrickRice-KSC: @timofurrer  Yeah, that definitely seems a bit fishy. I did just try this out, and attaching a JIRA instance doesn't make any difference, it still doesn't unmarshall properly, so we can rule that out.

armsnyder (dismissed):
🥳

internal/provider/resource_gitlab_service_jira.go
timofurrer: I've created an upstream issue here: https://gitlab.com/gitlab-org/gitlab/-/issues/362437

At the moment I think we can merge this. WDYT? @PatrickRice-KSC

armsnyder (approved):

internal/provider/resource_gitlab_service_jira.go
PatrickRice-KSC: Agreed!

PatrickRice-KSC (approved):

PR #1073, opened 2022-05-05
Implement resources and data sources for GitLab Agent for Kubernetes API

timofurrer:
This change set adds the following new resources / data sources:

* `resource gitlab_project_cluster_agent`
* `data gitlab_project_cluster_agent`
* `data gitlab_project_cluster_agents`

Corresponding to the new API here: https://docs.gitlab.com/ee/api/cluster_agents.html

This PR is currently blocked by:

* https://github.com/xanzy/go-gitlab/pull/1461

@nagyv I had quite some discussions about the naming of that API over at GitLab. In the provider we usually name resources / data sources similar to the API path, thus `/projects/:id/cluster_agents` becomes `gitlab_project_cluster_agent` ... Is that okay with you, too? Or do you have any other preference for that particular resource name?


timofurrer:
@PatrickRice-KSC Yes, please. The code shouldn't be affected by the draft state ;)

nagyv:
@timofurrer Good question. I don't have a very strong opinion, but a would prefer to call it `gitlab_cluster_agent`. 

The project relationship is there really only to store the agent configurations and management views. The idea is to have agents that maintain a connection with a cluster, and this is independent of the concept of a project.

At the same time, to keep consistency, I don't mind the `gitlab_project_cluster_agent` resource name.

internal/provider/data_source_gitlab_project_cluster_agent_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+					data "gitlab_project_cluster_agent" "this" {
+						project           = gitlab_project_cluster_agent.this.project
+						agent_id          = gitlab_project_cluster_agent.this.agent_id
+					}
+					`, testProject.ID,
+				),
+				Check: resource.ComposeTestCheckFunc(
+					testAccDataSourceGitlabProjectClusterAgent("gitlab_project_cluster_agent.this", "data.gitlab_project_cluster_agent.this"),
+				),
```
PatrickRice-KSC, line 35: @timofurrer  - I'm a little confused here, the `ComposeTestCheckFunc` is meant to combine multiple Test Functions into one, but there is only one test function here. Is there a reason to use the `ComposeTestCheckFunc` here?

Side note, I think I wrote a helper for this in a previous PR that we should probably deprecate/remove. Whoops!

PatrickRice-KSC (changes requested):
I didn't make it though all the files yet, but I have a couple comments/questions so far 🙂

internal/provider/data_source_gitlab_project_cluster_agent_test.go
```diff
@@ -0,0 +1,60 @@
...
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: providerFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: fmt.Sprintf(`
+					resource "gitlab_project_cluster_agent" "this" {
+						project           = "%d"
+						name              = "agent-1"
+					}
```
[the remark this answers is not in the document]
timofurrer: good point, done!

internal/provider/data_source_gitlab_project_cluster_agent_test.go
timofurrer: Copy & Paste mistake :D 

changed it anyways here.

internal/provider/schema_gitlab_project_cluster_agent.go
```diff
@@ -0,0 +1,50 @@
+package provider
```
PatrickRice-KSC, line 1: Hm... new approach for the schema separating this off into a new file. I assume this is done to prevent a lot of code duplication between the data source and the resource? It feels like we may be hitting a critical mass where we want another package for the schema files (and maybe helpers?), just so we don't have [schema, resource, resource tests, data sources, data source tests, helpers] all in one package.

PatrickRice-KSC (commented):
One one (more general) comment, otherwise it's a bit difficult to confirm with all the compile issues, but it overall looks solid to me :)

internal/provider/schema_gitlab_project_cluster_agent.go
timofurrer: Yes, it's exactly about reducing code duplication. I'm not sure about the packages in this case. What benefits would it bring besides "the critical mass" (which IMHO is not a very good reason alone - convince me otherwise, please 💃 ) ?

armsnyder: It would have to be for productivity reasons, right? For me personally I am ok with at least having the resources, data sources, schemas, and resource tests in the same package because the filenames make it easy to navigate. I could see wanting to move our various helpers and "main" provider code into new package/s though, since those can be hard to locate.

PatrickRice-KSC: @timofurrer  - Yeah, as adam says it's more around making it easy to find what you're looking for. Part of me kinda wishes we did <resource_name>-(resource|data source|schema) because then they'd all be grouped together when you were developing one of them, but I suspect that would just cause other things to be difficult to find.

I'm just a bit worried that with the number of files in the package it may get unweildy for people to find what they need to work on. The counter argument to this is of course... that's what an IDE is for. Ctrl+Click and find file make locating files easier. So I'm not passionate about the commentary, it was just something that went through my mind when I was reviewing and figured it was worth leaving a comment to see what you two thought. 

I do like the schema pattern though, so no issues there!

timofurrer: I've created https://github.com/gitlabhq/terraform-provider-gitlab/issues/1083 as a follow up of this conversion. Let's continue the discussion there :)

timofurrer:
> @timofurrer Good question. I don't have a very strong opinion, but a would prefer to call it `gitlab_cluster_agent`.
> 
> The project relationship is there really only to store the agent configurations and management views. The idea is to have agents that maintain a connection with a cluster, and this is independent of the concept of a project.
> 
> At the same time, to keep consistency, I don't mind the `gitlab_project_cluster_agent` resource name.

@armsnyder you've put a 👍  there ... What was your preference, `gitlab_cluster_agent` or `gitlab_project_cluster_agent` ? Or both would be okay?

armsnyder:
> > @timofurrer Good question. I don't have a very strong opinion, but a would prefer to call it `gitlab_cluster_agent`.
> > The project relationship is there really only to store the agent configurations and management views. The idea is to have agents that maintain a connection with a cluster, and this is independent of the concept of a project.
> > At the same time, to keep consistency, I don't mind the `gitlab_project_cluster_agent` resource name.
> 
> @armsnyder you've put a 👍 there ... What was your preference, `gitlab_cluster_agent` or `gitlab_project_cluster_agent` ? Or both would be okay?

Both are okay. 👍 If I had to choose I would go with `gitlab_cluster_agent`.

timofurrer:
I've renamed the resources to `gitlab_cluster_agent` 👍

timofurrer:
blocked by: https://github.com/gitlabhq/terraform-provider-gitlab/pull/1095

armsnyder (dismissed):
lgtm!

PatrickRice-KSC:
We're green across the board, I assume you're still waiting for 1095 though before you merge, is that the plan?

PatrickRice-KSC (approved):

armsnyder (approved):

PR #1078, opened 2022-05-11
Moving a subgroup does not force a recreation of the group and no longer deletes the underlying projects

### LINKED ISSUE #574: Moving subgroup deletes underlying projects
### Terraform Version
```
Terraform v0.14.6
+ provider registry.terraform.io/gitlabhq/gitlab v3.4.0
```

### Affected Resource(s)
- gitlab_group
- gitlab_project

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

```hcl
resource "gitlab_group" "ansible_roles" {
  name                    = "ansible-roles"
  path                    = "ansible-roles"
  description             = "Ansible roles"
  parent_id               = gitlab_group.ace.id        # <---- this is what I updated
  subgroup_creation_level = "owner"
  project_creation_level  = "maintainer"
  request_access_enabled  = true
}

resource "gitlab_project" "ansible_roles" {
  for_each = local.ansible_roles

  name                                             = each.key
  description                                      = each.value
  default_branch                                   = "master"
  namespace_id                                     = gitlab_group.ansible_roles.id
  visibility_level                                 = "internal"
  initialize_with_readme                           = true
  approvals_before_merge                           = 0
  only_allow_merge_if_pipeline_succeeds            = false
  only_allow_merge_if_all_discussions_are_resolved = false
  remove_source_branch_after_merge                 = true
  container_registry_enabled                       = false
  issues_enabled                                   = true
  merge_requests_enabled                           = true
  packages_enabled                                 = true
  pipelines_enabled                                = true
  shared_runners_enabled                           = false
  snippets_enabled                                 = true
  wiki_enabled                                     = false

  lifecycle {
    # guard against accidental destruction of this repository
    prevent_destroy = true
  }
}
```

### Plan output

```

  # gitlab_group.ansible_roles must be replaced
-/+ resource "gitlab_group" "ansible_roles" {
      ~ full_name                         = "old_path / ansible-roles" -> (known after apply)
      ~ full_path                         = "old_path/ansible-roles" -> (known after apply)
      ~ id                                = "3589" -> (known after apply)
        name                              = "ansible-roles"
      ~ parent_id                         = 2147 -> 6994 # forces replacement
      ~ runners_token                     = (sensitive value)
      ~ visibility_level                  = "internal" -> (known after apply)
      ~ web_url                           = "https://..{redacted}...ansible-roles" -> (known after apply)
        # (12 unchanged attributes hidden)
    }

  # gitlab_project.ansible_roles["base"] will be updated in-place
  ~ resource "gitlab_project" "ansible_roles" {
        id                                               = "4088"
        name                                             = "base"
      ~ namespace_id                                     = 3589 -> (known after apply)
        tags                                             = []
        # (29 unchanged attributes hidden)
    }
.... 
# repeat for other ansible role projects
```

### Expected Behavior

- Subgroup would relocated its new parent group
- Projects would follow

or at the very least there would be an error indicating that Terraform couldn't destroy my projects given `prevent=destroy`.

### Actual Behavior

- Subgroup was destroyed and recreated under new parent group (normal)
- All projects appear to have been deleted

### Apply Output

```
gitlab_group.ansible_roles: Destroying... [id=3589]
gitlab_group.ansible_roles: Still destroying... [id=3589, 10s elapsed]
gitlab_group.ansible_roles: Still destroying... [id=3589, 20s elapsed]
gitlab_group.ansible_roles: Destruction complete after 26s
gitlab_group.ansible_roles: Creating...
gitlab_group.ansible_roles: Creation complete after 1s [id=7048]
gitlab_project.ansible_roles["base"]: Modifying... [id=4088]
gitlab_project.ansible_roles["vault"]: Modifying... [id=4359]
....
# repeat for remaining ansible role projects

Error: PUT https://..{redacted}../api/v4/projects/4359/transfer: 404 {message: 404 Project Not Found}

  on projects.tf line 126, in resource "gitlab_project" "ansible_roles":
 126: resource "gitlab_project" "ansible_roles" {

Error: PUT https://..{redacted}../api/v4/projects/4088/transfer: 404 {message: 404 Project Not Found}

  on projects.tf line 126, in resource "gitlab_project" "ansible_roles":
 126: resource "gitlab_project" "ansible_roles" {

...
# repeat for remaining ansible role projects
```

### Steps to Reproduce

- Change subgroup parent id
- `terraform apply`

### Important Factoids

Pretty severe given:

1. moving a group (destroy/recreate) deletes all projects despite no indication from plan that this is what will happen
2. destructive action bypasses lifecycle destruction prevention

gro-gg:
## Description

fixes #574 

    - Create the resource
    - Update the attributes
    - Import the resource

### Tests

To test that the nested group is not recreated, I misused the AutoDevopsEnabled attribute. It ensures that the value is still true after transferring the nested group. If the group was recreated, it would be false again (default value). Please let me know if you have a better idea how to test this.

### DIFF
--- a/internal/provider/resource_gitlab_group.go
+++ b/internal/provider/resource_gitlab_group.go
@@ -369,7 +369,7 @@ func resourceGitlabGroupUpdate(ctx context.Context, d *schema.ResourceData, meta
 	}
 
 	if d.HasChange("parent_id") {
-		diagnostic := transferSubGroup(d, client)
+		diagnostic := transferSubGroup(ctx, d, client)
 		if diagnostic.HasError() {
 			return diagnostic
 		}
@@ -378,7 +378,7 @@ func resourceGitlabGroupUpdate(ctx context.Context, d *schema.ResourceData, meta
 	return resourceGitlabGroupRead(ctx, d, meta)
 }
 
-func transferSubGroup(d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+func transferSubGroup(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client := meta.(*gitlab.Client)
 	o, n := d.GetChange("parent_id")
 	parentId, ok := n.(int)
@@ -394,7 +394,7 @@ func transferSubGroup(d *schema.ResourceData, meta interface{}) diag.Diagnostics
 		log.Printf("[DEBUG] turn gitlab group %s from %v to a new top-level group", d.Id(), o)
 	}
 
-	_, _, err := client.Groups.TransferSubGroup(d.Id(), opt)
+	_, _, err := client.Groups.TransferSubGroup(d.Id(), opt, gitlab.WithContext(ctx))
 	if err != nil {
 		return diag.Errorf("error transfering group %s to new parent group %v: %s", d.Id(), parentId, err)
 	}
--- a/internal/provider/resource_gitlab_group.go
+++ b/internal/provider/resource_gitlab_group.go
@@ -141,7 +141,6 @@ var _ = registerResource("gitlab_group", func() *schema.Resource {
 				Description: "Id of the parent group (creates a nested group).",
 				Type:        schema.TypeInt,
 				Optional:    true,
-				ForceNew:    true,
 				Default:     0,
 			},
 			"runners_token": {
@@ -369,7 +368,38 @@ func resourceGitlabGroupUpdate(ctx context.Context, d *schema.ResourceData, meta
 		return diag.FromErr(err)
 	}
 
+	if d.HasChange("parent_id") {
+		diagnostic := transferSubGroup(d, client)
+		if diagnostic.HasError() {
+			return diagnostic
+		}
+	}
+
 	return resourceGitlabGroupRead(ctx, d, meta)
+}
+
+func transferSubGroup(d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*gitlab.Client)
+	o, n := d.GetChange("parent_id")
+	parentId, ok := n.(int)
+	if !ok {
+		return diag.Errorf("error converting parent_id %v into an int", n)
+	}
+
+	opt := &gitlab.TransferSubGroupOptions{}
+	if parentId != 0 {
+		log.Printf("[DEBUG] transfer gitlab group %s from %v to new parent group %v", d.Id(), o, n)
+		opt.GroupID = gitlab.Int(parentId)
+	} else {
+		log.Printf("[DEBUG] turn gitlab group %s from %v to a new top-level group", d.Id(), o)
+	}
+
+	_, _, err := client.Groups.TransferSubGroup(d.Id(), opt)
+	if err != nil {
+		return diag.Errorf("error transfering group %s to new parent group %v: %s", d.Id(), parentId, err)
+	}
+
+	return nil
 }
 
 func resourceGitlabGroupDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
--- a/internal/provider/resource_gitlab_group.go
+++ b/internal/provider/resource_gitlab_group.go
@@ -378,8 +378,7 @@ func resourceGitlabGroupUpdate(ctx context.Context, d *schema.ResourceData, meta
 	return resourceGitlabGroupRead(ctx, d, meta)
 }
 
-func transferSubGroup(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	client := meta.(*gitlab.Client)
+func transferSubGroup(ctx context.Context, d *schema.ResourceData, client *gitlab.Client) diag.Diagnostics {
 	o, n := d.GetChange("parent_id")
 	parentId, ok := n.(int)
 	if !ok {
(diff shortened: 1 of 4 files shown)


internal/provider/resource_gitlab_group.go
```diff
@@ -369,9 +368,40 @@ func resourceGitlabGroupUpdate(ctx context.Context, d *schema.ResourceData, meta
...
+
+	opt := &gitlab.TransferSubGroupOptions{}
+	if parentId != 0 {
+		log.Printf("[DEBUG] transfer gitlab group %s from %v to new parent group %v", d.Id(), o, n)
+		opt.GroupID = gitlab.Int(parentId)
+	} else {
+		log.Printf("[DEBUG] turn gitlab group %s from %v to a new top-level group", d.Id(), o)
+	}
+
+	_, _, err := client.Groups.TransferSubGroup(d.Id(), opt)
```
timofurrer, line 397: Let's use the context here from the CRUD function (has to be passed to the `transferSubGroup` function):

```suggestion
	_, _, err := client.Groups.TransferSubGroup(d.Id(), opt, gitlab.WithContext(ctx))
```

internal/provider/resource_gitlab_group.go
```diff
@@ -369,9 +368,40 @@ func resourceGitlabGroupUpdate(ctx context.Context, d *schema.ResourceData, meta
...
+		diagnostic := transferSubGroup(d, client)
+		if diagnostic.HasError() {
+			return diagnostic
+		}
+	}
+
 	return resourceGitlabGroupRead(ctx, d, meta)
 }
 
+func transferSubGroup(d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
timofurrer, line 381: (non-blocking): I think it would make sense to already pass the GitLab client and attributes needed directly to this function. Also, return an `error` instead of `Diagnostics` - we don't use the warning feature anyways.

WDYT?

internal/provider/resource_gitlab_group_test.go
```diff
@@ -199,27 +201,26 @@ func TestAccGitlabGroup_nested(t *testing.T) {
...
-			//			ProjectCreationLevel:  "maintainer", // default value
-			//			SubGroupCreationLevel: "owner",      // default value
-			//			TwoFactorGracePeriod:  48,           // default value
-			//			DefaultBranchProtection: 2,          // default value
-			// 			Parent:      &group,
-			// 		}),
-			// 	),
-			// },
+			{
+				Config: testAccGitlabNestedGroupConfig(rInt),
```
timofurrer, line 205: A few ideas to make this test step more transparent:

* doesn't the group id stay the same after the transfer? We may be able to check for this instead of the non-intuitive `auto_devops_enabled` attribute check
* another idea could actually be to verify what you wanted to achieve (or at least what the PR title states): testing if underlying projects are preserved during a group transfer. Since these are acceptance tests I think it's okay to even introduce a separate test case to test that particular behavior.

WDYT?

timofurrer (changes requested):
Thanks @gro-gg for the change! 

I've left a few comments. 

Back to you 🏓

internal/provider/resource_gitlab_group.go
gro-gg: completely agree, I forgot

internal/provider/resource_gitlab_group.go
gro-gg: no emotions, I will fix it...

internal/provider/resource_gitlab_group_test.go
gro-gg: Checking the group id was my first idea. But this means that I have to keep a state between the test steps. This seemed a bit ugly to me. I will try this approach again though. :+1:

timofurrer: I don't think it's necessarily a bad idea. Here is an example where it's already done: 
https://github.com/gitlabhq/terraform-provider-gitlab/blob/145117ca03d4f91a5fd2fec325c8a0c60106913e/internal/provider/resource_gitlab_project_issue_test.go#L92

gro-gg: I have also tried the idea with the projects. But I don't think that works with the setup of the test steps. If the project is created in the Config with Terraform, then it is also created again after a re-create of the group. So I need to test if the project was recreated or not. So I am facing the same problem as with the groups. :man_juggling:

timofurrer: You can create test projects outside of the terraform config, e.g.:

https://github.com/gitlabhq/terraform-provider-gitlab/blob/145117ca03d4f91a5fd2fec325c8a0c60106913e/internal/provider/resource_gitlab_project_tag_test.go#L19

timofurrer: ... ah I see, then you are not able to easily attach it to that group under test 🤔

gro-gg: I've added a test, that ensures the group id does not change

gro-gg:
done my work and rebased on master
ready for review :ping_pong:

timofurrer (approved):

timofurrer (approved):
Thanks for the contribution @gro-gg ! 🎉 

It looks good to me. 🚢 

Reminder for myself: we need to do something about the "testability" of these larger resources. The tests cases are somewhat a intransparent mix of feature testing ...

Issue #1080, opened 2022-05-13
Project issue use o references Doc improvement


adriano-fonseca:
### Feature Description

I have been trying without success to pass other issues as a reference to a project issue. Folloing the documentation that states the following:

`[references](https://registry.terraform.io/providers/gitlabhq/gitlab/latest/docs/resources/project_issue#references) (Map of String) The references of the issue.`

I managed to get one reference set using terraform function to map like below:

```
references = tomap({
    "full"     = "professional-services-delivery/partner_c/customer_f/PSE-00150#2",
    "relative" = "#2",
    "short"    = "#2"
  })
```

 However, no to sure how syntax I should use to pass more than one issue. We could improve the docs to facilitate more advanced use cases.

This is what I tried with no success. 
```
references = [{
    "full"     = professional-services-delivery/partner_c/customer_f/PSE-00150#2",
    "relative" = "#2",
    "short"    = "#2"
    },
    {
      "full"     = "professional-services-delivery/partner_c/customer_f/PSE-00150#4",
      "relative" = "#4",
      "short"    = "#4"
    }
  ]
```

Any help is more than welcome. With the proper guidance, I could submit this PR.

### Do you want to implement this?

- [X] I would like to implement this myself 👷

adriano-fonseca:
@timofurrer any advice on the correct syntax to pass several references to a project issue?

timofurrer:
@adriano-fonseca the `references` field is actually a map of reference for the issue at hand (like the short, relative and full reference) and not the references to other issues.

The reason you were able to set it at all is because this attribute (including others) are mistakenly marked as `Optional` which they are actually not. Fixed it in here: https://github.com/gitlabhq/terraform-provider-gitlab/pull/1081

There also doesn't seem to be a REST API to reference an issue from another. Or did you find one which you'd like to see in this resource?

adriano-fonseca:
I got it. A reference for the issue itself. Thanks for adjusting the docs.
I did some research and the API End point actually exists. See in [issue links docs](https://docs.gitlab.com/ee/api/issue_links.html)
I work at GitLab and we are using it heavily internally to automate PS project creation. 
This is something that I am missing hard, also epics and milestones creation:

https://docs.gitlab.com/ee/api/epics.html#new-epic
https://docs.gitlab.com/ee/api/milestones.html#create-new-milestone

timofurrer:
Oh great! Thanks for the links 👍 

[...]

PR #1081, opened 2022-05-13
resource/gitlab_project_issue: Remove `Optional` from `Computed`-only attributes

timofurrer:
Some attributes were mistakenly marked as `Optional`, but they are
`Computed`-only.

See #1080

### DIFF
--- a/docs/resources/project_issue.md
+++ b/docs/resources/project_issue.md
@@ -54,8 +54,6 @@ output "welcome_issue_web_url" {
 ### Optional
 
 - `assignee_ids` (Set of Number) The IDs of the users to assign the issue to.
-- `closed_at` (String) When the issue was closed. Date time string, ISO 8601 formatted, for example 2016-03-11T03:45:40Z.
-- `closed_by_user_id` (Number) The ID of the user that closed the issue. Use `gitlab_user` data source to get more information about the user.
 - `confidential` (Boolean) Set an issue to be confidential.
 - `created_at` (String) When the issue was created. Date time string, ISO 8601 formatted, for example 2016-03-11T03:45:40Z. Requires administrator or project/group owner rights.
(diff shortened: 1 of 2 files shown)


PatrickRice-KSC:
@timofurrer  - LGTM! 🎉

PatrickRice-KSC (dismissed):

docs/resources/project_issue.md
```diff
@@ -55,7 +55,6 @@ output "welcome_issue_web_url" {
 
 - `assignee_ids` (Set of Number) The IDs of the users to assign the issue to.
 - `closed_at` (String) When the issue was closed. Date time string, ISO 8601 formatted, for example 2016-03-11T03:45:40Z.
-- `closed_by_user_id` (Number) The ID of the user that closed the issue. Use `gitlab_user` data source to get more information about the user.
 - `confidential` (Boolean) Set an issue to be confidential.
 - `created_at` (String) When the issue was created. Date time string, ISO 8601 formatted, for example 2016-03-11T03:45:40Z. Requires administrator or project/group owner rights.
```
armsnyder, line 59: Did you miss some? I don't see how `created_at`, `closed_at`, or `updated_at ` can also be Optional, for example.

armsnyder (changes requested):

docs/resources/project_issue.md
timofurrer: Indeed, I've missed `closed_at`. However, both [`created_at`](https://docs.gitlab.com/ee/api/issues.html#:~:text=is%20false.-,created_at,-string) and [`updated_at`](https://docs.gitlab.com/ee/api/issues.html#:~:text=of%20an%20issue-,updated_at,-string) can be set using the created and update API endpoints, respectively ..

armsnyder: How bizarre 😆

armsnyder (approved):

PR #1048, opened 2022-04-30
resource/gitlab_topic: Add `title` argument support for GitLab 15.0

timofurrer:
This change set adds support for the newly introduced `title` field in GitLab 15.0 of the Topics API.

To support previous GitLab instances the `title` resource attribute is optional even though in the API it's required.

However, it's an error if the user sets the `title` for GitLab releases prior to 15.0 and if the user doesn't set it for GitLab releases from 15.0 on ...

Once we integrate 15.0 we have to adjust the tests, too to properly support the resource.

It's currently blocked by a new go-gitlab release we require.

Closes: #1043

PR #1095, opened 2022-05-22
Support Gitlab 15.0

timofurrer:
GitLab 15.0 has been released today 🎉 

This change set makes the provider compatible. We've missed a few things:

* The old K8s cluster integration was removed, therefore I've disabled the tests for GitLab 15.0 and onwards
* The `gitlab_managed_license` resource wasn't ready regarding the new `approval_status` values

@PatrickRice-KSC can you please verify the changes I did in the managed license resource? Did I miss something?

@PatrickRice-KSC @armsnyder @nagyv I've also changed the workflows to test the last three GitLab releases. WDYT? 🏓

### DIFF
--- a/internal/provider/data_source_gitlab_repository_file_test.go
+++ b/internal/provider/data_source_gitlab_repository_file_test.go
@@ -66,7 +66,6 @@ resource "gitlab_repository_file" "foo" {
 	branch = "main"
 	content = base64encode("Meow goes the cat")
 	commit_message = "feat: Meow"
-	execute_filemode = true
 }
 
 data "gitlab_repository_file" "foo" {
--- a/internal/provider/helper_test.go
+++ b/internal/provider/helper_test.go
@@ -68,22 +68,6 @@ func isRunningInCE() (bool, error) {
 	return !isEE, err
 }
 
-// orSkipFunc accepts many skipFunc and returns "true" if any returns true.
-func orSkipFunc(input ...SkipFunc) SkipFunc {
-	return func() (bool, error) {
-		for _, item := range input {
-			result, err := item()
-			if err != nil {
-				return false, err
-			}
-			if result {
-				return result, nil
-			}
-		}
-		return false, nil
-	}
-}
-
 // testAccCheckEE is a test helper that skips the current test if the GitLab version is not GitLab Enterprise.
 // This is useful when the version needs to be checked during setup, before the Terraform acceptance test starts.
 func testAccCheckEE(t *testing.T) {
--- a/internal/provider/helper_test.go
+++ b/internal/provider/helper_test.go
@@ -4,6 +4,7 @@
 package provider
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"os"
@@ -96,6 +97,37 @@ func testAccCheckEE(t *testing.T) {
 	if !strings.HasSuffix(version.Version, "-ee") {
 		t.Skipf("Test is skipped for non-Enterprise version of GitLab (was %q)", version.String())
 	}
+}
+
+func testAccRequiresLessThan(t *testing.T, requiredMaxVersion string) {
+	isLessThan, err := isGitLabVersionLessThan(context.TODO(), testGitlabClient, requiredMaxVersion)()
+	if err != nil {
+		t.Fatalf("Failed to fetch GitLab version: %+v", err)
+	}
+
+	if !isLessThan {
+		t.Skipf("This test is only valid for GitLab versions less than %s", requiredMaxVersion)
+	}
+}
+
+func testAccRequiresAtLeast(t *testing.T, requiredMinVersion string) {
+	isAtLeast, err := isGitLabVersionAtLeast(context.TODO(), testGitlabClient, requiredMinVersion)()
+	if err != nil {
+		t.Fatalf("Failed to fetch GitLab version: %+v", err)
+	}
+
+	if !isAtLeast {
+		t.Skipf("This test is only valid for GitLab versions newer than %s", requiredMinVersion)
+	}
+}
+
+func testAccIsRunningAtLeast(t *testing.T, requiredMinVersion string) bool {
+	isAtLeast, err := isGitLabVersionAtLeast(context.TODO(), testGitlabClient, requiredMinVersion)()
+	if err != nil {
+		t.Fatalf("Failed to fetch GitLab version: %+v", err)
+	}
+
+	return isAtLeast
 }
 
 // testAccCurrentUser is a test helper for getting the current user of the provided client.
--- a/internal/provider/resource_gitlab_managed_license.go
+++ b/internal/provider/resource_gitlab_managed_license.go
@@ -53,8 +53,8 @@ var _ = registerResource("gitlab_managed_license", func() *schema.Resource {
 				ForceNew:         false,
 				ValidateFunc:     validation.StringInSlice(managedLicenseAllowedValues, true),
 				DiffSuppressFunc: checkDeprecatedValuesForDiff,
-				Description: fmt.Sprintf(`The approval status of the license. Valid values are: %s. "approved" and "blacklisted" 
-				have been deprecated in favor of "allowed" and "denied"; use "allowed" and "denied" for GitLab versions 15.0 and higher. 
+				Description: fmt.Sprintf(`The approval status of the license. Valid values are: %s. "approved" and "blacklisted"
+				have been deprecated in favor of "allowed" and "denied"; use "allowed" and "denied" for GitLab versions 15.0 and higher.
 				Prior to version 15.0 and after 14.6, the values are equivalent.`, renderValueListForDocs(managedLicenseAllowedValues)),
 			},
 		},
@@ -65,9 +65,14 @@ func resourceGitlabManagedLicenseCreate(ctx context.Context, d *schema.ResourceD
 	client := meta.(*gitlab.Client)
 	project := d.Get("project").(string)
 
+	approvalStatus, err := stringToApprovalStatus(ctx, client, d.Get("approval_status").(string))
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
 	options := &gitlab.AddManagedLicenseOptions{
 		Name:           gitlab.String(d.Get("name").(string)),
-		ApprovalStatus: stringToApprovalStatus(d.Get("approval_status").(string)),
+		ApprovalStatus: approvalStatus,
 	}
 
 	log.Printf("[DEBUG] create gitlab Managed License on Project %s, with Name %s", project, *options.Name)
@@ -131,12 +136,13 @@ func resourceGitlabManagedLicenseUpdate(ctx context.Context, d *schema.ResourceD
 		return diag.FromErr(err)
 	}
 
-	opts := &gitlab.EditManagedLicenceOptions{
-		ApprovalStatus: stringToApprovalStatus(d.Get("approval_status").(string)),
+	approvalStatus, err := stringToApprovalStatus(ctx, client, d.Get("approval_status").(string))
+	if err != nil {
+		return diag.FromErr(err)
 	}
 
-	if d.HasChange("approval_status") {
-		opts.ApprovalStatus = stringToApprovalStatus(d.Get("approval_status").(string))
+	opts := &gitlab.EditManagedLicenceOptions{
+		ApprovalStatus: approvalStatus,
 	}
 
 	log.Printf("[DEBUG] update gitlab Managed License %s", d.Id())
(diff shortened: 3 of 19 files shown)


internal/provider/data_source_gitlab_repository_file_test.go
```diff
@@ -59,20 +60,24 @@ func testAccDataSourceGitlabRepositoryFile(src, n string) resource.TestCheckFunc
 }
 
 func testAccDataGitlabRepositoryFile(project string) string {
+	var executeFilemodeConfig string
+	if isExecuteFilemodeSupported, err := isGitLabVersionAtLeast(context.Background(), testGitlabClient, "14.10")(); err == nil && isExecuteFilemodeSupported {
+		executeFilemodeConfig = "execute_filemode = true"
+	}
```
PatrickRice-KSC, line 66: @timofurrer  - Out of curiosity, why this method instead of creating a separate test with a skipfunc? This certainly works, but it's not necessarily the pattern we've followed in the past for cross-version tests.

internal/provider/resource_gitlab_topic_test.go
```diff
@@ -256,6 +258,46 @@ func TestAccGitlabTopic_softDestroy(t *testing.T) {
...
 
+func TestAccGitlabTopic_titleSupport(t *testing.T) {
+	rInt := acctest.RandInt()
+
+	resource.Test(t, resource.TestCase{
+		ProviderFactories: providerFactories,
+		CheckDestroy:      testAccCheckGitlabTopicDestroy,
+		Steps: []resource.TestStep{
+			{
+				SkipFunc: isGitLabVersionAtLeast(context.Background(), testGitlabClient, "15.0"),
```
PatrickRice-KSC, line 269: @timofurrer  - since every step is using the SkipFunc, I'd suggest we move this to a pre-check.

internal/provider/resource_gitlab_topic_test.go
```diff
@@ -370,46 +412,71 @@ func testAccCheckGitlabTopicSoftDestroy(s *terraform.State) (err error) {
 }
 
 func testAccGitlabTopicRequiredConfig(rInt int) string {
+	var titleConfig string
+	if titleIsSupported, err := isGitLabVersionAtLeast(context.Background(), testGitlabClient, "15.0")(); err == nil && titleIsSupported {
+		titleConfig = fmt.Sprintf(`title = "Foo Req %d"`, rInt)
+	}
```
PatrickRice-KSC, line 418: Same comment as this pattern last time, seems like 15+ should maybe be a separate test. This matches our pattern of using in-line configs now anyway, right?

PatrickRice-KSC (changes requested):
Couple very small changes!

PatrickRice-KSC:
@timofurrer  - I like the workflow changes, nice!

internal/provider/data_source_gitlab_repository_file_test.go
```diff
@@ -59,20 +60,24 @@ func testAccDataSourceGitlabRepositoryFile(src, n string) resource.TestCheckFunc
 }
 
 func testAccDataGitlabRepositoryFile(project string) string {
+	var executeFilemodeConfig string
+	if isExecuteFilemodeSupported, err := isGitLabVersionAtLeast(context.Background(), testGitlabClient, "14.10")(); err == nil && isExecuteFilemodeSupported {
```
armsnyder, line 64: As an FYI for the future, you can use `context.TODO()` in cases where you need a context but do not have one. They're functionally identical, but `context.Background()` communicates a different intent.

internal/provider/helper_test.go
```diff
@@ -98,6 +83,12 @@ func testAccCheckEE(t *testing.T) {
 	}
 }
 
+func testAccSkipIsRunningOlderThan(t *testing.T, requiredMinVersion string) {
+	if notSupported, err := isGitLabVersionAtLeast(context.Background(), testGitlabClient, requiredMinVersion)(); err != nil || notSupported {
```
armsnyder, line 87: if err != nil then you should fail the test with `t.Fail()`

internal/provider/resource_gitlab_topic_test.go
```diff
@@ -370,46 +412,71 @@ func testAccCheckGitlabTopicSoftDestroy(s *terraform.State) (err error) {
 }
 
 func testAccGitlabTopicRequiredConfig(rInt int) string {
+	var titleConfig string
+	if titleIsSupported, err := isGitLabVersionAtLeast(context.Background(), testGitlabClient, "15.0")(); err == nil && titleIsSupported {
```
armsnyder, line 416: error handling needed

internal/provider/data_source_gitlab_repository_file_test.go
armsnyder, line 64: also error handling needed

internal/provider/resource_gitlab_managed_license.go
```diff
@@ -152,20 +148,26 @@ func resourceGitlabManagedLicenseUpdate(ctx context.Context, d *schema.ResourceD
...
-		"blacklisted": gitlab.LicenseBlacklisted,
-
-		// This is counter-intuitive, but currently the API response from the non-deprecated
-		// values is the deprecated values. So we have to map them here.
-		"allowed": gitlab.LicenseApproved,
-		"denied":  gitlab.LicenseBlacklisted,
-	}
+func stringToApprovalStatus(ctx context.Context, client *gitlab.Client, s string) *gitlab.LicenseApprovalStatusValue {
+	var value gitlab.LicenseApprovalStatusValue
+	if notSupported, err := isGitLabVersionAtLeast(ctx, client, "15.0")(); err != nil || notSupported {
```
armsnyder, line 153: error handling needed

internal/provider/resource_gitlab_repository_file_test.go
```diff
@@ -345,6 +346,10 @@ resource "gitlab_repository_file" "this" {
 }
 
 func testAccGitlabRepositoryFileUpdateConfig(projectID int) string {
+	var executeFilemodeConfig string
+	if isExecuteFilemodeSupported, err := isGitLabVersionAtLeast(context.Background(), testGitlabClient, "14.10")(); err == nil && isExecuteFilemodeSupported {
```
armsnyder, line 350: error handling needed

internal/provider/resource_gitlab_topic_test.go
```diff
@@ -370,46 +412,71 @@ func testAccCheckGitlabTopicSoftDestroy(s *terraform.State) (err error) {
...
 resource "gitlab_topic" "foo" {
   name = "foo-req-%d"
-}`, rInt)
+  %s
+}`, rInt, titleConfig)
 }
 
 func testAccGitlabTopicFullConfig(rInt int) string {
+	var titleConfig string
+	if titleIsSupported, err := isGitLabVersionAtLeast(context.Background(), testGitlabClient, "15.0")(); err == nil && titleIsSupported {
```
armsnyder, line 428: error handling needed

internal/provider/resource_gitlab_topic_test.go
```diff
@@ -370,46 +412,71 @@ func testAccCheckGitlabTopicSoftDestroy(s *terraform.State) (err error) {
...
   description = "Terraform acceptance tests"
   avatar      = "${path.module}/testdata/gitlab_topic/avatar.png"
   avatar_hash = filesha256("${path.module}/testdata/gitlab_topic/avatar.png")
-}`, rInt)
+}`, rInt, titleConfig)
 }
 
 func testAccGitlabTopicFullUpdatedAvatarConfig(rInt int) string {
+	var titleConfig string
+	if titleIsSupported, err := isGitLabVersionAtLeast(context.Background(), testGitlabClient, "15.0")(); err == nil && titleIsSupported {
```
armsnyder, line 443: missing error handling

internal/provider/resource_gitlab_topic_test.go
```diff
@@ -370,46 +412,71 @@ func testAccCheckGitlabTopicSoftDestroy(s *terraform.State) (err error) {
...
   description = "Terraform acceptance tests"
   avatar 	  = "${path.module}/testdata/gitlab_topic/avatar-update.png"
   avatar_hash = filesha256("${path.module}/testdata/gitlab_topic/avatar-update.png")
-}`, rInt)
+}`, rInt, titleConfig)
 }
 
 func testAccGitlabTopicAvatarWithoutHashConfig(rInt int) string {
+	var titleConfig string
+	if titleIsSupported, err := isGitLabVersionAtLeast(context.Background(), testGitlabClient, "15.0")(); err == nil && titleIsSupported {
```
armsnyder, line 458: error handling needed

internal/provider/resource_gitlab_topic_test.go
```diff
@@ -370,46 +412,71 @@ func testAccCheckGitlabTopicSoftDestroy(s *terraform.State) (err error) {
...
   name   = "foo-%d"
+  %s
   avatar = "${path.module}/testdata/gitlab_topic/avatar.png"
-}`, rInt)
+}`, rInt, titleConfig)
 }
 
 func testAccGitlabTopicSoftDestroyConfig(rInt int) string {
+	var titleConfig string
+	if titleIsSupported, err := isGitLabVersionAtLeast(context.Background(), testGitlabClient, "15.0")(); err == nil && titleIsSupported {
```
armsnyder, line 471: error handling needed

internal/provider/resource_gitlab_managed_license_test.go
```diff
@@ -37,7 +37,7 @@ func TestAccGitlabManagedLicense_basic(t *testing.T) {
 				Config:   testManagedLicenseConfig(rInt, "denied"),
 				Check: resource.ComposeTestCheckFunc(
 					//Even though the API accepts "denied", it still returns "blacklisted" until 15.0
```
armsnyder, line 39: Old comment can be deleted

internal/provider/resource_gitlab_topic_test.go
```diff
@@ -256,6 +258,46 @@ func TestAccGitlabTopic_softDestroy(t *testing.T) {
...
+		Steps: []resource.TestStep{
+			{
+				SkipFunc: isGitLabVersionAtLeast(context.Background(), testGitlabClient, "15.0"),
+				Config: fmt.Sprintf(`
+					resource "gitlab_topic" "this" {
+						name = "foo-%d"
+						title = "Foo-%d"
+					}
+				`, rInt, rInt),
+				ExpectError: regexp.MustCompile(`title is not supported by your version of GitLab. At least GitLab 15.0 is required`),
```
armsnyder, line 276: [regexp.QuoteMeta](https://pkg.go.dev/regexp#QuoteMeta) is better for creating a regex that matches a string literal

armsnyder (changes requested):
Thanks for the test matrix change. That's a huge help. Mostly my comments are about error handling. The problem is by ignoring the error we could get unexpected test results. You can pass the `testing.T` argument and call `t.Fail` if the error is not nil.

internal/provider/resource_gitlab_topic_test.go
timofurrer: `regexp.QuoteMeta` returns a string, not a regex. Since this code is copied, I think I'll leave it as-is. Are you okay with that?

internal/provider/resource_gitlab_topic_test.go
timofurrer: It's because it's a required attribute in 15.0 and we'd need to duplicate the entire test which has quite some test steps. I don't think it's worth that in this case.

Regarding using the in-line configs: I agree, it should be, but let's do this in a separate PR and focus here on getting 15.0 support :)

internal/provider/data_source_gitlab_repository_file_test.go
timofurrer: Yes, makes sense in this case!

internal/provider/resource_gitlab_topic_test.go
timofurrer: Actually, I cherry-picked it from my own code from https://github.com/gitlabhq/terraform-provider-gitlab/pull/1048. Sorry 🙈 

Still, I think it's easier to use `MustCompile` here, or should I first do the `QuoteMeta` and then turn into a regex?

timofurrer:
@PatrickRice-KSC @armsnyder I've addressed your feedback. Could you please give it another round? 🏓

internal/provider/resource_gitlab_topic_test.go
PatrickRice-KSC: Works for me :)

PatrickRice-KSC (approved):
Looks good from my changes. Thanks @timofurrer !

internal/provider/resource_gitlab_topic_test.go
armsnyder: In your case I don't think it's a big deal. The `.` characters technically mean "any character" in your regex, but realistically it doesn't matter. I was thinking back to other test code, where the error message sometimes has wackier characters, as in: https://github.com/gitlabhq/terraform-provider-gitlab/blob/89b7cd6e969e4efe4e748859bf48edfe980d9d96/internal/provider/resource_gitlab_project_variable_test.go#L168-L170

armsnyder (approved):
🚢

internal/provider/resource_gitlab_topic_test.go
timofurrer: Yes, given that it's "just test code" I was fine with it :)

PR #1098, opened 2022-05-24
Initial addition of error tracking settings resource

### LINKED ISSUE #457: Implement error_tracking API
There is currently no support for managing the Sentry integration via the `error_tracking` API https://docs.gitlab.com/ee/api/error_tracking.html. Seems pretty easy to add, though.

Shocktrooper:
Closes #457


Shocktrooper:
So much to my dismay it appears I was wrong about even the integrated error tracking working via the API.  It appears that the resources that come with this PR can not be used at all unless Error Tracking is turned on for the project in the UI (Applies to both integrated and separate Sentry backend).  This in my opinion makes this API even outside of terraform nigh unusable.  I will finish the work and hopefully GitLab upstream can allow Error Tracking to be enabled via the UI so there is no prerequisite whatsoever of UI interaction before the API can be used.

Issue #439, opened 2020-09-23
Fix S031: schema should not only enable Computed and configure MaxItems


armsnyder:
Check: [S031](https://github.com/bflad/tfproviderlint/tree/master/passes/S031)

```
./gitlab/data_source_gitlab_project.go:102:18: S031: schema should not only enable Computed and configure MaxItems
./gitlab/data_source_gitlab_projects.go:344:16: S031: schema should not only enable Computed and configure MaxItems
./gitlab/data_source_gitlab_projects.go:441:20: S031: schema should not only enable Computed and configure MaxItems
./gitlab/data_source_gitlab_projects.go:478:22: S031: schema should not only enable Computed and configure MaxItems
./gitlab/data_source_gitlab_projects.go:550:30: S031: schema should not only enable Computed and configure MaxItems
```

We should fix these reported issues, and then remove this check from the list of ignored checks in the gitlabproviderlint make target.

To reproduce, run `go run ./gitlabproviderlint ./gitlab` (pending #432)

Issue #438, opened 2020-09-23
Fix S024: ForceNew is extraneous in data source schema attributes


armsnyder:
Check: [S024](https://github.com/bflad/tfproviderlint/tree/master/passes/S024)

```
./gitlab/data_source_gitlab_projects.go:181:5: S024: ForceNew is extraneous in data source schema attributes
```

We should fix these reported issues, and then remove this check from the list of ignored checks in the gitlabproviderlint make target.

To reproduce, run `go run ./gitlabproviderlint ./gitlab` (pending #432)

Issue #434, opened 2020-09-23
Fix AT002: acceptance test function name should not include import


armsnyder:
Check: [AT002](https://github.com/bflad/tfproviderlint/tree/master/passes/AT002)

```
./gitlab/resource_gitlab_deploy_key_test.go:76:6: AT002: acceptance test function name should not include import
./gitlab/resource_gitlab_group_cluster_test.go:101:6: AT002: acceptance test function name should not include import
./gitlab/resource_gitlab_group_label_test.go:87:6: AT002: acceptance test function name should not include import
./gitlab/resource_gitlab_group_test.go:84:6: AT002: acceptance test function name should not include import
./gitlab/resource_gitlab_instance_cluster_test.go:103:6: AT002: acceptance test function name should not include import
./gitlab/resource_gitlab_project_cluster_test.go:103:6: AT002: acceptance test function name should not include import
./gitlab/resource_gitlab_project_level_mr_approvals_test.go:66:6: AT002: acceptance test function name should not include import
./gitlab/resource_gitlab_project_mirror_test.go:68:6: AT002: acceptance test function name should not include import
./gitlab/resource_gitlab_project_test.go:364:6: AT002: acceptance test function name should not include import
./gitlab/resource_gitlab_project_test.go:453:6: AT002: acceptance test function name should not include import
./gitlab/resource_gitlab_service_github_test.go:57:6: AT002: acceptance test function name should not include import
./gitlab/resource_gitlab_service_jira_test.go:67:6: AT002: acceptance test function name should not include import
./gitlab/resource_gitlab_service_pipelines_email_test.go:59:6: AT002: acceptance test function name should not include import
./gitlab/resource_gitlab_service_slack_test.go:77:6: AT002: acceptance test function name should not include import
```

We should fix these reported issues, and then remove this check from the list of ignored checks in the gitlabproviderlint make target.

To reproduce, run `go run ./gitlabproviderlint ./gitlab` (pending #432)

PR #1105, opened 2022-05-29
Remove a few technical debts regarding issues detected by the linter

timofurrer:
This change set addresses #434, #438 and #439 - which are all related to issues detected by the linter and specifically ignored by us in the past.

### DIFF
--- a/internal/provider/resource_gitlab_project_level_mr_approvals_test.go
+++ b/internal/provider/resource_gitlab_project_level_mr_approvals_test.go
@@ -7,24 +7,32 @@ import (
 	"fmt"
 	"testing"
 
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 	gitlab "github.com/xanzy/go-gitlab"
 )
 
 func TestAccGitlabProjectLevelMRApprovals_basic(t *testing.T) {
+	testAccCheckEE(t)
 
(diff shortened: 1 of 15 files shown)


internal/provider/resource_gitlab_project_level_mr_approvals_test.go
```diff
@@ -14,6 +14,7 @@ import (
 )
 
 func TestAccGitlabProjectLevelMRApprovals_basic(t *testing.T) {
+	testAccCheckEE(t)
```
PatrickRice-KSC, line 17: @timofurrer  - very much a nitpick, but since we're not creating any resources via the API before `resource.Test`, it feels like this should be moved into the PreCheck function instead. WDYT?

internal/provider/resource_gitlab_service_github_test.go
```diff
@@ -14,6 +14,8 @@ import (
 )
 
 func TestAccGitlabServiceGithub_basic(t *testing.T) {
+	testAccCheckEE(t)
```
PatrickRice-KSC, line 17: Same comment re:Precheck as above.

PatrickRice-KSC (commented):
Very minor nitpicks.

internal/provider/resource_gitlab_project_level_mr_approvals_test.go
timofurrer: Yes, you are right. I've refactored the test into creating the project out of the tf config :)

internal/provider/resource_gitlab_service_github_test.go
timofurrer: same here :)

PatrickRice-KSC (approved):
LGTM!

PR #1106, opened 2022-05-29
Suppress diff on the skip_confirmation attribute of gitlab_user

### LINKED ISSUE #1101: gitlab_user resource causes destroy/add of imported resource, because of the `skip_confirmation` attribute
### GitLab Provider version

3.14.0

### GitLab version

Gitlab EE 4.5

### Terraform version

Terraform v1.0.11

```hcl
resource "gitlab_user" "example_user" {
  name             = "Example User"
  username         = "exampleuser"
  email            = "user@example.com"
  is_admin         = true
  projects_limit   = 0
  can_create_group = false
  is_external      = false
  note             = "Ipsum Lorem."
}
```

### Relevant log output

```shell
# gitlab_user.example_user must be replaced
-/+ resource "gitlab_user" "example_user" {
      ~ id                = "4" -> (known after apply)
        name              = "Example User"
      ~ namespace_id      = 0 -> (known after apply)
      + skip_confirmation = true # forces replacement
        # (8 unchanged attributes hidden)
    }
```

### Description

I imported the user defined above (details differ due to sensitive information in actual user) using `terraform import gitlab.example_user 4`. I then run `terraform plan` and get the log output, which wants to recreate the user due to the `skip_confirmation` field. Setting this field to `true` doesn't change the behavior. I assume it is related to #491 ?

The `skip_confirmation` field shouldn't force a recreate.

(As an aside, it appears the import is incorrect. The `namespace_id` in the plan output says `0`, and I confirmed the state file had the same data. However, the `namespace_id` for the existing user was actually a different number.)

PatrickRice-KSC:
## Description

Fix an issue with gitlab_user where the `skip_confirmation` attribute results in a non-empty plan on every run.

Closes #1101

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/internal/provider/resource_gitlab_user.go
+++ b/internal/provider/resource_gitlab_user.go
@@ -79,6 +79,10 @@ var _ = registerResource("gitlab_user", func() *schema.Resource {
 				Optional:    true,
 				Default:     true,
 				ForceNew:    true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					// Skip Confirmation doesn't come back from the API, so this will always return a diff unless this skips.
+					return true
+				},
 			},
 			"projects_limit": {
 				Description: "Integer, defaults to 0.  Number of projects user can create.",
--- a/internal/provider/resource_gitlab_user.go
+++ b/internal/provider/resource_gitlab_user.go
@@ -74,15 +74,15 @@ var _ = registerResource("gitlab_user", func() *schema.Resource {
 				Default:     false,
 			},
 			"skip_confirmation": {
-				Description: "Boolean, defaults to true. Whether to skip confirmation.",
-				Type:        schema.TypeBool,
-				Optional:    true,
-				Default:     true,
-				ForceNew:    true,
-				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
-					// Skip Confirmation doesn't come back from the API, so this will always return a diff unless this skips.
-					return true
-				},
+				Description: `Boolean, defaults to true. Whether to skip confirmation.
+				
+This field does not import properly, and GitLab does not return the value of this field to verify it was set properly on apply.
+Terraform assumes the field was set properly upon first apply, and will update the field unless other values change.
+				`,
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
 			},
 			"projects_limit": {
 				Description: "Integer, defaults to 0.  Number of projects user can create.",
@@ -135,6 +135,7 @@ func resourceGitlabUserSetToState(d *schema.ResourceData, user *gitlab.User) {
 	d.Set("note", user.Note)
 	d.Set("state", user.State)
 	d.Set("namespace_id", user.NamespaceID)
+
 }
 
 func resourceGitlabUserCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
@@ -179,6 +180,11 @@ func resourceGitlabUserCreate(ctx context.Context, d *schema.ResourceData, meta 
 			return diag.FromErr(err)
 		}
 	}
+
+	// skip_confirmation doesn't come back on the read operation (or on any user operation),
+	// so assume it was set properly, and set it into state. This is an issue with GitLab.
+	// see https://github.com/gitlabhq/terraform-provider-gitlab/pull/1106/
+	d.Set("skip_confirmation", d.Get("skip_confirmation").(bool))
 
 	return resourceGitlabUserRead(ctx, d, meta)
 }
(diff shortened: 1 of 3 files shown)


PatrickRice-KSC:
@timofurrer  - I'm not really proficient in ruby, so I can't speak to how possible it is to expose the API via gitlab 🙂 . 

This is a problem with more than just import though, it'll happen any time a value is set into that field in a config, because we never set the field into state anywhere (likely because it's not returned from any of the APIs). You can see that code here: https://github.com/gitlabhq/terraform-provider-gitlab/blob/main/internal/provider/resource_gitlab_user.go#L123

We could potentially set it into state in create/update and assert that the value got set appropriately during insert/update, but that seems a bit of an anti-pattern since we've seen instances before where the GitLab API returned 200 and didn't actually set the value, so we wouldn't want to give the impression that an operation succeeded when it didn't.  I debated seeing if we could use `getChange`, but that compares state to value, which means since the state is always nil it still comes back with a change. I also debated whether there was something more elegant than just `return true` for the skip func, but it's all comparing to existing state which would be nil  :-/

timofurrer:
> This is a problem with more than just import though, it'll happen any time a value is set into that field in a config, because we never set the field into state anywhere (likely because it's not returned from any of the APIs).

> We could potentially set it into state in create/update and assert that the value got set appropriately during insert/update,

Yes, I guess we meant the same here (I was just assume that and wasn't specific enough, sorry) ... basically: 

It would require that we set the `skip_confirmation` attribute from the config to state during create / update - similar to what we do with a `token`. In the `update` we would just set the value from the config every time (because we don't know if there was a drift).

Then, there is still the import ... for which we may just use what's in the config ...

I'll dig into GitLab once I have some time.

olhado:
Just poking this so it stays alive. The lifecycle `ignore_changes` workaround is not great.

PatrickRice-KSC:
@timofurrer  - Should we just update this to set the state on create/update, and skip import verification for now? It would require a separate attribute to change in order to trigger an "update" operation because we'd still need the skip func, but at least it's slightly more graceful.

timofurrer:
@PatrickRice-KSC I still didn't have any time to dig into the GitLab side of things ... 

> Should we just update this to set the state on create/update, and skip import verification for now?

Yes, I'd skip import for now ..

PatrickRice-KSC:
@timofurrer  - I updated the way this is handled to be set directly from the config on create/update, removed the suppress function and set the attribute as computed. 

I also updated the test to check the value directly since the attribute doesn't exist on the *gitlab.User object (since it can't be retrieved from any API).

Shocktrooper:
@PatrickRice-KSC Is this good for a review?

internal/provider/resource_gitlab_user.go
```diff
@@ -74,11 +74,15 @@ var _ = registerResource("gitlab_user", func() *schema.Resource {
 				Default:     false,
 			},
 			"skip_confirmation": {
-				Description: "Boolean, defaults to true. Whether to skip confirmation.",
-				Type:        schema.TypeBool,
-				Optional:    true,
-				Default:     true,
-				ForceNew:    true,
+				Description: `Boolean, defaults to true. Whether to skip confirmation.
+
```
timofurrer, line 78: The rendering of this is not really nice:

image

Do you mind removing the blank line. It would change it to:

image

```suggestion
```

timofurrer (approved):
@PatrickRice-KSC 

IHMO, we could actually merge it given it provides some value for some people already 😊 

However, what would you think about just suppressing any change on this field?

Like we do here with the `skip_wait_for_default_branch_protection` attribute:

https://github.com/gitlabhq/terraform-provider-gitlab/blob/2aa1c94e80d23d2e668330ea7b0e9402cc2edc09/internal/provider/resource_gitlab_project.go#L683

internal/provider/resource_gitlab_user.go
Shocktrooper: @PatrickRice-KSC

PatrickRice-KSC: Wow, I'm not sure how I missed this comment for over a month. Jeez, sorry guys (and thanks for the reminder @Shocktrooper !)

I'll get those changes made so we can get this merged 😄

Shocktrooper: No worries 😄 I was just like huh I don't think there's anything else to do besides this review comment

PatrickRice-KSC:
> IHMO, we could actually merge it given it provides some value for some people already 😊
> 
> However, what would you think about just suppressing any change on this field?
> 
> Like we do here with the `skip_wait_for_default_branch_protection` attribute:

Hey @timofurrer  - I'm working through fixing the test failures, and when we set the DiffSuppressFunc to just return `true`, the change no longer works how we want it to, because it suppresses the initial _creation_ diff as well, and the value _always returns false_. So I think we need to leave the DiffSuppressFunc off, and just let it set the value into state on the first apply. Thoughts?

PR #1109, opened 2022-06-01
WIP: GraphQL Helpers and Adding current_user data source

### LINKED ISSUE #1117: Add a data_source for determining the current_user
### Feature Description

We'd talked during the community hour about their having been a historical request for the ability to pull the current user via a data source. I couldn't find the issue, so I'm creating a new one. 

The issue will add a new data_source that will allow a user to determine the current_user, as specified by the input API token to the provider.

Expected usage:
```
data "gitlab_current_user" "current" {}
```

PatrickRice-KSC:
## Description

This PR starts to implement helper utilities for using GraphQL Methods, and is being open very early (before any of the project resources are really ready _at all_) so that we can collaborate on the POC methods and approach.

Closes #1117 

    - Create the resource
    - Update the attributes
    - Import the resource


PatrickRice-KSC:
@armsnyder  | @timofurrer  - Ignore the changes to the project resource for now (which are just scaffolding for actually applying the limiting of the job token), and check out the graphql helper and test. This is essentially a bare-bones pattern for using GraphQL for calls. It essentially involves a couple of things:

- I'm using the existing go-gitlab client to create a new request instead of building it from a base http/net request. This allows us to re-use authentication and headers we create with the provider. Since the go-gitlab client sets the `/api/v4`, this is overwritten in the helper method.
- There is a generic `GraphQLQuery` struct that just accepts one "query" parameter, which is the graphQL query that gets sent to the server. This allows us to pass that in directly within "NewRequest", which expects a json object to marshal. I'd tried to make that easier and just accept a query string, but that gets a lot messier within the helper, and this is significantly easier to read and maintain IMO.
- Within the helper_test, you can see I'm defining a custom response struct to parse out the data. In many cases, we can probably wrap the go-gitlab classes with a "data" parent to unwrap the properties, but I do expect that we'll have a lot more custom response structs since each terraform resource may have different attributes and structures that it's unwrapping to get the data it wants.

This feels like a pretty basic MVP approach for GraphQL here. The queries are not at all typesafe, but I'm not sure how we could easily make that work. The _responses_ are type safe since we're defining structs for them.

Very interested in feedback and thoughts! I'll keep working on the project side as well, so we'll have another implementation to look at, but wanted to get this PR open so you guys could check it out as we went.

PatrickRice-KSC:
I'm just cleaning up git history a bit, so closing this PR and opening a new one.

Issue #1110, opened 2022-06-01
invalid memory address or nil pointer dereference


lucymh:
### GitLab Provider version

3.15.0

### GitLab version

GitLab Enterprise Edition 15.1.0-pre

### Terraform version

1.0.7

### Relevant log output

```shell
Stack trace from the terraform-provider-gitlab_v3.15.0 plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xb8 pc=0xb6a0a5]

goroutine 5879 [running]:
github.com/gitlabhq/terraform-provider-gitlab/internal/provider.resourceGitlabServiceMicrosoftTeamsRead({0xee57a8, 0xc000471800}, 0xc000099580, {0xd2f5c0, 0xc0002d9000})
	github.com/gitlabhq/terraform-provider-gitlab/internal/provider/resource_gitlab_service_microsoft_teams.go:149 +0x2c5
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).read(0xc0003a6620, {0xee57e0, 0xc00068cfc0}, 0xd, {0xd2f5c0, 0xc0002d9000})
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.16.0/helper/schema/resource.go:724 +0x12e
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).RefreshWithoutUpgrade(0xc0003a6620, {0xee57e0, 0xc00068cfc0}, 0xc00017c1a0, {0xd2f5c0, 0xc0002d9000})
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.16.0/helper/schema/resource.go:1015 +0x585
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*GRPCProviderServer).ReadResource(0xc00000d278, {0xee5738, 0xc0004eb280}, 0xc0004eb300)
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.16.0/helper/schema/grpc_provider.go:613 +0x574
github.com/hashicorp/terraform-plugin-go/tfprotov5/tf5server.(*server).ReadResource(0xc0000fa460, {0xee57e0, 0xc00068ca20}, 0xc0004710e0)
	github.com/hashicorp/terraform-plugin-go@v0.9.0/tfprotov5/tf5server/server.go:746 +0x48c
[...]

PR #1113, opened 2022-06-01
Fix handling removal from state of microsoft teams integration resource

timofurrer:
This resulted in a nil pointer panic, because when a project wasn't found `p` would obv. be `nil`. 

Closes: #1110


PatrickRice-KSC (approved):
Makes sense, thanks @timofurrer !

PR #1118, opened 2022-06-05
Add current_user data source

### LINKED ISSUE #1117: Add a data_source for determining the current_user

PatrickRice-KSC:
## Description

Closes #1117

Successor of #1109

    - Create the resource
    - Update the attributes
    - Import the resource


internal/provider/data_source_gitlab_current_user.go
```diff
@@ -0,0 +1,101 @@
...
+	query := GraphQLQuery{
+		`query {currentUser {name, bot, groupCount, id, namespace{id} publicEmail, username}}`,
+	}
+	log.Printf("[DEBUG] executing GraphQL Query %s to retrieve current user", query.Query)
+
+	var response CurrentUserResponse
+	_, err := SendGraphQLRequest(context.Background(), client, query, &response)
+	if err != nil {
+		return diag.FromErr(err)
+	}
```
timofurrer, line 71:
```suggestion
	if _, err := SendGraphQLRequest(ctx, client, query, &response); err != nil {
		return diag.FromErr(err)
	}
```

internal/provider/data_source_gitlab_current_user.go
```diff
@@ -0,0 +1,101 @@
...
+	Data struct {
+		CurrentUser GraphQLUser `json:"currentUser"`
+	} `json:"data"`
+}
+
+type GraphQLUser struct {
+	Name       string `json:"name"`
+	Bot        bool   `json:"bot"`
+	GroupCount int    `json:"groupCount"`
+	ID         string `json:"id"` // This is purposefully a string, as in some APIs it comes back as a globally unique ID
```
timofurrer, line 95: Out of curiosity: what's a `globally unique ID` and how would the non-integer representation look like?

internal/provider/data_source_gitlab_current_user_test.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/xanzy/go-gitlab"
+)
+
+func TestAccDataSourceGitlabCurrentUser_basic(t *testing.T) {
+	//The root user has no public email by default, set the public email so it shows up properly.
+	_, _, _ = testGitlabClient.Users.ModifyUser(1, &gitlab.ModifyUserOptions{
```
timofurrer, line 15: I suggest that we add a `t.Cleanup` function to revert that back - especially for when we start to parallelise test execution.

internal/provider/data_source_gitlab_current_user_test.go
```diff
@@ -0,0 +1,38 @@
...
+			{
+				Config: `data "gitlab_current_user" "this" {}`,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet("data.gitlab_current_user.this", "id"),
+					resource.TestCheckResourceAttrSet("data.gitlab_current_user.this", "name"),
+					resource.TestCheckResourceAttrSet("data.gitlab_current_user.this", "username"),
+					resource.TestCheckResourceAttrSet("data.gitlab_current_user.this", "bot"),
+					resource.TestCheckResourceAttrSet("data.gitlab_current_user.this", "group_count"),
+					resource.TestCheckResourceAttrSet("data.gitlab_current_user.this", "namespace_id"),
+					resource.TestCheckResourceAttrSet("data.gitlab_current_user.this", "public_email"),
```
timofurrer, line 33: We actually know the values and I think it makes sense to assert for them here. WDYT?

internal/provider/graphql_helper.go
```diff
@@ -0,0 +1,30 @@
...
+		return nil, err
+	}
+
+	_, err = client.Do(request, objectToParseForResponse)
+	if err != nil {
+		return nil, err
+	}
+
+	return objectToParseForResponse, nil
+}
```
timofurrer, line 25: I did a small refactoring. WDYT?

```suggestion
func SendGraphQLRequest(ctx context.Context, client *gitlab.Client, query GraphQLQuery, response interface{}) (interface{}, error) {
	request, err := client.NewRequest("POST", "", query, nil)
	if err != nil {
		return nil, err
	}
	// Overwrite the path of the existing request, as otherwise the go-gitlab client appends /api/v4 instead.
	request.URL.Path = "/api/graphql"

	if _, err = client.Do(request, response); err != nil {
		return nil, err
	}

	return response, nil
}
```

internal/provider/graphql_helper.go
```diff
@@ -0,0 +1,30 @@
...
+
+// Represents a GraphQL call to the API. All graphQL calls are a string passed to the "query" parameter, so they should be included here.
```
timofurrer, line 27: NIT:

```suggestion
// Represents a GraphQL call to the API. All GraphQL calls are a string passed to the "query" parameter, so they should be included here.
```

timofurrer (changes requested):
Pretty solid already! I've left a few suggestions :) 🏓

internal/provider/data_source_gitlab_current_user_test.go
PatrickRice-KSC: I'm good with it. I didn't assert them because they're really easy to change as part of the startup scripts. That's why in the helper test I only checked username even though I obtain more than that.

internal/provider/data_source_gitlab_current_user_test.go
PatrickRice-KSC: @timofurrer - Makes sense - I'll get that added. It's worth noting that "admin@example.com" is the _only_ value we could set here based on our current setup, so any other test would be setting it to exactly the same value

PatrickRice-KSC:
@timofurrer  - Should be good for a re-review here 🏓

internal/provider/data_source_gitlab_current_user.go
timofurrer: very interesting! 

> I do suspect that in some cases since we have a mixture of GraphQL and REST we may have to post-process out the unique ID and store it in a separate spot for folks to use in downstream resources.

That's exactly my doubt here - we would be able to e.g. use the `namespace_id` field to create a project for a specific user ...

I need to do a little bit of research about those ids and the format. Let's see if for now and to consistently allow interoperability we could just use the last part of the `globally unique id` when splitting by `/`. 
WDYT?

timofurrer: @PatrickRice-KSC I've added some logic to make this data source more interoperable with the other data sources and resources by separating the `id` from the `global_id`.

From what I could gather always the part after the `/` is an integer id mapping the normal internal id we are used to from the REST API. 

WDYT? Please feel free to change it in the way you like it!

PatrickRice-KSC: @timofurrer  - Makes complete sense to me, and I agree with keeping the base attribute as the integer and pre-pending the GraphQL ID with "global". I suspect that's something we should standardize across GraphQL resources (because almost every ID that comes back via GraphQL is in the global format so far that I've seen).

timofurrer (approved):
LGTM 🎉

Issue #1099, opened 2022-05-25
expires_at on gitlab_project_membership


bbr32:
### Feature Description

Hi,

It would be nice to add "expires_at" field on [gitlab_project_membership](https://registry.terraform.io/providers/gitlabhq/gitlab/latest/docs/resources/project_membership) resource, like so :
```hcl
resource "gitlab_project_membership" "example" {
  project_id   = "67890"
  user_id      = 1234
  access_level = "guest"
  expires_at   = "2099-01-01" # Set expiration date for project user membership
}
```
This field is already present on the [Gitlab API resource to add a member to project](https://docs.gitlab.com/ee/api/members.html#add-a-member-to-a-group-or-project). And has already been implemented in the provider with the [group_share_group](https://registry.terraform.io/providers/gitlabhq/gitlab/latest/docs/resources/group_share_group) resource, should be quick to implement.

### Do you want to implement this?

- [X] I would like to implement this myself 👷

timofurrer:
Sounds like a great addition! Since you marked that you want to implement it yourself we are very much looking forward to review a PR 🎉

bbr32:
Hi, I've made the code changes, but I am having a 403 when pushing my branch to the repo (tried with PAT, SSH key, disabling 2FA... works fine on other public repos).
Is there a specific setup to push a branch and open a PR or is this a issue on my side ? 
Thanks in advance !

timofurrer:
Awesome! Alright, you can't push directly to this repository, because that requires you to be a maintainer. However, you can push the branch to your own fork and easily create a PR to this repo from there - nothing special is required for that :)

bbr32:
Oh ok thx ! Here is the PR https://github.com/gitlabhq/terraform-provider-gitlab/pull/1122

PR #1122, opened 2022-06-06
Add expiration date for gitlab_project_membership

bbr32:
## Description

This PR closes https://github.com/gitlabhq/terraform-provider-gitlab/issues/1099

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/internal/provider/resource_gitlab_project_membership.go
+++ b/internal/provider/resource_gitlab_project_membership.go
@@ -48,6 +48,12 @@ var _ = registerResource("gitlab_project_membership", func() *schema.Resource {
 				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice(validProjectAccessLevelNames, false)),
 				Required:         true,
 			},
+			"expires_at": {
+				Description:  "Expiration date for the project membership. Format: `YYYY-MM-DD`",
+				Type:         schema.TypeString,
+				ValidateFunc: validateDateFunc,
+				Optional:     true,
+			},
 		},
 	}
 })
@@ -57,11 +63,13 @@ func resourceGitlabProjectMembershipCreate(ctx context.Context, d *schema.Resour
 
 	userId := d.Get("user_id").(int)
 	projectId := d.Get("project_id").(string)
+	expiresAt := d.Get("expires_at").(string)
 	accessLevelId := accessLevelNameToValue[d.Get("access_level").(string)]
 
 	options := &gitlab.AddProjectMemberOptions{
 		UserID:      &userId,
 		AccessLevel: &accessLevelId,
+		ExpiresAt:   &expiresAt,
 	}
 	log.Printf("[DEBUG] create gitlab project membership for %d in %s", options.UserID, projectId)
 
@@ -115,10 +123,12 @@ func resourceGitlabProjectMembershipUpdate(ctx context.Context, d *schema.Resour
 
 	userId := d.Get("user_id").(int)
 	projectId := d.Get("project_id").(string)
+	expiresAt := d.Get("expires_at").(string)
 	accessLevelId := accessLevelNameToValue[strings.ToLower(d.Get("access_level").(string))]
 
 	options := gitlab.EditProjectMemberOptions{
 		AccessLevel: &accessLevelId,
+		ExpiresAt:   &expiresAt,
 	}
 	log.Printf("[DEBUG] update gitlab project membership %v for %s", userId, projectId)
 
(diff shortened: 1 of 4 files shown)


internal/provider/resource_gitlab_project_membership.go
```diff
@@ -48,6 +48,12 @@ var _ = registerResource("gitlab_project_membership", func() *schema.Resource {
 				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice(validProjectAccessLevelNames, false)),
 				Required:         true,
 			},
+			"expires_at": {
+				Description:  "Expiration date for the project membership. Format: `YYYY-MM-DD`",
+				Type:         schema.TypeString, // Format YYYY-MM-DD
```
timofurrer, line 53: NIT: since the `Description` already contains that information close the the type definition I think it's okay to omit that comment:

```suggestion
				Type:         schema.TypeString,
```

internal/provider/resource_gitlab_project_membership.go
```diff
@@ -153,7 +163,9 @@ func resourceGitlabProjectMembershipSetToState(d *schema.ResourceData, projectMe
 	d.Set("project_id", projectId)
 	d.Set("user_id", projectMember.ID)
 	d.Set("access_level", accessLevelValueToName[projectMember.AccessLevel])
-
+	if projectMember.ExpiresAt != nil {
+		d.Set("expires_at", projectMember.ExpiresAt.String())
+	}
```
timofurrer, line 168: If `projectMember.ExpiresAt` comes back as `nil` we have to clear the `expires_at` state field:

```suggestion
	if projectMember.ExpiresAt != nil {
		d.Set("expires_at", projectMember.ExpiresAt.String())
	} else {
	         d.Set("expires_at", "")
	}
```

internal/provider/resource_gitlab_project_membership_test.go
```diff
@@ -35,7 +35,8 @@ func TestAccGitlabProjectMembership_basic(t *testing.T) {
 			{
 				Config: testAccGitlabProjectMembershipUpdateConfig(rInt),
 				Check: resource.ComposeTestCheckFunc(testAccCheckGitlabProjectMembershipExists("gitlab_project_membership.foo", &membership), testAccCheckGitlabProjectMembershipAttributes(&membership, &testAccGitlabProjectMembershipExpectedAttributes{
-					access_level: fmt.Sprintf("guest"), // nolint // TODO: Resolve this golangci-lint issue: S1039: unnecessary use of fmt.Sprintf (gosimple)
+					access_level: fmt.Sprintf("guest"),      // nolint // TODO: Resolve this golangci-lint issue: S1039: unnecessary use of fmt.Sprintf (gosimple)
+					expiresAt:    fmt.Sprintf("2099-01-01"), // nolint // TODO: Resolve this golangci-lint issue: S1039: unnecessary use of fmt.Sprintf (gosimple)
```
timofurrer, line 39: We don't need the `fmt.Sprintf` here:

```suggestion
					access_level: "guest",
					expiresAt:    "2099-01-01",
```

timofurrer (changes requested):
Thanks @bbr32 for the contribution 🤝 it already looks pretty solid!

I've left a few nitpicks and one suggestion to properly reset the state when `expires_at` is removed in an update (not sure if that is even possible with the API ...).

Back to you 🏓

internal/provider/resource_gitlab_project_membership.go
bbr32: thx, I've added your suggestion

internal/provider/resource_gitlab_project_membership_test.go
bbr32: removed that as well

internal/provider/resource_gitlab_project_membership.go
```diff
@@ -153,7 +163,11 @@ func resourceGitlabProjectMembershipSetToState(d *schema.ResourceData, projectMe
 	d.Set("project_id", projectId)
 	d.Set("user_id", projectMember.ID)
 	d.Set("access_level", accessLevelValueToName[projectMember.AccessLevel])
-
+	if projectMember.ExpiresAt != nil {
+		d.Set("expires_at", projectMember.ExpiresAt.String())
+	} else {
+	         d.Set("expires_at", "")
```
timofurrer, line 169: Indentation doesn't match [here](https://github.com/gitlabhq/terraform-provider-gitlab/runs/6807171411?check_suite_focus=true) - Unfortunately, that came from my suggestion for which I didn't use proper tabs. Sorry, for that.

```suggestion
		d.Set("expires_at", "")
```

timofurrer (changes requested):
@bbr32 

> I've applied the changes, to be honest

Yes, I'm aware - thanks for going through it anyways!

> shoud I apply your changes to that code as well ?

Yes, that's always appreciated. However, in favor of keeping PRs small I suggest to address that in another PR.

I've left one small suggestion for something my suggestion broke.

internal/provider/resource_gitlab_project_membership.go
bbr32: my bad, I forgot to run the `make reviewable` command :'(
Ok I will apply your suggestions in another PR then !

PR #1126, opened 2022-06-08
Start parallelizing most of the acceptance tests

timofurrer:
The only ones which are not yet `resource.ParallelTest`s are the deploy key tests
which reuse keys and some instance-based resources / data sources.

It brings down the CI testing time a couple of minutes, but not too drastically. This is mainly because the Ci machines "only" provide 2 CPU cores which is then used by `go test`. Manually increasing it for Ci led to a lot of issues somehow and I wasn't able to debug it yet.

The better news is that on my local machine testing time is down to ~6 minutes, which more than half in my case.

My suggestion would be to merge this for now and improve on follow ups. 

***

Another problem is that because we perform setups outside of the `resource.ParallelTest` it's not parallelized and takes up a lot of time.

Therefore, I've created https://github.com/hashicorp/terraform-plugin-sdk/issues/979 - maybe you can also chime in there if you have any good ideas :)

### DIFF
--- a/internal/provider/data_source_gitlab_project_issues_test.go
+++ b/internal/provider/data_source_gitlab_project_issues_test.go
@@ -14,7 +14,7 @@ func TestAccDataSourceGitlabProjectIssues_basic(t *testing.T) {
 	testProject := testAccCreateProject(t)
 	testIssues := testAccCreateProjectIssues(t, testProject.ID, 25)
 
-	resource.Test(t, resource.TestCase{
+	resource.ParallelTest(t, resource.TestCase{
 		ProviderFactories: providerFactories,
 		Steps: []resource.TestStep{
 			{
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,6 +4,6 @@
         "GITLAB_TOKEN": "ACCTEST1234567890123",
         "GITLAB_BASE_URL": "http://127.0.0.1:8080"
     },
-    "go.testFlags": ["-count=1", "-v", "-tags=acceptance"],
+    "go.testFlags": ["-count=1", "-v", "-tags=acceptance", "-parallel=2"],
     "go.buildFlags": ["-tags=acceptance"]
 }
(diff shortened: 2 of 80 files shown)


internal/provider/data_source_gitlab_project_issues_test.go
```diff
@@ -14,7 +14,7 @@ func TestAccDataSourceGitlabProjectIssues_basic(t *testing.T) {
 	testProject := testAccCreateProject(t)
 	testIssues := testAccCreateProjectIssues(t, testProject.ID, 25)
```
PatrickRice-KSC, line 15: Note for other reviewers (if any choose to check this out 😄  )- even though CreatProjectIssues doesn't clean up after itself, the issues will be deleted when the project does.

PatrickRice-KSC (changes requested):
I think we need to add a config here to limit the gitpod environment to 2 parallel tests. It looks like it defaults to 5 parallel tests, and the GitLab environment spun up in GitPod gets a lot of deadlock failures on tests when it runs 5 at once :-/

I dropped tests one by one until I got down to 2 and was getting pretty reliably success rates. It was pretty reliable at 3 also, but I was still hitting occasional failures.

timofurrer:
@PatrickRice-KSC The GitPod setup is a good point. We should be able to simply set `-parallel 2` when running the tests or maybe more "out-of-the-box" set `GOMAXPROCS=2` for the GitPod environment?

PatrickRice-KSC:
> We should be able to simply set -parallel 2 when running the tests or maybe more "out-of-the-box" set GOMAXPROCS=2 for the GitPod environment?

@timofurrer  - Yeah, the `-parallel=2` flag was exactly what I was using to test in gitpod, and we can just add it to the test args in the vscode settings where we already have stuff like the GITLAB_TOKEN and such 🙂

timofurrer:
> and we can just add it to the test args in the vscode settings 

@PatrickRice-KSC okay, this may be good enough for now. Even though people use vscode outside of GitPod - it's a good start for now :)

PatrickRice-KSC (approved):
LGTM! Let's ship it!

PR #1129, opened 2022-06-10
resource/gitlab_project: Only execute custom `default_branch` logic for GitLab < 14.10

timofurrer:
GitLab 14.10 released with a bugfix for the `default_branch` field in
the Projects REST API so that our custom logic is no longer required.

see https://gitlab.com/gitlab-org/gitlab/-/issues/333426

Once GitLab 15.1 is out I suggest to remove that code entirely.


PatrickRice-KSC (approved):
LGTM, thanks @timofurrer !

PR #1131, opened 2022-06-13
Added `gitlab_release_link` as a resource and data source

Moon1706:
## Description
Add resource for creation and management Gitlab release links (https://github.com/gitlabhq/terraform-provider-gitlab/issues/1035).

    - Create the resource
    - Update the attributes
    - Import the resource


Moon1706:
@timofurrer 
Review please. And check pr-acceptance-ee job.

internal/provider/data_source_gitlab_release_links.go
```diff
@@ -0,0 +1,82 @@
...
+			PerPage: 20,
+			Page:    1,
+		})
+
+	var releaseLinks []*gitlab.ReleaseLink
+	for options.Page != 0 {
+		paginatedReleaseLinks, resp, err := client.ReleaseLinks.ListReleaseLinks(project, tagName, &options, gitlab.WithContext(ctx))
+		if err != nil {
+			return diag.FromErr(err)
+		}
```
PatrickRice-KSC, line 61: More a question for @timofurrer - Right now most of our "plural" data sources result in a 404 if they return 0 results due to how the underlying APIs work.. This can cause an issue in configurations where it's perfectly legit to have 0 results for a value. 

Is this really how we want these data sources to work, or should we catch 404 here using the `is404` function and return and break the loop which will just result in an empty list. The empty list will still cause an error in terraform configuration that _expect_ a value (because the downstream resources would break), but would still function perfectly fine in other scenarios where you may be just using the values as a loop in a `count` or `foreach` meta configuration.

Most of the other major cloud provider configurations are... inconsistent in this regard as well, with some data sources erroring, and some returning an empty array.

internal/provider/schema_gitlab_release_link.go
```diff
@@ -0,0 +1,85 @@
...
+			Type:        schema.TypeString,
+			Required:    true,
+		},
+		"url": {
+			Description: "The URL of the link. Link URLs must be unique within the release.",
+			Type:        schema.TypeString,
+			Required:    true,
+		},
+		"filepath": {
+			Description: "Relatively path for a [Direct Asset link](https://docs.gitlab.com/ee/user/project/releases/index.html#permanent-links-to-release-assets).",
```
PatrickRice-KSC, line 38: Slight grammar correction 🙂 
```suggestion
			Description: "Relative path for a [Direct Asset link](https://docs.gitlab.com/ee/user/project/releases/index.html#permanent-links-to-release-assets).",
```

internal/provider/schema_gitlab_release_link.go
```diff
@@ -0,0 +1,85 @@
...
+			Type:        schema.TypeString,
+			Optional:    true,
+		},
+		"link_type": {
+			Description:      fmt.Sprintf("The type of the link: %s. Defaults to %s.", renderValueListForDocs(validLinkTypes), validLinkTypes[0]),
```
PatrickRice-KSC, line 43: Update slightly to match the verbiage we use for other similar lists 🙂 
```suggestion
			Description:      fmt.Sprintf("The type of the link. Valid values are %s. Defaults to %s.", renderValueListForDocs(validLinkTypes), validLinkTypes[0]),
```

PatrickRice-KSC (changes requested):
@Moon1706  overall looking really good! I've requested a couple smaller changes, 1 to match a PR that was just merged, and 2 just around grammar and styling 🙂 . I left one comment for Timo that won't block merging just around general practices for data sources that we'll just resolve when it's the last comment left.

internal/provider/schema_gitlab_release_link.go
Moon1706: Okay)

internal/provider/data_source_gitlab_release_link_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccDataSourceGitlabReleaseLink_basic(t *testing.T) {
+
+	project := testAccCreateProject(t)
+	releases := testAccCreateReleases(t, project, 2)
+
+	resource.Test(t, resource.TestCase{
```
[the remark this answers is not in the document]
Moon1706: Yea, it works with no problems.

internal/provider/data_source_gitlab_release_links.go
Moon1706: Done. Additionally, I've added a test for this case.

PatrickRice-KSC:
Restarted the test, since the failed test likely is a race condition due to container registry enablement being an async operation. We'll keep an eye on that if we see it elsewhere in PRs.

PatrickRice-KSC:
Looks like we're golden. I'll give @timofurrer  a couple days to review (he tends to look for different things than I do 🙂 ) then we can merge.

PatrickRice-KSC (approved):

internal/provider/data_source_gitlab_release_links.go
```diff
@@ -0,0 +1,85 @@
...
+
+	return &schema.Resource{
+		Description: `The ` + "`gitlab_release_links`" + ` data source allows get details of release links.
+
+**Upstream API**: [GitLab REST API docs](https://docs.gitlab.com/ee/api/releases/links.html)`,
+
+		ReadContext: dataSourceGitlabReleaseLinksRead,
+		Schema: map[string]*schema.Schema{
+			"project": {
+				Description: "The ID or [URL-encoded path of the project](https://docs.gitlab.com/ee/api/index.html#namespaced-path-encoding).",
```
timofurrer, line 22: While the API "only" accepts URL-encoded paths, that's not the case for this provider, because `go-gitlab` will correctly URL-encode it for you. Let's use this instead:

```suggestion
				Description: "The ID or full path to the project.",
```

internal/provider/data_source_gitlab_release_links.go
```diff
@@ -0,0 +1,85 @@
...
+		releaseLinks = append(releaseLinks, paginatedReleaseLinks...)
+		options.Page = resp.NextPage
+	}
+
+	log.Printf("[DEBUG] get list release links project/tagName: %s/%s", project, tagName)
+	d.SetId(buildTwoPartID(&project, &tagName))
+	var err error
+	if err = d.Set("release_links", flattenGitlabReleaseLinks(project, tagName, releaseLinks)); err != nil {
+		return diag.Errorf("Failed to set release links to state: %v", err)
+	}
```
timofurrer, line 75: NIT:

```suggestion
	if err := d.Set("release_links", flattenGitlabReleaseLinks(project, tagName, releaseLinks)); err != nil {
		return diag.Errorf("Failed to set release links to state: %v", err)
	}
```

internal/provider/resource_gitlab_release_link.go
```diff
@@ -0,0 +1,153 @@
...
+import (
+	"context"
+	"fmt"
+	"log"
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	gitlab "github.com/xanzy/go-gitlab"
```
timofurrer, line 12: NIT:

```suggestion
	"github.com/xanzy/go-gitlab"
```

internal/provider/resource_gitlab_release_link.go
```diff
@@ -0,0 +1,153 @@
...
+)
+
+var _ = registerResource("gitlab_release_link", func() *schema.Resource {
+	return &schema.Resource{
+		Description: `The ` + "`gitlab_release_link`" + ` resource allows to manage the lifecycle of a release links.
```
timofurrer, line 17:
```suggestion
		Description: `The ` + "`gitlab_release_link`" + ` resource allows to manage the lifecycle of a release link.
```

internal/provider/resource_gitlab_release_link.go
```diff
@@ -0,0 +1,153 @@
...
+func resourceGitlabReleaseLinkCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*gitlab.Client)
+	project := d.Get("project").(string)
+	tagName := d.Get("tag_name").(string)
+	name := d.Get("name").(string)
+	url := d.Get("url").(string)
+
+	options := &gitlab.CreateReleaseLinkOptions{
+		Name: &name,
+		URL:  &url,
```
timofurrer, line 41: NIT:

```suggestion
		Name: gitlab.String(name),
		URL:  gitlab.String(url),
```

internal/provider/data_source_gitlab_release_links.go
timofurrer: @PatrickRice-KSC @Moon1706 Yes, I agree to your judgement. My only concern where a `404` indeed is an error is if it is returned from any other page than the first, because that should actually be "accessible" - w.r.t. permissions and existence. 

But for the sake of simplicity I think it's "okay" to handle it as you've implemented it right now.

timofurrer (changes requested):
@Moon1706 thanks for the nice work once again 💚 

I've left some feedback with suggestions to apply before a merge (require to re-generate docs, too) and some nitpicks. 

@PatrickRice-KSC thanks for the review! I've left an answer to the question you've asked :)

internal/provider/data_source_gitlab_release_links.go
Moon1706: Done

internal/provider/data_source_gitlab_release_links.go
Moon1706: @timofurrer If you want, I can append an additional section of condition (`if is404(err) && (options.Page == 1)`)

timofurrer: @Moon1706 Yes, let's do that!

timofurrer (approved):
@Moon1706 It's an approve from my side once the additional condition from https://github.com/gitlabhq/terraform-provider-gitlab/pull/1131#discussion_r906648382 has been implemented 🎉 

Good work 🤝

internal/provider/data_source_gitlab_release_links.go
PatrickRice-KSC: Agreed, I like this pattern!

Moon1706: Done

timofurrer (approved):
Thanks @Moon1706 for the change and following through the feedback! 

Good work 🎉 🤝 

LGTM 🚢

PR #1143, opened 2022-06-25
Add official support for 15.1 and drop 14.9

timofurrer:


PatrickRice-KSC (approved):
LGTM - I'll wait until Monday in case @nagyv or @armsnyder  have any additional notes, but I validated against docker-hub that these are the most recent patch versions of each of those minor point releases, so I think we're good to go.

timofurrer:
@PatrickRice-KSC thanks for double checking! However, I'd really like to merge this asap so that the other PRs can test against them already. If Adam or Viktor have some inputs let's put them in another iteration :) 

I hope that's okay with you 🤞 🎉

PatrickRice-KSC:
No problem - just wasn't sure if you'd requested extra reviews above normal for a specific reason :)

Issue #1141, opened 2022-06-24
Add `gitlab_cluster_agent_token` resource


longuniquename:
### Feature Description

The current GitLab provider for Terraform (3.15.1) does not support token creation for cluster agents.

It would be nice to have something like:

```tf
resource "gitlab_cluster_agent_token" "example" {
  agent_id    = "default"
  name        = "token-tf"
  description = "terraform managed token"
}
```

Upstream API to cover: https://docs.gitlab.com/ee/api/cluster_agents.html#create-an-agent-token

### Use case

We'd like to create a k8s cluster with terraform and add it to an existing gitlab project. We already can create cluster agent in the project and deploy gitlab cluster agent to the k8s cluster. The only missing piece is creating the token and passing it to cluster agent helm chart variables.

```tf
resource "helm_release" "gitlab_agent" {
  name             = "gitlab-agent"
  namespace        = "gitlab-agent"
  create_namespace = true
  repository       = "https://charts.gitlab.io"
  chart            = "gitlab-agent"
  version          = "0.6.1"

  set {
    name  = "config.token"
    value = gitlab_cluster_agent_token.example.token
  }
}

```

timofurrer:
Hey @longuniquename - thanks for the feature request! 

This resource was long pending on my side and I've just pushed the new resource in the PR #1147 🎉

PR #1147, opened 2022-06-25
New Resource `gitlab_cluster_agent_token`

timofurrer:
This change set adds support for the `gitlab_cluster_agent_token`
resource, representing the [Agents Token
API](https://docs.gitlab.com/ee/api/cluster_agents.html#create-an-agent-token).

Closes: #1141


internal/provider/resource_gitlab_cluster_agent_token_test.go
```diff
@@ -0,0 +1,165 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					// Verify that computed attributes have been correctly set to state
+					testAccResourceGitlabClusterAgentTokenGet("gitlab_cluster_agent_token.this", &sutClusterAgentToken),
+					resource.TestCheckResourceAttrWith("gitlab_cluster_agent_token.this", "token_id", func(value string) error {
+						expectedValue := fmt.Sprintf("%d", sutClusterAgentToken.ID)
+						if value != expectedValue {
+							return fmt.Errorf("should be equal to %s", expectedValue)
+						}
+						return nil
+					}),
```
PatrickRice-KSC, line 90: @timofurrer  - Hm, would we expect new resource to all test computed values this way? The test below here checking the time formatting makes sense, but this type of a test almost feels redundant to me compared to a unit test on the state function. We're essentially comparing a GET response from the API against a GET response from the API, with the only formatting between them being the state function, aren't we?

PatrickRice-KSC (commented):
Question on the test method, otherwise it looks good to me :)

internal/provider/resource_gitlab_cluster_agent_token_test.go
PatrickRice-KSC: I think it's that last "given a dedicated function... that we can unit test" that I'm reacting to here, which I think is a newer pattern. In the past for computed values we've just made sure that they were set as opposed to matched the value that came back from the API because logically as long as they're set they _should_, given the read method is usually almost a straight passthrough from the API.

I don't really have a _huge_ issue with this, clearly it'll work and the tests will pass. It just feels a bit like it's more code to manage and given we could potentially unit test the centralized function it feels like that may be a better way to do it that would result in less code breakage over time.

timofurrer: @PatrickRice-KSC so to ensure that the attribute is still set and that "response to state" function is actually being called you would still do a `resource.TestCheckResourceAttrSet()` call? What we would miss here is that this "response to state" function is actually properly called (which is a very straight forward thing and will probably never be wrong ...)



> I don't really have a _huge_ issue with this,

Some Pros and Cons to me:

As-is:
1. Pro: fully end-2-end
1. Pro: checks are together with the config (which is nice to reason about the correctness of a resource at first glance as opposed to going through the acceptance test and additionally checking the unit tests for correctness)
1. Pro: it's not really adding runtime to the acceptance test execution
1. Con: hard / impossible to test error conditions from API responses
1. Con: *could have been a unit test* - therefore could have been tested without a GitLab instance

Only Unit Test:

1. Pro: testing is smaller scope and isolation
1. Pro: very easy to test error cases
1. Con: misses bigger scope

Bottom line to me is that we eventually need both. I like (2) of *as-is* very much and I'd like to keep it. But I guess for more complex *response to state* function we'd also like to test error conditions etc and require additional unit tests.

Concrete proposal for this change set: leave as-is.

I'm including this into the testing concept I'm drafting.

PatrickRice-KSC: @timofurrer  - I think that makes sense. I would say that I think one more con the current setup is the verbosity of it making the tests a bit harder to read. I feel like we could solve that with a helper method too for performing this type of a check, since all the checks follow a very standard flow of "set expected value with formatting; equals comparison; return nil". 

I'm going to go ahead and merge this for now, and if/when we add a helper method, we can refactor it next time we touch the code.

Cheers for the back-and-forth!

PatrickRice-KSC (approved):

Issue #1151, opened 2022-06-26
`auto_devops_enabled` in `gitlab_project` ignored during creation of the resource


mskarbek:
### GitLab Provider version

3.15.1

### GitLab version

GitLab Community Edition 15.1.0

### Terraform version

Terraform v1.2.3

### Relevant Terraform Configuration

```hcl
resource "gitlab_project" "minio" {
  name = "minio"
  path = "minio"
  namespace_id = gitlab_group.containers.id
  ci_config_path = ".gitlab-ci.yaml"
  auto_devops_enabled = false
  auto_cancel_pending_pipelines = "enabled"
  build_timeout = 600
  container_registry_enabled = false
  default_branch = "main"
  lfs_enabled = false
  merge_method = "ff"
  packages_enabled = false
  shared_runners_enabled = true
  snippets_enabled =false
  wiki_enabled = false
}
```

### Relevant log output

After first apply:
```shell
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # gitlab_project.minio will be created
  + resource "gitlab_project" "minio" {
      + allow_merge_on_skipped_pipeline                  = false
      + analytics_access_level                           = (known after apply)
      + approvals_before_merge                           = 0
      + archived                                         = false
      + auto_cancel_pending_pipelines                    = "enabled"
      + auto_devops_deploy_strategy                      = (known after apply)
      + auto_devops_enabled                              = false
      + autoclose_referenced_issues                      = (known after apply)
      + build_git_strategy                               = (known after apply)
      + build_timeout                                    = 600
      + builds_access_level                              = (known after apply)
      + ci_config_path                                   = ".gitlab-ci.yaml"
      + ci_forward_deployment_enabled                    = true
      + container_registry_access_level                  = (known after apply)
      + container_registry_enabled                       = false
      + default_branch                                   = "main"
      + forking_access_level                             = (known after apply)
      + http_url_to_repo                                 = (known after apply)
      + id                                               = (known after apply)
      + issues_access_level                              = (known after apply)
      + issues_enabled                                   = true
      + lfs_enabled                                      = false
      + merge_method                                     = "ff"
      + merge_pipelines_enabled                          = false
      + merge_requests_access_level                      = (known after apply)
      + merge_requests_enabled                           = true
      + merge_trains_enabled                             = false
      + mirror                                           = false
[...]

PR #1152, opened 2022-06-27
resource/gitlab_project: Fix passing `false` to API for explicitly set optional attributes

timofurrer:
I think at one point we have to decide on a consistent strategy about how and when to set optional attributes for the API options ... I have a few ideas and I'll write them up in an issue.

Closes: #1151

### DIFF
--- a/internal/provider/resource_gitlab_group.go
+++ b/internal/provider/resource_gitlab_group.go
@@ -183,7 +183,9 @@ func resourceGitlabGroupCreate(ctx context.Context, d *schema.ResourceData, meta
 		options.ShareWithGroupLock = gitlab.Bool(v.(bool))
 	}
 
-	if v, ok := d.GetOk("require_two_factor_authentication"); ok {
+	// nolint:staticcheck // SA1019 ignore deprecated GetOkExists
+	// lintignore: XR001 // TODO: replace with alternative for GetOkExists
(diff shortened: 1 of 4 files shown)


PatrickRice-KSC (commented):
@timofurrer  - So I ran a quick test, and I verified that this did fix the issue (which logically it would), but it feels like we should have at least one test with some of the "false default" values running to make sure the fix works. Thoughts?

timofurrer:
@PatrickRice-KSC absolutely. I've just added a test case, but didn't use `use_custom_template`, and the attributes required when importing from an URL to avoid a huge test setup. Since those are anyways `false` by default on upstream I think we are good here.

timofurrer:
@PatrickRice-KSC note that I've also added the fixes for the group resource as described in #1154.

PatrickRice-KSC:
> I've just added a test case, but didn't use `use_custom_template`, and the attributes required when importing from an URL to avoid a huge test setup. Since those are anyways `false` by default on upstream I think we are good here.

Makes complete sense, that'd be a ton of setup!

> note that I've also added the fixes for the group resource as described in https://github.com/gitlabhq/terraform-provider-gitlab/issues/1154.

Excellent, nice to knock out a couple issues with one PR!

PatrickRice-KSC (approved):

PR #1153, opened 2022-06-27
fix resource type in docs for project_approval_rule.md

mtesch-um:
Pretty sure this is what was meant in the example.

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/docs/resources/project_approval_rule.md
+++ b/docs/resources/project_approval_rule.md
@@ -58,8 +58,8 @@ resource "gitlab_project_approval_rule" "example-three" {
   user_ids           = [for user in data.gitlab_user.users : user.id]
 }
 
-# Example using `approval_rule`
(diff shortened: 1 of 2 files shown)


timofurrer (changes requested):
Good catch @mtesch-um !

Unfortunately, the files in `docs/resources` are generated and the examples are coming from the `examples` folder. You can make that same change and run `make reviewable` to generate the docs. 

I'd also suggest to change the comment above the resource:

```hcl
# Example using `approval_rule` using `any_approver` as rule type
resource "gitlab_project_approval_rule" "any_approver" {
  project            = 5
  name               = "Any name"
  rule_type          = "any_approver"
  approvals_required = 1
}
```

WDYT?

mtesch-um:
@timofurrer done!

Issue #1157, opened 2022-06-29
API calls to update project fail when certain fields are not passed from gitlab provider to underlying library


olhado:
### GitLab Provider version

3.15.0

### GitLab version

Gitalb 14.5.2 EE (Premium)

### Terraform version

Terraform v1.2.3

### Relevant Terraform Configuration

```hcl
resource "gitlab_project" "proj" {

  name                     = "test project"
  namespace_id             = gitlab_group.my_grp.id
  
  analytics_access_level   = "disabled"
  shared_runners_enabled   = true
}
```

### Relevant log output

```shell
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the
following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # gitlab_project will be updated in-place
  ~ resource "gitlab_project" "proj" {
      ~ analytics_access_level                           = "disabled" -> "private"
        id                                               = "<REDACTED>"
        name                                             = "test prioject"
        tags                                             = []
        # (57 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.

gitlab_project.proj: Modifying... [id=<REDACTED>]
╷
[...]

PR #1158, opened 2022-06-29
Fix a case where a change to a project in terraform can never apply w…

olhado:
…hen certain fields are modified.

Because the gitlab API actually expects at least on field out of a subset of fields in an update call (aside from `id`), if you only modify fields not in this subset, the Gitlab API will return a 400 complaining that one of the needed fields is missing.

One of these special fields is `name`. As this seems to be a safe field to assume always exists (we'll still defensively check if nil though), we pass it in always.

## Description

#1157 

- [ X] I acknowledge that all of the following items are true, where applicable:
    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/internal/provider/resource_gitlab_project.go
+++ b/internal/provider/resource_gitlab_project.go
@@ -1211,7 +1211,11 @@ func resourceGitlabProjectUpdate(ctx context.Context, d *schema.ResourceData, me
 	options := &gitlab.EditProjectOptions{}
 	transferOptions := &gitlab.TransferProjectOptions{}
 
-	if d.HasChange("name") {
+	// Always send the name field, to satisfy the requirement of having one
+	// of the project attributes listed below in the update call
+	// https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/lib/api/helpers/projects_helpers.rb#L120-188
+	// if d.HasChange("name") {
+	if d.Get("name").(string) != nil {
 		options.Name = gitlab.String(d.Get("name").(string))
 	}
 
--- a/internal/provider/resource_gitlab_project.go
+++ b/internal/provider/resource_gitlab_project.go
@@ -1208,14 +1208,16 @@ func resourceGitlabProjectRead(ctx context.Context, d *schema.ResourceData, meta
 func resourceGitlabProjectUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client := meta.(*gitlab.Client)
 
-	options := &gitlab.EditProjectOptions{}
-	transferOptions := &gitlab.TransferProjectOptions{}
-
 	// Always send the name field, to satisfy the requirement of having one
 	// of the project attributes listed below in the update call
 	// https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/lib/api/helpers/projects_helpers.rb#L120-188
-	// if d.HasChange("name") {
-	if d.Get("name").(string) != nil {
+	options := &gitlab.EditProjectOptions{
+		Name: gitlab.String(d.Get("name").(string)),
+	}
+
+	transferOptions := &gitlab.TransferProjectOptions{}
+
+	if d.HasChange("name") {
 		options.Name = gitlab.String(d.Get("name").(string))
 	}
 
--- a/internal/provider/resource_gitlab_project_test.go
+++ b/internal/provider/resource_gitlab_project_test.go
@@ -1171,9 +1171,6 @@ func TestAccGitlabProject_UpdateAnalyticsAccessLevel(t *testing.T) {
 						visibility_level = "public"
 						analytics_access_level = "disabled"
 					}`, rInt),
-				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr("gitlab_project.this", "analytics_access_level", "disabled"),
-				),
 			},
 			// Verify Import
 			{
--- a/internal/provider/resource_gitlab_project_test.go
+++ b/internal/provider/resource_gitlab_project_test.go
@@ -1139,6 +1139,56 @@ func TestAccGitlabProject_templateMutualExclusiveNameAndID(t *testing.T) {
 	})
 }
 
+// Gitlab update project API call requires one from a subset of project fields to be set (See #1157)
+// If only a non-blessed field is changed, this test checks that the provider ensures the code won't return an error.
+func TestAccGitlabProject_UpdateAnalyticsAccessLevel(t *testing.T) {
+	var received gitlab.Project
+	rInt := acctest.RandInt()
+
+	resource.Test(t, resource.TestCase{
+		ProviderFactories: providerFactories,
+		CheckDestroy:      testAccCheckGitlabProjectDestroy,
+		Steps: []resource.TestStep{
+			// Create minimal test project
+			{
+				Config: fmt.Sprintf(`
+					resource "gitlab_project" "this" {
+						name = "foo-%d"
+						visibility_level = "public"
+					}`, rInt),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGitlabProjectExists("gitlab_project.this", &received),
+				),
+			},
+			// Verify Import
+			{
+				ResourceName:      "gitlab_project.this",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+			// Update `analytics_access_level`
+			{
+				Config: fmt.Sprintf(`
+					resource "gitlab_project" "this" {
+						name = "foo-%d"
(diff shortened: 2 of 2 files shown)


internal/provider/resource_gitlab_project.go
```diff
@@ -1211,7 +1211,11 @@ func resourceGitlabProjectUpdate(ctx context.Context, d *schema.ResourceData, me
 	options := &gitlab.EditProjectOptions{}
```
timofurrer, line 1211: Instead of setting `options.Name` we can already set the `.Name` field during struct initialization:

```suggestion
	options := &gitlab.EditProjectOptions{
		Name: gitlab.String(d.Get("name").(string)),
	}
```

When doing so the check here:

https://github.com/gitlabhq/terraform-provider-gitlab/blob/b5a42dd1612facb1a1a8ea1ef0b9226ec898926c/internal/provider/resource_gitlab_project.go#L1480

will always be `true` - thus, we can remove the check entirely - even if an edit is theoretically not required if the options are equal to only having `Name` set to the already set project name.

timofurrer (changes requested):
Thanks for taking this up and creating the PR 🎉 

I've left a comment for some implementation guidance. In addition, I think we should add a test. You can use the `analytics_access_level` which didn't work in #1158 .

timofurrer:
@olhado no worries, we can help you out 🎉 

For the test case: I've just added another test case in the project resource within another open PR:

https://github.com/gitlabhq/terraform-provider-gitlab/pull/1152/files#diff-ed3c393c094bd3af56044aadf6a5394aae5a278a77cff854c582d6c5aed9b1f0R316

... you should be able to pretty much follow the same patterns :)

You can run your new test against any GitLab instance (e.g. the one with `make testacc-up` or even gitlab.com: `make testacc RUN=TestAccGitlabGroup_SetDefaultFalseBooleansOnCreate` (make sure to use your test name).

In case you are using GitPod: the GitLab test instance is already properly connected and you only need to run `make testacc RUN=...)`.

olhado:
Will pick this up in a few hours. That said, I am having issues with `make testacc-up`:
```
 make testacc-up
docker-compose up -d gitlab-ce
[+] Running 2/2
 ⠿ Container terraform-provider-gitlab_gitlab-ce_1  Recreated                                                     10.2s
 ⠿ Container terraform-provider-gitlab-gitlab-ce-1  Started                                                        0.2s
./scripts/await-healthy.sh
Waiting for GitLab container to become healthy...................
```
It ends up dying with these last logs:
```
terraform-provider-gitlab-gitlab-ce-1  | Recipe: gitlab::database_migrations
terraform-provider-gitlab-gitlab-ce-1  |   * ruby_block[check remote PG version] action nothing (skipped due to action :nothing)
terraform-provider-gitlab-gitlab-ce-1  |   * rails_migration[gitlab-rails] action run
terraform-provider-gitlab-gitlab-ce-1  |     * bash[migrate gitlab-rails database] action run
terraform-provider-gitlab-gitlab-ce-1  |
terraform-provider-gitlab-gitlab-ce-1  |       ================================================================================
terraform-provider-gitlab-gitlab-ce-1  |       Error executing action `run` on resource 'bash[migrate gitlab-rails database]'
terraform-provider-gitlab-gitlab-ce-1  |       ================================================================================
terraform-provider-gitlab-gitlab-ce-1  |
terraform-provider-gitlab-gitlab-ce-1  |       Mixlib::ShellOut::CommandTimeout
terraform-provider-gitlab-gitlab-ce-1  |       --------------------------------
terraform-provider-gitlab-gitlab-ce-1  |       Command timed out after 3600s:
terraform-provider-gitlab-gitlab-ce-1  |       Command exceeded allowed execution time, process terminated
terraform-provider-gitlab-gitlab-ce-1  |       ---- Begin output of "bash"  ----
```

internal/provider/resource_gitlab_project.go
olhado: Ah, this is cleaner. But now I wonder, is this going to cause spurious audit events in Gitlab, where the name is "changed"? We'll want to test for that, I think.

PatrickRice-KSC:
hey @olhado! 👋 

We've seen issues with the M1 machines before, specifically when running the x86 docker images like you mention. Very likely something weird to do with the ARM ->x86 emulation 🙁 . We have a default gitpod integration set up that you can try like Timo mentions that should help you out a bunch, because that way the docker-compose file will run inside the GitPod environment (which is x86, so it won't have the issue). To try it out with this PR, you can navigate to gitpod.io/#https://github.com/gitlabhq/terraform-provider-gitlab/pull/1158 and log in with your GitHub account - our setup in the repo will handle the rest and automatically start the environment for you! There's a terminal there you can use to run the make commands Timo mentions above also.

internal/provider/resource_gitlab_project_test.go
```diff
@@ -1139,6 +1139,56 @@ func TestAccGitlabProject_templateMutualExclusiveNameAndID(t *testing.T) {
...
+			// Update `analytics_access_level`
+			{
+				Config: fmt.Sprintf(`
+					resource "gitlab_project" "this" {
+						name = "foo-%d"
+						visibility_level = "public"
+						analytics_access_level = "disabled"
+					}`, rInt),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGitlabProjectExists("gitlab_project.this", &received),
```
timofurrer, line 1178: NIT: same here, this check is not really necessary.

```suggestion
```

internal/provider/resource_gitlab_project_test.go
```diff
@@ -1139,6 +1139,56 @@ func TestAccGitlabProject_templateMutualExclusiveNameAndID(t *testing.T) {
...
+	resource.Test(t, resource.TestCase{
+		ProviderFactories: providerFactories,
+		CheckDestroy:      testAccCheckGitlabProjectDestroy,
+		Steps: []resource.TestStep{
+			// Create minimal test project
+			{
+				Config: fmt.Sprintf(`
+					resource "gitlab_project" "this" {
+						name = "foo-%d"
+						visibility_level = "public"
```
timofurrer, line 1157: Let's set `analytics_access_level = "private"` to make sure that it's for sure not `disabled` during create, so that actually an update is happening.

```suggestion
						visibility_level                = "public"
						analytics_access_level = "private"
```

internal/provider/resource_gitlab_project_test.go
```diff
@@ -1139,6 +1139,56 @@ func TestAccGitlabProject_templateMutualExclusiveNameAndID(t *testing.T) {
 	})
 }
 
+// Gitlab update project API call requires one from a subset of project fields to be set (See #1157)
+// If only a non-blessed field is changed, this test checks that the provider ensures the code won't return an error.
+func TestAccGitlabProject_UpdateAnalyticsAccessLevel(t *testing.T) {
+	var received gitlab.Project
```
timofurrer, line 1145: Given my other comments - you can remove the `received` variable.

timofurrer (changes requested):
I've left another set of comments with suggestions. You may just apply them :)

internal/provider/resource_gitlab_project_test.go
```diff
@@ -1139,6 +1139,52 @@ func TestAccGitlabProject_templateMutualExclusiveNameAndID(t *testing.T) {
...
+			{
+				Config: fmt.Sprintf(`
+					resource "gitlab_project" "this" {
+						name = "foo-%d"
+						visibility_level = "public"
+						analytics_access_level = "disabled"
+					}`, rInt),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("gitlab_project.this", "analytics_access_level", "disabled"),
+				),
```
timofurrer, line 1176: I've missed that last time: because it's the only check left we don't need the compose.

```suggestion
				resource.TestCheckResourceAttr("gitlab_project.this", "analytics_access_level", "disabled"),
```

olhado: Do we still need the `Check:` though?

timofurrer: hmmm, you actually don't :) 👍 The subsequent `Verify Import` will make sure that what's coming from the read API matches the config and also this very step checks if the plan is empty after the apply..

```suggestion
```

olhado:
Just want to say, thanks to this team for helping a newcomer along the way here. It probably took you longer than doing it yourself, but it will make me more likely to push fixes for anything else I might find in the future. 😄 

FYI, the tests seem to be written in a couple of different ways. Is there a recommended structure (using verify import blocks vs composing test check chains)?

timofurrer (approved):
Good work @olhado 🎉 

LGTM 🚢

Issue #696, opened 2021-08-15
No changes detected in gitlab_project.approvals_before_merge 


nderraugh:
### Terraform Version
Terraform v1.0.4
on darwin_amd64

### Affected Resource
- gitlab_project

### Terraform Configuration Files
```hcl
terraform {
  required_providers {
    gitlab = {
      source  = "gitlabhq/gitlab"
      version = "3.7.0"
    }
  }
}
data "gitlab_group" "somegroup" {
  full_path = "somegroup"
}
resource "gitlab_project" "project" {
  name                                             = "some name"
  namespace_id                                     = gitlab_group.somegroup.id
  approvals_before_merge                           = 2
  merge_method                                     = "ff"
  description                                      = "description"
  remove_source_branch_after_merge                 = true
  only_allow_merge_if_all_discussions_are_resolved = true
  only_allow_merge_if_pipeline_succeeds            = true
}
```

### Expected Behavior
The provider should detect the difference between the number of approvals configured in TF and Gitlab.  Running `terraform plan` should show the changes made outside of TF and indicate that applying the current changes will overwrite those changes.

### Actual Behavior
The provider does not detect any difference between TF and the configuerd Gitlab project.

### Steps to reroduce
Configure the code above for a test repository.  Change the Gitlab project config to a single approval required before merge, e.g. project > Settings > General > Merge request (MR) approvals > Any eligible User > 2 -> 1 and re-run `terraform plan`.

PatrickRice-KSC:
https://gitlab.com/groups/gitlab-org/-/epics/7572 is the epic for removing the `approvals_before_merge`, which looks based on conversations like it's now slated to be removed in 15.0

timofurrer:
Let's have a look for our next release 👍

timofurrer:
Unfortunately, this field still exists after 15.0 and I have the feeling it's not gonna be removed before 16.0 at least.

I don't see any feasible solution from the provider side and therefore I've created https://github.com/gitlabhq/terraform-provider-gitlab/pull/1161 which just adds a note for the `approvals_before_merge` field. 

@PatrickRice-KSC WDYT?

PatrickRice-KSC:
[...]

PR #1161, opened 2022-07-03
Note the danger of using `approvals_before_merge` attribute in the project resource

timofurrer:
Closes: #696


PatrickRice-KSC (approved):
Merging so we can get the warning into 3.16!

Issue #1167, opened 2022-07-11
Let the provider run without Admin rights


SeameX:
### Feature Description

Hello,
since version 3.16 we can't use the provider without admin rights anymore.

```
gitlab_project.this: Creating...
╷
│ Error: Failed to discover if branch protection is enabled by default or not for project 544: GET https://gitlab.timocom.net/api/v4/application/settings: 403 {message: 403 Forbidden}
│ 
│   with gitlab_project.this,
│   on main.tf line 1, in resource "gitlab_project" "this":
│    1: resource "gitlab_project" "this" {
│ 
╵
```

It would be awesome if you could add a feature toggle to disable that.

### Do you want to implement this?

- [ ] I would like to implement this myself 👷

timofurrer:
@SeameX thanks for the report! This indeed is an issue we've introduced by accident - to solve another issue.

I'll have a look at it asap and try to release a new version soon.

Issue #1127, opened 2022-06-09
Unable to Create Project: error while waiting for branch main to reach 'protected' status


tschechniker:
### GitLab Provider version

3.15.1

### GitLab version

14.10.3

### Terraform version

1.1.3

### Relevant Terraform Configuration

```hcl
resource "gitlab_project" "test_project" {
    name = "test"
    description = "test"
    default_branch = "main"
    initialize_with_readme = true
}
```

### Relevant log output

```shell
Error: error while waiting for branch main to reach 'protected' status, timeout while waiting for state to become 'true'
```

### Description

Currently i'm unable to create a new Project if it's not in a group where branch protection is turned off in the group settings. It seems like the default behaviour of gitlab has changed and in the project creation there is no default branch protection any more. This also applies if i create the project manually. There is not branch protection created for the default branch.

https://github.com/gitlabhq/terraform-provider-gitlab/blob/main/internal/provider/resource_gitlab_project.go#L1083 here you rely on that there will always be a branch protection. And course of the fact that gitlab will not create one the timeout of 2 minutes hit always.

Till now i was unable to find any working workaround for this issue. Am i the only one?

timofurrer:
Thanks for the report @tschechniker 🎉 

Interesting, we've a coupled of these issues lately also related to the `initialize_with_readme = true`, see #1111. If you set it to `false`, are you experiencing the same issue?

> [...] here you rely on that there will always be a branch protection.

Not really *always* as for groups for which it is turned of, we don't rely on that, see:

https://github.com/gitlabhq/terraform-provider-gitlab/blob/dbc86a65b8268761ee28211036a8e5b01a44b79f/internal/provider/resource_gitlab_project.go#L1761

What I'm wondering is what your `default_branch_protection` on instance level is? Can you [verify that](https://docs.gitlab.com/ee/api/settings.html#get-current-application-settings)? Did this change for your instance?

tschechniker:
if i turn initialize_with_readme to off it works yes but this will break all other steps so no option for me currently. 

I will check what the instance level branch protection is set to.

timofurrer:
> if i turn initialize_with_readme to off it works yes but this will break all other steps so no option for me currently.

Alright, yes, was just out of curiosity and not meant to be a workaround :)

> I will check what the instance level branch protection is set to.

Thanks, keep as posted.

I'll draft a PR to also check the application defaults, since that's a bug IMHO anyways ...

[...]

PR #1169, opened 2022-07-11
resource/gitlab_project: Fix admin token requirement to check default branch protection

timofurrer:
This bugfix change set gracefully handles checking for the
instance-level default branch protection setting if the provider token
is not from an administrator user.
This bug was introduced in
https://github.com/gitlabhq/terraform-provider-gitlab/pull/1128 in an
attempt to check the instance-level wide settings - which unfortunately,
and not to my knowledge at the time requires an admin token.

In case the token is not from an admin the project resource won't check
for the settings and by default restore the faulty behavior before #1128
which was to wait for the default branch protection to be created.
However, if this won't happen because it's disabled on the
instance-level the user can set
`skip_wait_for_default_branch_protection = true` which skips waiting for
the default branch protection.

Closes: #1167

### DIFF
--- a/internal/provider/resource_gitlab_project.go
+++ b/internal/provider/resource_gitlab_project.go
@@ -671,7 +671,20 @@ branch using a ` + "`DELETE`" + ` request. Then define the desired branch protec
 		Importer: &schema.ResourceImporter{
 			StateContext: schema.ImportStatePassthroughContext,
 		},
-		Schema: resourceGitLabProjectSchema,
+		Schema: constructSchema(resourceGitLabProjectSchema, map[string]*schema.Schema{
+			"skip_wait_for_default_branch_protection": {
+				Description: `If ` + "`true`" + `, the default behavior to wait for the default branch protection to be created is skipped.
+This is necessary if the current user is not an admin and the default branch protection is disabled on an instance-level.
+There is currently no known way to determine if the default branch protection is disabled on an instance-level for non-admin users.
+This attribute is only used during resource creation, thus changes are suppressed and the attribute cannot be imported.
+`,
+				Type:     schema.TypeBool,
+				Optional: true,
+				DiffSuppressFunc: func(string, string, string, *schema.ResourceData) bool {
+					return true
+				},
+			},
+		}),
 		CustomizeDiff: customdiff.All(
 			customdiff.ComputedIf("path_with_namespace", namespaceOrPathChanged),
 			customdiff.ComputedIf("ssh_url_to_repo", namespaceOrPathChanged),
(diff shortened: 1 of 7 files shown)


internal/provider/resource_gitlab_project.go
```diff
@@ -1106,41 +1115,43 @@ func resourceGitlabProjectCreate(ctx context.Context, d *schema.ResourceData, me
 		}
 	}
 
-	// If the project is assigned to a group namespace and the group has *default branch protection*
-	// disabled (`default_branch_protection = 0`) then we don't have to wait for one.
-	waitForDefaultBranchProtection, err := expectDefaultBranchProtection(ctx, client, project)
-	if err != nil {
-		return diag.Errorf("Failed to discover if branch protection is enabled by default or not for project %d: %+v", project.ID, err)
-	}
+	if d.Get("skip_wait_for_default_branch_protection").(bool) {
```
PatrickRice-KSC, line 1118: @timofurrer  - I think you want this to be a "not" check, right? Essentially, you are entering this `if` branch if they _do_ want to wait for the branch protection, which means the `skip` flag should be set to false. It's a weird double negative, but I _think_ that's the intention here.

PatrickRice-KSC: @timofurrer  - Nevermind, ran through a couple examples for myself, I'm good with the logic now.

internal/provider/resource_gitlab_project.go
```diff
@@ -671,7 +671,16 @@ branch using a ` + "`DELETE`" + ` request. Then define the desired branch protec
...
 		},
-		Schema: resourceGitLabProjectSchema,
+		Schema: constructSchema(resourceGitLabProjectSchema, map[string]*schema.Schema{
+			"skip_wait_for_default_branch_protection": {
+				Description: `If ` + "`true`" + `, the default behavior to wait for the default branch protection to be created is skipped.
+This is necessary if the current user is not an admin and the default branch protection is disabled on an instance-level.
+There is currently no known way to determine if the default branch protection is disabled on an instance-level for non-admin users.`,
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
```
PatrickRice-KSC, line 681: @timofurrer  - Technically setting this to false creates a breaking change. Configs that would previously work will now stop working. Do we need to set this to `true` for backwards compatibility?

PatrickRice-KSC: Resolved with inverting the boolean check.

timofurrer: hmmm .. can you elaborate? We previously had this https://github.com/gitlabhq/terraform-provider-gitlab/issues/1127 which caused a timeout (which is now the case with a non-admin token and this set to `false`).

PatrickRice-KSC (approved):

PatrickRice-KSC (approved):

PatrickRice-KSC:
Holding little bit on this before we merge it - Timo and I have an outstanding discussion on Discord around whether we need to store this value in state. We likely do, because otherwise we'll have non-empty plans on subsequent runs.

PatrickRice-KSC (approved):

Issue #1174, opened 2022-07-13
gitlab_user_sshkey resources do not converge


exarkun:
### GitLab Provider version

3.14.0

### GitLab version

14.9.3

### Terraform version

0.13.7

### Relevant Terraform Configuration

```hcl
resource "gitlab_group_access_token" "bot-token" {
  group = 123
  name = "Bot Token"
  access_level = "developer"
  scopes = [ "api" ]
}

resource "tls_private_key" "bot-key" {
  algorithm = "ED25519"
}

resource "gitlab_user_sshkey" "bot-key" {
  user_id = gitlab_group_access_token.bot-token.user_id
  title = "bot-key"
  key = tls_private_key.bot-key.public_key_openssh
}
```

### Relevant log output

```shell
# gitlab_user_sshkey.bot-key must be replaced
-/+ resource "gitlab_user_sshkey" "bot-key" {
      ~ created_at = "2022-07-13T19:49:34Z" -> (known after apply)
      ~ id         = "14:5" -> (known after apply)
      ~ key        = <<~EOT # forces replacement
          - ssh-ed25519 AAAA... Bot Token (whetstone.privatestorage-staging.com)
          + ssh-ed25519 AAAA...
        EOT
      ~ key_id     = 5 -> (known after apply)
        title      = "bot-key"
        user_id    = 14
    }
```

### Description

GitLab seems to modify the supplied key by adding a username and hostname.  The `gitlab_user_sshkey` detects this difference and decides it needs to push its value.  However, no matter how many times it does this, GitLab just re-modifies the value and the deployment never converges (ie, changes are always seen as required).

timofurrer:
@exarkun thanks for the report!

The problem here is that the `tls_private_key` resource generates the `public_key_openssh` in a [friendly format for `./ssh/authorized_keys`](https://github.com/hashicorp/terraform-provider-tls/issues/163#issuecomment-1066724040) by appending a newline character `\n` at the end, you can verify this in the state file:

![image](https://user-images.githubusercontent.com/1008252/178914980-adfea48c-8a8a-4c96-becb-8c17564df20f.png)

In the provider we don't trim any whitespaces from the key at the moment (but I've just created #1175 to mitigate that). 

Therefore, an easy workaround would be to trim it in your config:

```hcl
resource "gitlab_group_access_token" "bot-token" {
  group = 123
  name = "Bot Token"
  access_level = "developer"
  scopes = [ "api" ]
}

[...]

PR #1175, opened 2022-07-14
resource/gitlab_user_sshkey: Ignore leading and trailing whitespaces in key

timofurrer:
This change set handles ssh keys with leading or trailing whitespaces gracefully - they are simply ignored when diffing and there won't trigger an update.

This is a probably a common source of issue when using it together with `tls_private_key.public_key_openssh` which appends a newline.

Closes: #1174


PatrickRice-KSC (approved):
LGTM! 🚀

Issue #1166, opened 2022-07-11
Feature request: Support user GPG keys management


yan12125:
### Feature Description

Hi, could you consider adding a new resource for managing GPG keys like [the one](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/user_gpg_key) from the GitHub provider? Apparently go-gitlab library already [supports that](https://github.com/xanzy/go-gitlab/pull/1330).

An example may be: (modified from examples for `github_user_gpg_key` and `gitlab_user_sshkey`)
```hcl
data "gitlab_user" "example" {
  username = "example-user"
}

resource "gitlab_user_gpgkey" "example" {
  user_id            = data.gitlab_user.example.id
  key                = "-----BEGIN PGP PUBLIC KEY BLOCK-----\n...\n-----END PGP PUBLIC KEY BLOCK-----"
}
```

When the field `key` is changed, there might be a forced replacement as current API [does not support updating an existing key](https://gitlab.com/gitlab-org/gitlab/-/issues/296977).

### Do you want to implement this?

- [x] I would like to implement this myself 👷 (may not be able to due to little experience with Go)

yan12125:
I started looking into implementing it, and got a question. I'd like to manage GPG keys for the current user with a user token. Some people may want to manage GPG keys for whatever users with an admin token. The two scenarios need different APIs (ex: `POST /user/gpg_keys` and admin-only `POST /users/:id/gpg_keys`). How can I design the resource to fit both scenarios?

timofurrer:
@yan12125 I suppose the simplest approach would be to just use the appropriate API depending on if the `user_id` attribute is given in the config. 

Let's consider the two examples:

```hcl
data "gitlab_user" "example" {
  username = "example-user"
}

resource "gitlab_user_gpgkey" "example" {
  user_id            = data.gitlab_user.example.id
  armored_public_key = "-----BEGIN PGP PUBLIC KEY BLOCK-----\n...\n-----END PGP PUBLIC KEY BLOCK-----"
}
```

This would use the `/users/:id/gpg_keys` API and require an admin token, because the `user_id` is given. AFAIK, even if the `:id` in the API call is the currently authenticated user - it is only available for admins!

The following example would use the `/user/gpg_keys` API and would operate on the currently authenticated users gpg keys determined by the token used:

```hcl
resource "gitlab_user_gpgkey" "example" {
[...]

Issue #1039, opened 2022-04-22
Failure in `main` pipeline


timofurrer:
We seem to have an issue in the acceptance tests running on `main` branch:

https://github.com/gitlabhq/terraform-provider-gitlab/runs/6124104979?check_suite_focus=true#step:7:217

```
=== RUN   TestAccGitlabProject_basic
    resource_gitlab_project_test.go:123: Step 1/19 error: After applying this test step and performing a `terraform refresh`, the plan was not empty.
        stdout
        
        
        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
          ~ update in-place
        
        Terraform will perform the following actions:
        
          # gitlab_project.foo will be updated in-place
          ~ resource "gitlab_project" "foo" {
                id                                               = "73"
                name                                             = "foo-1835780171598199123"
                tags                                             = [
                    "bar",
                    "foo",
                ]
                # (64 unchanged attributes hidden)
        
              ~ container_expiration_policy {
                  ~ enabled     = false -> true
                    # (4 unchanged attributes hidden)
                }
        
                # (1 unchanged block hidden)
            }
        
        Plan: 0 to add, 1 to change, 0 to destroy.
```

It seems that the `container_expiration_policy` has a new value. Is that because we've enabled the container registry? Why didn't it fail earlier?

timofurrer:
Also, it only seems to affect EE.

timofurrer:
Okay, it seems to be a sporadic issue ... a retry passed ...

PatrickRice-KSC:
I'll check this out, but I'd venture that since it seems to be saying there is a pending change on the container_expiration_policy and that it's sporatic, it's likely that the container expiration policy is a background apply similar to what we've seen on other resources and it needs a wait block to ensure it's consistent.

timofurrer:
Yes, it's what I suspected, too.

It goes into the same direction of https://gitlab.com/gitlab-org/gitlab/-/issues/352299

PatrickRice-KSC:
So I ran the same make script I wrote back when I was testing the previous fix to run the test until it failed, and over 1,000 tests completed successfully before my gitpod environment shut down 🙁 

That would seem to indicate if it _is_ a race condition, it's a _very_ rare one. I can still put in that fix, but I'm going to have a really hard time seeing if it actually fixes the issue. I'll give it another try to see if I can get it to repro.

timofurrer:
[...]

PR #1181, opened 2022-07-21
Add resource gitlab_user_gpgkey

yan12125:
## Description

Somehow `terraform plan` always reports the field armored_public_key as
changed.

Closes https://github.com/gitlabhq/terraform-provider-gitlab/issues/1166

    - Create the resource
    - Update the attributes
    - Import the resource


yan12125:
> Somehow terraform plan always reports the field key as changed.

Here are steps to reproduce:
* Create a resource like this:

```hcl
resource "gitlab_user_gpgkey" "yan12125" {
    key = <<KEY
-----BEGIN PGP PUBLIC KEY BLOCK-----

xxx
-----END PGP PUBLIC KEY BLOCK-----
KEY
}
```

* Import the existing key with `terraform import gitlab_user_gpgkey.yan12125 123`, where 123 is the key ID for the current user.
* `terraform plan` reports changed resource: (personal information removed)

```
Warning: Provider development overrides are in effect

The following provider development overrides are set in the CLI configuration:
 - gitlabhq/gitlab in /home/yen/Projects/terraform-provider-gitlab

The behavior may therefore not match any released version of the provider and applying changes may cause the state to become incompatible with published releases.
gitlab_user_gpgkey.yan12125: Refreshing state... [id=130860]

Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:

  # gitlab_user_gpgkey.yan12125 must be replaced
-/+ resource "gitlab_user_gpgkey" "yan12125" {
      ~ created_at = "2022-07-19T16:45:31Z" -> (known after apply)
      ~ id         = "123" -> (known after apply)
      ~ key        = <<-EOT # forces replacement
            -----BEGIN PGP PUBLIC KEY BLOCK-----
            
            xxx
            -----END PGP PUBLIC KEY BLOCK-----
        EOT
      ~ key_id     = 123 -> (known after apply)
    }

─────────────────────────────────────────────────────────────────────────────

Note: You didn't use the -out option to save this plan, so Terraform can't
guarantee to take exactly these actions if you run "terraform apply" now.
```

The key recorded in terraform.tfstate is the same as my actual key, but somehow my code makes terraform treat them as different.

timofurrer:
@yan12125 The diff is because with your `<<KEY .. KEY` syntax the key has a trailing newline which is removed by GitLab. It's a similar situation as I've described in https://github.com/gitlabhq/terraform-provider-gitlab/issues/1174#issuecomment-1184056524.

I suggest that you remove leading and trailing whitespaces from the key in the create function.

yan12125:
Thanks! Now `terraform plan` no longer reports differences. Next steps include adding tests and docs. I will adding tests first.

yan12125:
Added docs and tests - now ready for reviews :)

internal/provider/resource_gitlab_user_gpgkey.go
```diff
@@ -0,0 +1,187 @@
...
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	gitlab "github.com/xanzy/go-gitlab"
+)
+
+var _ = registerResource("gitlab_user_gpgkey", func() *schema.Resource {
+	return &schema.Resource{
+		Description: `The ` + "`" + `gitlab_user_gpgkey` + "`" + ` resource allows to manage the lifecycle of a GPG key assigned to the current user or a specific user.
```
timofurrer, line 18: I think it makes sense to add a little info box to let readers know that managing keys for arbitrary users requires administrative privileges. WDYT?

```suggestion
		Description: `The ` + "`" + `gitlab_user_gpgkey` + "`" + ` resource allows to manage the lifecycle of a GPG key assigned to the current user or a specific user.
		
-> Managing GPG keys for arbitrary users requires admin privileges.
```

internal/provider/resource_gitlab_user_gpgkey.go
```diff
@@ -0,0 +1,187 @@
...
+
+**Upstream API**: [GitLab API docs](https://docs.gitlab.com/ee/api/users.html#get-a-specific-gpg-key)`,
```
timofurrer, line 20: NIT:

```suggestion
**Upstream API**: [GitLab REST API docs](https://docs.gitlab.com/ee/api/users.html#get-a-specific-gpg-key)`,
```

internal/provider/resource_gitlab_user_gpgkey.go
```diff
@@ -0,0 +1,187 @@
...
+	options := &gitlab.AddGPGKeyOptions{
+		Key:   gitlab.String(strings.TrimSpace(d.Get("key").(string))),
+	}
+
+	var key *gitlab.GPGKey
+	var err error
+	userID, userIDOk := d.GetOk("user_id")
+	if userIDOk {
+		isAdmin, isAdminErr := isCurrentUserAdmin(ctx, client)
+		if isAdminErr != nil {
```
timofurrer, line 73: NIT: (let's stick to the `err` idiom :)) 

```suggestion
		isAdmin, err := isCurrentUserAdmin(ctx, client)
		if err != nil {
```

internal/provider/resource_gitlab_user_gpgkey.go
```diff
@@ -0,0 +1,187 @@
...
+			key = k
+			break
+		}
+	}
+
+	if key == nil {
+		log.Printf("Could not find GPG key %d for user %d", keyID, userID)
+		d.SetId("")
+		return nil
+	}
```
timofurrer, line 127: Any particular reason for not using `client.Users.GetGPGKey` and `client.users.GetGPGKeyForUser` ?

This would simplify the code quite a bit (and we wouldn't need to implement pagination logic, which is a problem in the current implementation ;))

Something along the lines of this: (didn't actually test this)

```suggestion
	var key *gitlab.GPGKey
	if userID != 0 {
		key, _, err = client.Users.GetGPGKeyForUser(userID, keyID, gitlab.WithContext(ctx))
	} else {
		key, _, err = client.Users.GetGPGKey(keyID, gitlab.WithContext(ctx))
	}
	if err != nil {
		if is404(err) {
			log.Printf("Could not find GPG key %d for user %d, removing from state", keyID, userID)
			d.SetId("")
			return nil
		}
		return diag.FromErr(err)
	}
```

internal/provider/resource_gitlab_user_gpgkey.go
```diff
@@ -0,0 +1,187 @@
...
+
+	if userID != 0 {
+		d.Set("user_id", userID)
+	}
+	d.Set("key_id", keyID)
+	d.Set("key", strings.TrimSpace(key.Key))
+	if key.CreatedAt != nil {
```
timofurrer, line 134: NIT: probably always could to check, but did you actually run into this? I don't think it will every be `nil` ...

internal/provider/resource_gitlab_user_gpgkey.go
```diff
@@ -0,0 +1,187 @@
...
+	client := meta.(*gitlab.Client)
+
+	_, keyID, err := resourceGitlabUserGPGKeyParseID(d.Id())
+	if err != nil {
+		return diag.Errorf("unable to parse user GPG key resource id: %s: %v", d.Id(), err)
+	}
+
+	if userID, ok := d.GetOk("user_id"); ok {
+		isAdmin, isAdminErr := isCurrentUserAdmin(ctx, client)
+		if isAdminErr != nil {
```
timofurrer, line 150: NIT:

```suggestion
		isAdmin, err := isCurrentUserAdmin(ctx, client)
		if err != nil {
```

internal/provider/resource_gitlab_user_gpgkey_test.go
```diff
@@ -0,0 +1,285 @@
...
+	var key gitlab.GPGKey
+	testUser := testAccCreateUsers(t, 1)[0]
+
+	resource.ParallelTest(t, resource.TestCase{
+		ProviderFactories: providerFactories,
+		CheckDestroy:      testAccCheckGitlabUserGPGKeyDestroy,
+		Steps: []resource.TestStep{
+			// Create a user + gpgkey
+			{
+				Config: testAccGitlabUserGPGKeyConfig(testUser.ID, testEd25519GPGPubKey),
```
timofurrer, line 102: Let's include the config in the test here, see https://github.com/gitlabhq/terraform-provider-gitlab/issues/1008#:~:text=Prefer%20writing%20config%20in%2Dline%20to%20wrapping%20it%20in%20a%20function.

internal/provider/resource_gitlab_user_gpgkey_test.go
```diff
@@ -0,0 +1,285 @@
...
+				Check: resource.ComposeTestCheckFunc(
```
timofurrer, line 103: I think it's enough to have subsequent import verification test steps in this case, see https://github.com/gitlabhq/terraform-provider-gitlab/issues/1008#:~:text=and%20data%20sources.-,Use%20ImportStateVerify%20TestSteps%20as%20a%20replacement%20for%20writing%20custom%20%22CheckExists%22%20and%20%22CheckAttributes%22%20check%20funcs.,-Undesirable%3A

internal/provider/resource_gitlab_user_gpgkey_test.go
```diff
@@ -0,0 +1,285 @@
...
+			// Update the key
+			{
+				Config: testAccGitlabUserGPGKeyConfig(testUser.ID, updatedEd25519GPGPubKey),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGitlabUserGPGKeyExists("gitlab_user_gpgkey.foo_key", &key),
+					testAccCheckGitlabUserGPGKeyAttributes(&key, &testAccGitlabUserGPGKeyExpectedAttributes{
+						Key:       updatedEd25519GPGPubKey,
+					}),
+				),
+			},
```
timofurrer, line 119: Can you add a test case for when someone switches from the current user to another arbitrary user gpg key (and maybe vice-versa) ?

go.mod
```diff
@@ -62,3 +62,5 @@ require (
 	google.golang.org/protobuf v1.28.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
+
+replace github.com/xanzy/go-gitlab => github.com/yan12125/go-gitlab v0.69.1-0.20220718135531-011d76e914b5
```
timofurrer, line 66: Just adding this comment to block from an accidental merge :)

timofurrer (changes requested):
It already looks quite good! I've added a few nitpicks and other comments.

Back to you 🏓

yan12125:
Thanks! I applied some trivial changes. Others need some time :)

internal/provider/resource_gitlab_user_gpgkey.go
yan12125: > Any particular reason for not using `client.Users.GetGPGKey` and `client.users.GetGPGKeyForUser` ?

Because I copy the code from `resource_gitlab_user_sshkey.go` :joy: 

Your suggested fix "almost" works, many thanks! The only missing bit is another bug in go-gitlab - https://github.com/xanzy/go-gitlab/pull/1512

timofurrer: > Because I copy the code from `resource_gitlab_user_sshkey.go` 😂

Yes, I've only recently added this to GitLab itself: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/81790

> Your suggested fix "almost" works, many thanks! The only missing bit is another bug in go-gitlab - https://github.com/xanzy/go-gitlab/pull/1512

Awesome! Thanks for creating to PR over at go-gitlab 🎉

internal/provider/resource_gitlab_user_gpgkey_test.go
yan12125: That looks reasonable. I switched to `ImportStateVerify` in 61381e5ca48d5e3f969e54b328bbd94a4e9c61fa. Not quite sure if the usage is idiomatic.

internal/provider/resource_gitlab_user_gpgkey.go
yan12125: > did you actually run into this?

Nope. Again, I copied codes from `resource_gitlab_user_sshkey.go` without checking it in depth lol

internal/provider/resource_gitlab_user_gpgkey_test.go
yan12125: Done in 4794d477e21f973f5e85fcb43db159f85d4d164c, thanks!

yan12125:
Rebased on top of master as upgrading go-gitlab conflicts with upgrades of other dependencies (https://github.com/gitlabhq/terraform-provider-gitlab/pull/1184)

examples/resources/gitlab_user_gpgkey/resource.tf
```diff
@@ -0,0 +1,14 @@
+data "gitlab_user" "example" {
+  username = "example-user"
+}
+
+# Manages a GPG key for the specified user. An admin token is required if `user_id` is specified.
+resource "gitlab_user_sshkey" "example" {
```
timofurrer, line 6:
```suggestion
resource "gitlab_user_gpgkey" "example" {
```

examples/resources/gitlab_user_gpgkey/resource.tf
```diff
@@ -0,0 +1,14 @@
...
+  user_id = data.gitlab_user.example.id
+  key     = "-----BEGIN PGP PUBLIC KEY BLOCK-----\n...\n-----END PGP PUBLIC KEY BLOCK-----"
+}
+
+# Manages a GPG key for the current user
+resource "gitlab_user_sshkey" "example_user" {
```
timofurrer, line 12:
```suggestion
resource "gitlab_user_gpgkey" "example_user" {
```

internal/provider/resource_gitlab_user_gpgkey_test.go
timofurrer: @yan12125 I've noticed your 👍 , but haven't found the test. Did you forget to push it?

timofurrer (changes requested):
A few minor things and make sure to run `make reviewable` before the next push, to re-generate the docs etc.

Back to you 🏓

timofurrer:
@yan12125 I'm updating go-gitlab via https://github.com/gitlabhq/terraform-provider-gitlab/pull/1195

internal/provider/resource_gitlab_user_gpgkey_test.go
yan12125: Sorry for confusion. I like the idea but had no time to add that yesterday. Now a test is expanded :)

yan12125:
Thanks for the reminder about `make renewable`. I run it locally with Go 1.19, and it reports that `io/ioutil` is deprecated. I didn't find that error on CI checks, so I keep that as-is.
```
internal/provider/config.go:7:2: SA1019: "io/ioutil" has been deprecated since Go 1.16: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details. (staticcheck)
        "io/ioutil"
        ^
```

timofurrer:
@yan12125 Can you change that code in the config.go file to use `os` and `os.ReadFile` instead of `ioutils`?

yan12125:
> Can you change that code in the config.go file to use `os` and `os.ReadFile` instead of ioutils?

I can try it out. Is it better to make changes in this PR or another PR?

timofurrer:
@yan12125 let's do another PR, since you've asked ;)

yan12125:
>  let's do another PR, since you've asked ;)

Here you go~ https://github.com/gitlabhq/terraform-provider-gitlab/pull/1199

---

Regarding this PR, I rebased on top of the latest main branch and drop changes to go.mod/go.sum.

timofurrer:
@yan12125 I think we are just *a rebase away* 🎉 (I've merged https://github.com/gitlabhq/terraform-provider-gitlab/pull/1199)

yan12125:
Great thanks! Rebased and squashed :)

yan12125:
Test failures seem unrelated. Both 15.1.3-ce.0 and 15.1.3-ee.0 tests fail with:
```
    resource_gitlab_project_test.go:121: Step 19/19 error: After applying this test step and performing a `terraform refresh`, the plan was not empty.
        stdout
        
        
        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
          ~ update in-place
        
          # gitlab_project.foo will be updated in-place
          ~ resource "gitlab_project" "foo" {
                id                                               = "93"
                name                                             = "foo-1604740959322635418"
                tags                                             = [
                    "bar",
                    "foo",
                ]
                # (63 unchanged attributes hidden)
        
              ~ container_expiration_policy {
                  ~ enabled     = false -> true
                    # (4 unchanged attributes hidden)
                }
            }
        
        Plan: 0 to add, 1 to change, 0 to destroy.
```
Which appears the same issue as https://github.com/gitlabhq/terraform-provider-gitlab/issues/1039.

timofurrer (approved):
LGTM 🎉

timofurrer:
@yan12125 I've re-triggered the tests and it's all looking good now. Nice work 🎉 🚢

yan12125:
Thank you very much for the time!

PR #1188, opened 2022-07-26
Officially support GitLab 15.0, 15.1 and 15.2

timofurrer:
A few days late, but better than never :)

### DIFF
--- a/internal/provider/resource_gitlab_service_jira.go
+++ b/internal/provider/resource_gitlab_service_jira.go
@@ -81,7 +81,7 @@ var _ = registerResource("gitlab_service_jira", func() *schema.Resource {
 				Sensitive:   true,
 			},
 			"jira_issue_transition_id": {
-				Description: "The ID of a transition that moves issues to a closed state. You can find this number under the JIRA workflow administration (Administration > Issues > Workflows) by selecting View under Operations of the desired workflow of your project. By default, this ID is set to 2. **Note**: importing this field is currently not supported.",
+				Description: "The ID of a transition that moves issues to a closed state. You can find this number under the JIRA workflow administration (Administration > Issues > Workflows) by selecting View under Operations of the desired workflow of your project. By default, this ID is set to 2. *Note**: importing this field is only supported since GitLab 15.2.",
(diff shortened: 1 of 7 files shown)


timofurrer:
I'll also follow up here to remove all the GitLab 14.X compatibility code ...

internal/provider/resource_gitlab_service_jira.go
```diff
@@ -185,10 +185,12 @@ func resourceGitlabServiceJiraRead(ctx context.Context, d *schema.ResourceData,
...
-	// FIXME: The API or go-gitlab doesn't return `jira_issue_transition_id` properly,
-	//        therefore we don't overwrite the value set in the config, in case it's not set in the API
-	//        response. This currently impacts the import of this attribute.
-	if jiraService.Properties.JiraIssueTransitionID != "" {
+
+	hasJiraIssueTransitionIDFixed, err := isGitLabVersionAtLeast(ctx, client, "15.2")()
+	if err != nil {
+		return diag.Errorf("failed to check if `jira_issue_transition_id` is properly supported in GitLab version: %v", err)
+	}
+	if hasJiraIssueTransitionIDFixed || jiraService.Properties.JiraIssueTransitionID != "" {
```
PatrickRice-KSC, line 193: @timofurrer  - Is the version check required if the field will be `""` prior to version 15.2? If seems like it's either redundant (because it'll always be "") or the operator for the if statement should be a `&&` instead.

WDYT?

timofurrer: @PatrickRice-KSC It should be correct as is. Because of the referenced bug in the comment prior to GitLab 15.2 we didn't set the `jira_issue_transision_id` state explicitly because it is never returned from the API. If we don't set it explictly the value from the config is used. If we wouldn't do this you'll always get a non-empty plan when applying because you'd set `""` always.

For GitLab 15.2 we always want to set it to the value returned from the API (because of the bug fix) OR we want to have the behavior as before which is the second `!= ""` condition.

I hope it makes more sense now 😄

PatrickRice-KSC (approved):
Looks good!

Issue #1197, opened 2022-08-03
Add gitlab_repository_tree data source 


Shocktrooper:
### Feature Description

Seeing as we do not currently have a way to query files in an existing GitLab project I would like to introduce a new `gitlab_repository_tree` data source to be able to fetch this data 

References:

- https://github.com/xanzy/go-gitlab/blob/master/repositories.go
- https://docs.gitlab.com/ee/api/repositories.html#list-repository-tree

### Do you want to implement this?

- [X] I would like to implement this myself 👷

Shocktrooper:
It appears `ref` is actually required and does not default to the default branch like the documentation says.  I will file a bug report later about this as omitting the ref results in a 404 according to my testing

PR #1198, opened 2022-08-04
Added code for Repository Tree Datasource

Shocktrooper:
References: #1197


PatrickRice-KSC:
@Shocktrooper - Ill take a look at it this weekend likely, unless timo beats me to it :)

Shocktrooper:
@PatrickRice-KSC Thank you 😄  this should hopefully be a quick one as there isn't much code to it

internal/provider/data_source_gitlab_repository_tree.go
```diff
@@ -0,0 +1,135 @@
...
+			"path": {
+				Description: "The path inside repository. Used to get content of subdirectories.",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			"ref": {
+				Description: "The name of a repository branch or tag.",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
```
timofurrer, line 35: NIT: Let's define the required attributes first:

```suggestion
			"ref": {
				Description: "The name of a repository branch or tag.",
				Type:        schema.TypeString,
				Required:    true,
			},
			"path": {
				Description: "The path inside repository. Used to get content of subdirectories.",
				Type:        schema.TypeString,
				Optional:    true,
			},
```

internal/provider/data_source_gitlab_repository_tree_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+func TestAccDataSourceGitlabRepositoryTree_basic(t *testing.T) {
+	testProject := testAccCreateProject(t)
+
+	resource.ParallelTest(t, resource.TestCase{
+		ProviderFactories: providerFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: fmt.Sprintf(`
+					resource "gitlab_repository_file" "foo" {
```
timofurrer, line 21: It would also be nice to just add some files in the `PreConfig` hook instead of having them in the tf config here.

timofurrer (changes requested):
Looks almost perfect - thanks 🎉  I've left a nitpick and it would be great if you could add an example file for the docs.

Out of curiosity: do you have a concrete use case for this data source?

internal/provider/data_source_gitlab_repository_tree_test.go
Shocktrooper: Should I just use an API call or create a new `testAccCreateProjectFile` function for this purpose

timofurrer: hmmm, let's go with the function. I think we can also use it in the tests for the `gitlab_repository_file` data source at one point :)

Shocktrooper:
> Looks almost perfect - thanks 🎉 I've left a nitpick and it would be great if you could add an example file for the docs.
> 
> Out of curiosity: do you have a concrete use case for this data source?

@timofurrer I do actually have an example of how I am going to use it.

So I have a centralized configuration project that you can think of as the brain of my DevOps toolchain.  It defines everything from SSO into GitLab groups to project/environment level IAM roles in AWS to project configuration in GitLab.  One of the problems I hope to solve with this is that we are scoping AWS IAM roles for projects and with that we are going to be allowing developers to define their own IAM deployment role policies vs passing the responsibility to the DevOps team.  In order to not allow all developers to see the sensitive centralized configuration project that will use said IAM role policies we will be putting them in a separate repo that developers can see and make PR's to and which can be gated before deployment with PR approvals.  With this data-source we will be reading from a project specific folder in the separate GitLab repo that will have IAM policies defined to then attach them to the deployed role.  This `gitlab_repository_tree` data-source will feed into an array of `gitlab_repository_file` data-sources to then feed into an array of `aws_iam_policy` resources which will then be attached to the `aws_iam_role` resource for the projects AWS environment.

internal/provider/data_source_gitlab_repository_tree_test.go
Shocktrooper: All done

internal/provider/data_source_gitlab_repository_tree.go
Shocktrooper: Reordered these and also did the same for the repo file resource/datasource

Shocktrooper:
Looks like there was an intermittent failure in the project resource not sure if you can re-run the test or if we can push it anyways with that failure pending the review is good of the code

timofurrer:
@Shocktrooper thanks for the explanation! 

I've re-triggered the test. It's related to #1039 .

timofurrer (approved):
LGTM 🎉

PR #1199, opened 2022-08-04
Fix golangci by moving away from deprecated io/ioutil

yan12125:
## Description

golangci-lint 1.42 on go 1.19 reports:
```
internal/provider/config.go:7:2: SA1019: "io/ioutil" has been deprecated since Go 1.16: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details. (staticcheck)
        "io/ioutil"
        ^
```
Not sure why it does not appear on GitHub CI.

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/internal/provider/config.go
+++ b/internal/provider/config.go
@@ -4,8 +4,8 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
-	"io/ioutil"
 	"net/http"
+	"os"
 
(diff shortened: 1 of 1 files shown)


timofurrer (approved):
Awesome! LGTM 🎉

Issue #1187, opened 2022-07-26
Question: is it possible to manage Gitlab admin settings?


iverberk:
I would like to use this provider to set a baseline configuration for Gitlab. For example to restrict the user project limit or disable sign-ups. Can it be done with the provider? I've scanned the issues and documentation but was unable to find a suitable resource.

timofurrer:
It's currently not possible. I assume you are referring to basically managing the [Application settings API](https://docs.gitlab.com/ee/api/settings.html) in a resource with this provider, right?

@JanKoppe worked on this quite some time ago, see https://github.com/gitlabhq/terraform-provider-gitlab/issues/957
Are there any updates @JanKoppe from your side? Are you still interested in landing this in the provider? Should someone else take this over?

filipowm:
@iverberk For the time being, I recommend you looking at [GitLab Configuration as Code](https://github.com/Roche/gitlab-configuration-as-code) (GCasC), which could do what you exactly expect, unfortunately outside of Terraform, but still easy to use (either manually or from within some CI/CD pipeline).

JanKoppe:
Hey, no updates from my side on #957 - anyone who wants to work on it: Feel free to pick up where I left off! This is way down on my backlog right now.

Thank you so much @filipowm, this pretty much looks like it could help us out with the original issue, which prompted me to start work on such a resource.

During the development of such a resource we noticed more and more that this does not really fit the model of Terraforms state mangagement very well, and it seems like a dedicated tool could be the way to go here. Running an additional tool in the CD pipeline is a compromise we can live with very well. Will give this a spin in the coming days.

timofurrer:
I'm closing this in favor of #957 and https://github.com/gitlabhq/terraform-provider-gitlab/pull/1201

timofurrer:
We've implemented a first experimental version of a `gitlab_application_settings` resource which will be released with [v.3.17.0](https://github.com/gitlabhq/terraform-provider-gitlab/milestone/17). We'd appreciate feedback!

Issue #957, opened 2022-03-14
Resource for GitLab instance (application) settings


JanKoppe:
### Feature Description

Hello,

I would like to be able to set the instance wide "Application" settings from Terraform, when provisioning a new GitLab instance with Terraform. Right now, there's no way to do that with this provider, as far as I can see.

I'm specifically talking about these settings that apply to self-hosted instances: https://docs.gitlab.com/ee/api/settings.html

I think it would fit very well into this provider and provide huge value. Currently, I'm trying to do this using null_resources and manual invocations of curl, which is quite buggy.

Do you agree that this should be placed in this provider? If so, I'd be willing to work on it and see if I can give an implementation for such a resource.

### Do you want to implement this?

- [X] I would like to implement this myself 👷

timofurrer:
> Do you agree that this should be placed in this provider? If so, I'd be willing to work on it and see if I can give an implementation for such a resource.

Yes, I think it would be nice.

However, I'm a little bit concerned that there is no actual *create* - as the resource can only exist once and kinda already exist outside of terraform. Thus, we need some dummy id and call the *update* API on *create*, too. We don't have such a resource yet - and this will most likely also be the only one ^^. 

Feel free to take a stab at it and I'll be happy to review!

JanKoppe:
> However, I'm a little bit concerned that there is no actual create - as the resource can only exist once and kinda already exist outside of terraform. Thus, we need some dummy id and call the update API on create, too. We don't have such a resource yet - and this will most likely also be the only one ^^.

Yes, that's quite an anti-pattern to what you're usually doing in Terraform. But I think there will be a solution possible - easiest way probably would be to generate an artificial ID out of connection parameters, or something like that.

> Feel free to take a stab at it and I'll be happy to review!

Great, will do so! Let's see if I can contribute a bit here. :)

timofurrer:
Sounds good. If you need any help implementing this, just ping me on the Discord server ;)

JanKoppe:
Short update: Progress is ongoing, datasource is coming along, but the underlying go-gitlab library needs a few changes as well, lots of API changes are not reflected in that lib yet.

Also, I'm struggling a bit with the difficulties of the settings not being a real resource - especially what to do on a "delete", or when a parameter is being removed.

My approach so far is to keep this very simple and analog to the GitLab API design with just a single resource type, that has parameters for all the possible settings options.

[...]

PR #1201, opened 2022-08-05
resource/gitlab_application_settings: New experimental resource

timofurrer:
This change set implements the new `gitlab_application_settings`
resource which provides a way to change the GitLab Application Settings
using the API at
https://docs.gitlab.com/ee/api/settings.html#change-application-settings.

Even though I suggest that we can merge this first experimental version
of the resource there are still quite some open questions for me.

I'd like to tackle them over at
https://github.com/gitlabhq/terraform-provider-gitlab/issues/957 and
give the community the change to take part in the discussion on how to
move forward with this.

Refs: #957 #1187


timofurrer:
@PatrickRice-KSC can you do me a favor and check if you can reproduce the CI failure locally?

PatrickRice-KSC:
Hey @timofurrer  - for sure. I tried this out on both GitPod (Failed, see first image below), and on a local ubuntu 21.04 vagrant machine (Failed, see second image below)

GitPod:
image

Local Vagrant:
image

Shocktrooper:
@timofurrer As I was looking through the code I don't think I saw anything specific but does there need to be anything done specifically to handle the setting that only appear once someone has an EE license vs a CE

https://docs.gitlab.com/ee/api/settings.html#:~:text=Users%20on%20GitLab%20Premium%20or%20Ultimate%20may%20also%20see%20these%20parameters%3A

timofurrer:
> @timofurrer As I was looking through the code I don't think I saw anything specific but does there need to be anything done specifically to handle the setting that only appear once someone has an EE license vs a CE
> 
> https://docs.gitlab.com/ee/api/settings.html#:~:text=Users%20on%20GitLab%20Premium%20or%20Ultimate%20may%20also%20see%20these%20parameters%3A

The fields which are unavailable because of the license are "just not working" - you may set the attribute in the config, but will get an upstream error message - you may also access the attribute, but it won't have a value in it. 

I think for a first version that's good enough. Actually, we do the same in many resources - which is only not really optimal.

PatrickRice-KSC:
@timofurrer  | @Shocktrooper  - Is this PR ready for review now? Happy to take a look yet today or tomorrow!

timofurrer:
@PatrickRice-KSC yes, it is :)

internal/provider/resource_gitlab_application_settings.go
```diff
@@ -0,0 +1,78 @@
...
+~> This is an **experimental resource**. By nature it doesn't properly fit into how Terraform resources are meant to work.
+   Feel free to join the [discussion](https://github.com/gitlabhq/terraform-provider-gitlab/issues/957) if you have any
+   ideas or questions regarding this resource.
+
+~> All ` + "`" + `gitlab_application_settings` + "`" + ` use the same ID ` + "`" + `gitlab` + "`" + `.
+
+!> This resource does not implement any destroy logic, it's a no-op at this point.
+   It's also not possible to revert to the previous settings.
+
+-> Requires at administrative privileges on GitLab.
```
PatrickRice-KSC, line 27: @timofurrer  - Seems obvious, but should we note that this doesn't work on the SaaS offering as a result just to be explicit? Something like this:

```suggestion
-> Requires at administrative privileges on a self-hosted GitLab instance.
```

internal/provider/schema_gitlab_application_settings.go
```diff
@@ -0,0 +1,3054 @@
...
+
+		"admin_mode": {
+			Description: "Require administrators to enable Admin Mode by re-authenticating for administrative tasks.",
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Computed:    true,
+		},
+
+		"admin_notification_email": {
+			Description: "Deprecated: Use abuse_notification_email instead. If set, abuse reports are sent to this address. Abuse reports are always available in the Admin Area.",
```
PatrickRice-KSC, line 19: @timofurrer  - I'm not sure I feel extremely strong about this, but it seems weird to me to be building a new resource and have immediately deprecated fields. I know it's aligning to GitLab's API, but I can't think of a use-case for using this field in a new config other than using it for a very old version of GitLab. 

WDYT?

internal/provider/schema_gitlab_application_settings.go
```diff
@@ -0,0 +1,3054 @@
...
+			Optional:    true,
+			Computed:    true,
+		},
+
+		"allow_local_requests_from_hooks_and_services": {
+			Description: "(Deprecated: Use allow_local_requests_from_web_hooks_and_services instead) Allow requests to the local network from hooks and services.",
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Computed:    true,
+		},
```
PatrickRice-KSC, line 73: Same comment as above.

internal/provider/schema_gitlab_application_settings.go
```diff
@@ -0,0 +1,3054 @@
...
+
+		"asset_proxy_whitelist": {
+			Description: "(Deprecated: Use asset_proxy_allowlist instead) Assets that match these domains are not proxied. Wildcards allowed. Your GitLab installation URL is automatically allowlisted. GitLab restart is required to apply changes.",
+			Type:        schema.TypeList,
+			Elem:        &schema.Schema{Type: schema.TypeString},
+			Optional:    true,
+			Computed:    true,
+		},
```
PatrickRice-KSC, line 124: Same comment as above

internal/provider/schema_gitlab_application_settings.go
```diff
@@ -0,0 +1,3054 @@
...
+
+		"default_branch_name": {
+			Description: "Instance-level custom initial branch name (introduced in GitLab 13.2).",
+			Type:        schema.TypeString,
+			Optional:    true,
+			Computed:    true,
+		},
+
+		"default_branch_protection": {
+			Description: "Determine if developers can push to the default branch. Can take: 0 (not protected, both users with the Developer role or Maintainer role can push new commits and force push), 1 (partially protected, users with the Developer role or Maintainer role can push new commits, but cannot force push) or 2 (fully protected, users with the Developer or Maintainer role cannot push new commits, but users with the Developer or Maintainer role can; no one can force push) as a parameter. Default is 2.",
```
PatrickRice-KSC, line 247: @timofurrer  - This one's a bit odd. It feels like we should align to our `valid values are:` syntax here, but the description of the integers is pretty important to the use of the resource here, so I feel like we should leave it the way it is.

WDYT?

internal/provider/schema_gitlab_application_settings.go
```diff
@@ -0,0 +1,3054 @@
...
+
+		"default_group_visibility": {
+			Description: "What visibility level new groups receive. Can take private, internal and public as a parameter. Default is private.",
+			Type:        schema.TypeString,
+			Optional:    true,
+			Computed:    true,
+		},
+
+		"default_project_creation": {
+			Description: "Default project creation protection. Can take: 0 (No one), 1 (Maintainers) or 2 (Developers + Maintainers)",
```
PatrickRice-KSC, line 268:
```suggestion
			Description: "Default project creation protection. Can take: 0 (No one), 1 (Maintainers) or 2 (Developers + Maintainers).",
```

internal/provider/schema_gitlab_application_settings.go
```diff
@@ -0,0 +1,3054 @@
...
+
+		"diff_max_lines": {
+			Description: "Maximum lines in a diff.",
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Computed:    true,
+		},
+
+		"disable_feed_token": {
+			Description: "Disable display of RSS/Atom and calendar feed tokens (introduced in GitLab 13.7)",
```
PatrickRice-KSC, line 345:
```suggestion
			Description: "Disable display of RSS/Atom and calendar feed tokens (introduced in GitLab 13.7).",
```

internal/provider/schema_gitlab_application_settings.go
```diff
@@ -0,0 +1,3054 @@
...
+
+		"external_pipeline_validation_service_url": {
+			Description: "URL to use for pipeline validation requests.",
+			Type:        schema.TypeString,
+			Optional:    true,
+			Computed:    true,
+		},
+
+		"external_pipeline_validation_service_token": {
+			Description: "Optional. Token to include as the X-Gitlab-Token header in requests to the URL in external_pipeline_validation_service_url.",
```
PatrickRice-KSC, line 666:
```suggestion
			Description: "Token to include as the X-Gitlab-Token header in requests to the URL in external_pipeline_validation_service_url.",
```

I don't think we need to have "Optional" in the description since it'll already be in the optional section of the resource due to `Optional: true` 🙂

internal/provider/schema_gitlab_application_settings.go
```diff
@@ -0,0 +1,3054 @@
...
+
+		"gravatar_enabled": {
+			Description: "Enable Gravatar.",
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Computed:    true,
+		},
+
+		"hashed_storage_enabled": {
+			Description: "Create new projects using hashed storage paths: Enable immutable, hash-based paths and repository names to store repositories on disk. This prevents repositories from having to be moved or renamed when the Project URL changes and may improve disk I/O performance. (Always enabled in GitLab versions 13.0 and later, configuration is scheduled for removal in 14.0)",
```
PatrickRice-KSC, line 757:
```suggestion
			Description: "Create new projects using hashed storage paths: Enable immutable, hash-based paths and repository names to store repositories on disk. This prevents repositories from having to be moved or renamed when the Project URL changes and may improve disk I/O performance. (Always enabled in GitLab versions 13.0 and later, configuration is scheduled for removal in 14.0).",
```

PatrickRice-KSC: Also, while this one isn't flagged explicitly as deprecated, 14.10 is _barely_ outside the range of what we've talked about supporting in the past. I'm good with leaving this one hear since folks are undoubtedly still using the provider for older GitLab versions, but feels like someone keeping their provider very up-to-date would also be keeping GitLab equally up-to-date.

Thoughts here?

internal/provider/schema_gitlab_application_settings.go
```diff
@@ -0,0 +1,3054 @@
...
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Computed:    true,
+		},
+
+		"user_email_lookup_limit": {
+			Description: "caution Deprecated in GitLab 14.9 will be removed in 15.0. Replaced by search_rate_limit. Max number of requests per minute for email lookup. Default: 60. To disable throttling set to 0.",
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Computed:    true,
```
PatrickRice-KSC, line 1226:
```suggestion
			Description: "Deprecated in GitLab 14.9, will be removed in 15.0. Replaced by search_rate_limit. Max number of requests per minute for email lookup. Default: 60. To disable throttling set to 0.",
			Type:        schema.TypeInt,
			Optional:    true,
			Computed:    true,
```

Align verbiage to other Deprecation notices. Also kinda same comment as deprecated resources outside the range of versions we've discussed supporting before (and this one has a replacement field)

internal/provider/schema_gitlab_application_settings.go
```diff
@@ -0,0 +1,3054 @@
...
+
+		"repository_checks_enabled": {
+			Description: "GitLab periodically runs git fsck in all project and wiki repositories to look for silent disk corruption issues.",
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Computed:    true,
+		},
+
+		"repository_size_limit": {
+			Description: "Size limit per repository (MB)",
```
PatrickRice-KSC, line 1281:
```suggestion
			Description: "Size limit per repository (MB).",
```

internal/provider/schema_gitlab_application_settings.go
```diff
@@ -0,0 +1,3054 @@
...
+
+		"shared_runners_text": {
+			Description: "Shared runners text.",
+			Type:        schema.TypeString,
+			Optional:    true,
+			Computed:    true,
+		},
+
+		"sidekiq_job_limiter_mode": {
+			Description: "track or compress. Sets the behavior for Sidekiq job size limits. Default: ‘compress’.",
```
PatrickRice-KSC, line 1368: @timofurrer  - I think we should use the `valid values are: ` syntax here.

internal/provider/schema_gitlab_application_settings.go
```diff
@@ -0,0 +1,3054 @@
...
+
+		"snippet_size_limit": {
+			Description: "Max snippet content size in bytes. Default: 52428800 Bytes (50MB).",
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Computed:    true,
+		},
+
+		"snowplow_app_id": {
+			Description: "The Snowplow site name / application ID. (for example, gitlab)",
```
PatrickRice-KSC, line 1447:
```suggestion
			Description: "The Snowplow site name / application ID (for example, gitlab).",
```

internal/provider/schema_gitlab_application_settings.go
```diff
@@ -0,0 +1,3054 @@
...
+			Type:        schema.TypeString,
+			Optional:    true,
+			Computed:    true,
+		},
+
+		"snowplow_collector_hostname": {
+			Description: "The Snowplow collector hostname. (for example, snowplow.trx.gitlab.net)",
```
PatrickRice-KSC, line 1454:
```suggestion
			Description: "The Snowplow collector hostname (for example, snowplow.trx.gitlab.net).",
```

internal/provider/schema_gitlab_application_settings.go
```diff
@@ -0,0 +1,3054 @@
...
+			Type:        schema.TypeString,
+			Optional:    true,
+			Computed:    true,
+		},
+
+		"snowplow_cookie_domain": {
+			Description: "The Snowplow cookie domain. (for example, .gitlab.com)",
```
PatrickRice-KSC, line 1461:
```suggestion
			Description: "The Snowplow cookie domain (for example, .gitlab.com).",
```

internal/provider/schema_gitlab_application_settings.go
```diff
@@ -0,0 +1,3054 @@
...
+			Optional:    true,
+			Computed:    true,
+		},
+
+		"throttle_unauthenticated_requests_per_period": {
+			Description: "(Deprecated in GitLab 14.3. Use throttle_unauthenticated_web_requests_per_period or throttle_unauthenticated_api_requests_per_period instead.) Max requests per period per IP.",
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Computed:    true,
+		},
```
PatrickRice-KSC, line 1620: Same comment about deprecations here.

PatrickRice-KSC (changes requested):
@timofurrer  That is a _huge_ PR, I'll probably need to take a second pass through it after you changes just to double check all the sets and everything in the schema file; my brain was getting a bit sketchy partway through reviewing it!

Lots of comments, but they really boil down to 2 topics: Do we really want to support already deprecated attributes that are almost a year deprecated in some cases, and just fixing minor grammar in descriptions.

Thanks!

Shocktrooper:
@PatrickRice-KSC I think those are left in because of the auto generation script. We should probably tweak that to leave those out of anything. An ignorelist persay

internal/provider/resource_gitlab_application_settings.go
timofurrer: Nice one 🎉

timofurrer:
> @PatrickRice-KSC I think those are left in because of the auto generation script. We should probably tweak that to leave those out of anything. An ignorelist persay

@Shocktrooper @PatrickRice-KSC Yes, that's exactly what happened here. Also the verbiage alignments and ending sentences with a dot etc are because they are not really aligned in the GitLab docs itself.
I think I'll apply your suggestions @PatrickRice-KSC and remove the deprecated attributes - let's see how often that endpoint really changes and has impacts for us. Maybe I'll add an OpenAPI spec for that endpoint and see what we can do to have it up-to-date in GitLab ...

PatrickRice-KSC:
@Shocktrooper  | @timofurrer  - whew, I'm glad most of that code was generated. I got about 1/3 of the way through the PR and thought you guys had written that by hand 😮‍💨 . That makes sense - thanks!

internal/provider/resource_gitlab_application_settings_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+package provider
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccGitlabApplicationSettings_basic(t *testing.T) {
```
Shocktrooper, line 12: What about adding an EE only test for the fields only available in premium and above.  One thing we also have to think about in the future is if we end up building out the acceptance tests some more unless we do it in a non-parallel fashion and revert all of the tested settings to their initial state after testing then those settings could potentially influence other acceptance tests

internal/provider/schema_gitlab_application_settings.go
timofurrer: I've removed all the deprecated fields from the resource.

internal/provider/schema_gitlab_application_settings.go
timofurrer: Yes, let's leave it the way it is.

internal/provider/schema_gitlab_application_settings.go
timofurrer: Yes, while I agree and since this is generated - I propose to leave it as is. It would be a huge effort to streamline the entire descriptions.

internal/provider/schema_gitlab_application_settings.go
timofurrer: Yes, again - while I agree I'd still leave it in for now and remove once it's gone upstream - just to remove my effort here a little bit :)

internal/provider/schema_gitlab_application_settings.go
timofurrer: Since this is generated I'd highly prefer to not do this unless someone really asks for this. Again, it would be too much effort to streamline them all ...

internal/provider/resource_gitlab_application_settings_test.go
timofurrer: Yes, you are right - I think for other resources it makes a lot of sense and is easy to do. Given that this resource is kinda experimental and I don't really know where this is going or how much maintenance burden it is I'd like to keep it as-is.

However, I completely agree that we should think about what do to if we change settings here or other resources which affect the entire instance ...

PatrickRice-KSC:
@timofurrer  - Makes sense to me, given it's experimental I think we can align to standards a bit more in a next push. I can go through and resolve threads and take another look at it here in a bit to merge.

PatrickRice-KSC (approved):

PR #1204, opened 2022-08-07
Implement `gitlab_project_hook` and `gitlab_project_hooks` data sources

timofurrer:
This change set adds the `gitlab_project_hook` and `gitlab_project_hooks` data sources.


internal/provider/helper_test.go
```diff
@@ -322,6 +322,28 @@ func testAccAddProjectMembers(t *testing.T, pid interface{}, users []*gitlab.Use
...
+		})
+		if err != nil {
+			t.Fatalf("could not create project hook: %v", err)
+		}
+		t.Cleanup(func() {
+			_, err := testGitlabClient.Projects.DeleteProjectHook(pid, hook.ID)
+			if err != nil {
+				t.Fatalf("could not cleanup project hook: %v", err)
+			}
+		})
```
PatrickRice-KSC, line 341: @timofurrer  - Do we have a weird potential race condition here with the executed order of the cleanup scripts? In most of our project-bound resources in the `helper_test` class we don't clean them up because we allow the deletion of the project to clean them up (see protected branches, project members, etc); This makes sense because it could cause an ordering issue if they're trying to clean them up after the project was already deleted. the `t.Fatalf` block would be hit here if the project was cleaned up before the hooks were, right?

Seems like in this case we'd be OK to leave off the cleanup method specifically because they'd be cleaned up when the project was.

internal/provider/schema_gitlab_project_hook.go
```diff
@@ -0,0 +1,143 @@
...
+			Type:        schema.TypeString,
+			Required:    true,
+		},
+		"project_id": {
+			Description: "The id of the project for the hook.",
+			Type:        schema.TypeInt,
+			Computed:    true,
+		},
+		"hook_id": {
+			Description: "The id of the project hook",
```
PatrickRice-KSC, line 21:
```suggestion
			Description: "The id of the project hook.",
```

internal/provider/schema_gitlab_project_hook.go
```diff
@@ -0,0 +1,143 @@
...
+		"project": {
+			Description: "The name or id of the project to add the hook to.",
+			Type:        schema.TypeString,
+			Required:    true,
+		},
+		"project_id": {
+			Description: "The id of the project for the hook.",
+			Type:        schema.TypeInt,
+			Computed:    true,
+		},
```
PatrickRice-KSC, line 19: @timofurrer  - We seem to have gained `project_id` as part of migrating from the resource -> schema. If this just an attempt to always have the numeric value available for downstream resources as a result of something like #1165 ?

PatrickRice-KSC (changes requested):

internal/provider/schema_gitlab_project_hook.go
timofurrer: Yes, exactly. And it's actually returned as `project_id` by the upstream API, too.

internal/provider/helper_test.go
timofurrer: Yes, indeed - I've removed it. We actually we have the same potential issue in other helpers ...

PatrickRice-KSC: Yeah, I saw a couple there with the same issue. I may open a PR to see about closing that up, even though it seems to be pretty rare.

PatrickRice-KSC (approved):

PR #1205, opened 2022-08-07
Use `crazy-max/ghaction-import-gpg` instead of the one from hashicorp

timofurrer:
HashiCorp's `ghaction-import-gpg` is deprecated, see
https://github.com/hashicorp/ghaction-import-gpg/issues/11#issuecomment-1185107410.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,7 @@
 # This GitHub action can publish assets for release when a tag is created.
 # Currently its setup to run on any tag that matches the pattern "v*" (ie. v0.1.0).
 #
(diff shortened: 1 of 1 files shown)


timofurrer:
This also aligns with the scaffolding project: https://github.com/hashicorp/terraform-provider-scaffolding/blob/main/.github/workflows/release.yml#L36

PatrickRice-KSC (approved):
@timofurrer  - Interesting that hashi is using a GHA module that doesn't seem to have security scanning in it's pipeline. I did a quick overview of the source and there doesn't seem to be anything untoward (at least not that's obvious to me), and it clearly has a massive community so it's very low risk.

PR #1206, opened 2022-08-08
Fix GitLab external URL configuration

timofurrer:
This will fix the acceptance testing instance to have a proper external
URL configured. With that redirects within GitLab will work (e.g. from
the WebIDE back to the repository) and stuff like clone URLs etc.

### DIFF
--- a/scripts/gitlab.rb
+++ b/scripts/gitlab.rb
@@ -1,3 +1,6 @@
+external_url "http://127.0.0.1:8080"
+nginx['listen_port'] = 80
+
 pages_external_url 'http://127.0.0.1:5051'
 pages_nginx['redirect_http_to_https'] = false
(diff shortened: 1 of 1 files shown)


timofurrer:
This change seems to have the drawback that GitLab cannot reach a repository it hosts with the external URL. We are currently having this project in the `TestAccGitlabProject_importURL` which tries to import a repository hosted on the same GitLab instance.

```
Error: POST http://127.0.0.1:8080/api/v4/projects: 422 {message: http://127.0.0.1:8080/root/base-137441643613142598.git is not a valid HTTP Git repository}
```

One possible solution might be to change the nginx port to the same as the port used for the container port forwarding. Let's see ...

timofurrer:
I'm currently having some issues with getting this ready - since this is not urgent - I'm going to close for now and will re-open at a later time ;)

PR #1209, opened 2022-08-10
resource/gitlab_group_membership: Support removal options

timofurrer:
This change will support the two options when removing a user from a
group using `gitlab_group_membership`. These are
`skip_subresources_on_destroy` and `unassign_issuables_on_destroy`.
Both default to `false` and the change is backwards compatible.

Closes: #1194

/cc @1oglop1


timofurrer:
@PatrickRice-KSC do you mind giving this a review? 😊

PatrickRice-KSC (approved):

Issue #792, opened 2022-01-23
How do handle cases where GitLab implicitly creates a resource resulting in tf ownership confusion


timofurrer:
There has been a discussion in #266 about what to do in the following situation:

```hcl
resource "gitlab_group" "group" {
}

resource "gitlab_group_membership" "membership" {
  user_id = <ID of the user owning the GITLAB_TOKEN>
}
```

... in which GitLab implicitly adds group membership to the `gitlab_goup.group` for the user owning the token which was used to create the group. This leads to a `group membership already exists` error for the `gitlab_group_membership.membership` resource.

The following questions need to be answered:

- [ ] Is this behavior a bug in the gitlab terraform provider?
- [ ] If yes, how should the `gitlab_group_membership` resource behave?

armsnyder:
This relates to #749. Generally, what do we want to do when GitLab creates an upstream resource out-of-band that would map onto another Terraform resource?

armsnyder:
The main issue that a lot of folks are facing right now has to do with the default branch protection when you create a GitLab project, but this is a problem in a lot of situations like the `gitlab_group_membership` case, where creating a group also automatically adds the requesting user as a member.

Let's brainstorm some options. I can think of 2:

1. Strictly follow the GitLab API. Allow GitLab to set its defaults like group members and branch protections, and require Terraform users to manually import those resources.
   - Pros: Adheres to GitLab API, so less risk in dealing with breaking changes to the provider that diverge from the GitLab API in the future.
   - Cons: Bad Terraform user experience, requiring multiple applies and imports.
2. Support new attributes on these resources `gitlab_project` and `gitlab_group` that let the user opt-into behavior that would allow them to manage these resources. For example `gitlab_project` could have an attribute `disable_branch_protection` which defaults to `false`. When set, we would have logic in the `Create` function to explicitly remove the branch protection created by GitLab for a project initialized with a branch and README.
   - Pros: Better Terraform user experience, allowing them to apply config in a single pass.
   - Cons: More tech debt, diverges from GitLab API

I'm swaying toward option 2 personally. Thoughts?

timofurrer:
I don't see any additional options than these two as well and I agree with your pros and cons. I'm also wondering if other providers have the same issue - I'll do some research in the next few days.

[...]

PR #1216, opened 2022-08-17
resource/gitlab_branch_protection: Allow to manage branch protection for default branch

timofurrer:
This change to the `gitlab_branch_protection` resource will allow to
protect the default branch of a project without importing. It basically
works by unprotecting and protecting the default branch again in case it
already exists when the resource is created. This allows the following
code to properly work:

```hcl
    name = "example"
    initialize_with_readme = true
}

resource "gitlab_branch_protection" "default_branch" {
    branch = gitlab_project.example.default_branch

    // ... whatever settings ...
}
```

The drawback of this is that it quickly unprotects the default branch to
protect it again in the above scenario. However, it already does that
for pretty much any update to the resource attributes, because of the
lack of a proper upstream update API. In addition, it's just a special
case and maintenance effort - but I guess that's what docs are for and the
suppose that the benefits for the users outweigh the maintenance burden.

If tried to implement something along the lines of https://github.com/gitlabhq/terraform-provider-gitlab/issues/792#issuecomment-1029148287
but it turns out it's harder to implement than expected.

The problems are:

* lack of diff / plan customization on the schema: https://discuss.hashicorp.com/t/ignore-the-order-of-a-complex-typed-list/42242
* when implementing the possibility to protect branches in the
  `gitlab_project` resource we'd also need to allow to share the
  projects with groups in add members in the resource, because those
  groups or users maybe used in the branch protection. There again, we
  need better plan / diff control then what the provider sdk provides us
  at the moment.
* it basically would mean an integration / duplication / re-use of the
  `gitlab_project_share_group`, `gitlab_project_membership` and
  `gitlab_branch_protection` in the `gitlab_project` resource. Which I
  guess would be okay with the proper implementation, but then again the
  currently proposed "solution" in this PR may be good enough.

Refs: #792
Closes: #671

@PatrickRice-KSC @armsnyder @nagyv @Shocktrooper What are your thoughts here?

### DIFF
--- a/docs/resources/branch_protection.md
+++ b/docs/resources/branch_protection.md
@@ -4,6 +4,13 @@ page_title: "gitlab_branch_protection Resource - terraform-provider-gitlab"
 subcategory: ""
 description: |-
   The gitlab_branch_protection resource allows to manage the lifecycle of a protected branch of a repository.
+  ~> Branch Protection Behavior for the default branch
+     Depending on the GitLab instance, group or project setting the default branch of a project is created automatically by GitLab behind the scenes.
+     Due to some https://github.com/gitlabhq/terraform-provider-gitlab/issues/792 limitations https://discuss.hashicorp.com/t/ignore-the-order-of-a-complex-typed-list/42242 in the Terraform Provider SDK and the GitLab API,
+     when creating a new project and trying to manage the branch protection setting for its default branch the gitlab_branch_protection resource will
+     automatically take ownership of the default branch without an explicit import by unprotecting and properly protecting it again.
+     Having multiple gitlab_branch_protection resources for the same project and default branch will result in them overriding each other - make sure to only have a single one.
+     This behavior might change in the future.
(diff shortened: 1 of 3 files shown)


PatrickRice-KSC:
@timofurrer  - I think that this approach makes a lot of sense; it lives within the expectations of the current terraform SDK, makes managing the default branch more terraform-native (without the need for a local-exec), and is relatively logical for how it works.

I do wonder if the limitations you mention in the https://github.com/gitlabhq/terraform-provider-gitlab/issues/792#issuecomment-1029148287 thread are why we see things like the AWS S3 terraform provider explicitly going _away_ from nested blocks and towards these individual resources, and whether that's something that we'll need to emulate going forward. It feels cleaner to have them nested under the project resource, but doesn't seem to work very well within the SDK.

PatrickRice-KSC (approved):

davidfischer-ch:
Thank you :)

PR #1228, opened 2022-08-24
Enable dummy SAML provider to allow saml group link features

timofurrer:
This change configures a dummy SAML provider so that we can properly test the saml group link resource contributed with #1215 .

image

image

### DIFF
--- a/scripts/gitlab.rb
+++ b/scripts/gitlab.rb
@@ -18,3 +18,21 @@ gitlab_rails['application_settings_cache_seconds'] = 0
 gitlab_rails['env'] = {
     'IN_MEMORY_APPLICATION_SETTINGS' => 'false'
 }
+
+# Enable SAML authentication for GitLab (required for SAML group links).
+# see https://docs.gitlab.com/ee/integration/saml.html
+gitlab_rails['omniauth_allow_single_sign_on'] = ['saml']
+gitlab_rails['omniauth_block_auto_created_users'] = false
+gitlab_rails['omniauth_providers'] = [
+  {
+    name: "saml",
+    label: "Dummy Test Provider",
+    args: {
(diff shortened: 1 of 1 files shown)


scripts/gitlab.rb
```diff
@@ -1,20 +1,38 @@
...
+gitlab_rails['omniauth_allow_single_sign_on'] = ['saml']
+gitlab_rails['omniauth_block_auto_created_users'] = false
+gitlab_rails['omniauth_providers'] = [
+  {
+    name: "saml",
+    label: "Dummy Test Provider",
+    args: {
+      assertion_consumer_service_url: "http://127.0.0.1/users/auth/saml/callback",
+      idp_cert_fingerprint: "aa:aa:aa:aa:aa:aa:aa:aa:aa:aa:aa:aa:aa:aa:aa:aa",
+      idp_sso_target_url: "https://dummy.provider.com/sso/saml",
```
PatrickRice-KSC, line 33:
```suggestion
      idp_sso_target_url: "https://dummy.example.com/sso/saml",
```

Can we use example.com here since it's set aside as a non-real site for this kind of usage? "provider.com" is a real site (as outdated as it is 🙂 )

PatrickRice-KSC (changes requested):

scripts/gitlab.rb
timofurrer: Oh yeah, good catch! 🎉

timofurrer: I've changed all the URLs - cause they don't really matter anyways.

PatrickRice-KSC (approved):

PatrickRice-KSC:
@timofurrer  - like we discussed on discord, approved but not merged. Feel free to merge when you're ready 🙂

PR #1231, opened 2022-08-26
resource: fix typo in error message

Nhqml:
## Description

Fix typo in the error message of the import command for `gitlab_group_label`. The doc states [here](https://registry.terraform.io/providers/gitlabhq/gitlab/latest/docs/resources/group_label#import) that it should be `terraform import gitlab_group_label.example <group ID>:<label name>` but the error message says `terraform import gitlab_group_label.example <group ID>.<label name>`

- [X] I acknowledge that all of the following items are true, where applicable:
    - Create the resource
    - Update the attributes
    - Import the resource


timofurrer (approved):
Good catch @Nhqml and thank you 🎉

Issue #1133, opened 2022-06-15
Implementation of Group Membership Lock


rejackson73:
### Feature Description

The current GitLab provider for Terraform does not support the premium features for the Group resource, most notably the ability to "Prevent adding new members to projects within this group." The GitLab API does support this control utilizing the "membership_lock" boolean but it doesn't appear to be implemented in the provider.

### Do you want to implement this?

- [ ] I would like to implement this myself 👷

timofurrer:
Thanks for the request!

This feature is an upstream API alignment in the `gitlab_group` resource.

* [Upstream docs](https://docs.gitlab.com/ee/api/groups.html#:~:text=or%20public.-,membership_lock,Prevent%20adding%20new%20members%20to%20projects%20within%20this%20group.,-extra_shared_runners_minutes_limit)
* `go-gitlab` already supports it.

timofurrer:
Actually @jtymes implemented this in the GitLab REST API for the provider, see [this MR](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/82271#:~:text=This%20MR%20exposes%20the%20membership_lock%20field%20so%20this%20can%20be%20implemented%20as%20part%20of%20the%20GitLab%20Terraform%20provider.) :)

jtymes:
Yeah it was on my list to add here, but life has been hectic so it would be good if someone else wants it.

PR #1232, opened 2022-08-27
Align `gitlab_group` with missing attributes for create

timofurrer:
This adds a few missing attributes from the create API to the `gitlab_group` resource. There are still some missing ones which are only supported in the update API - which we can add later.

This is mainly to satisfy the request from #1133 😊 

Closes: #1133


jtymes (approved):

alexjfisher:
@jtymes You approved this, but it was never merged and now the bot has closed it.  Is there any reason it couldn't be reopened and merged?  Thanks.

PR #1243, opened 2022-09-05
resource/gitlab_group_saml_link: Add new resource

timofurrer:
Continuation of #1215


examples/resources/gitlab_group_saml_link/resource.tf
```diff
@@ -0,0 +1,5 @@
+resource "gitlab_group_saml_link" "test" {
+  group_id        = "12345"
+  access_level    = "Developer"
```
PatrickRice-KSC, line 3: @timofurrer  I think we should make the example lowercase, since we explicitly have the `stringInSlice` function set to case sensitive:

```suggestion
  access_level    = "developer"
```

PatrickRice-KSC (changes requested):
One small change on the example, then we're good to go!

PR #1252, opened 2022-09-18
Add data source gitlab_groups

l-with:
## Description

This PR solves issue #1257 .

I added a data source gitlab_groups equivalent to gitlab_users. This is useful to retrieve all groups of an gitlab instance.

    - Create the resource
    - Update the attributes
    - Import the resource


l-with:
@timofurrer Thank you very much for your attention! Please be so kind to approve the waiting workflows.

l-with:
The PR was added to milestone v3.19.0. I received no comments.
The PR passed all checks.
Please tell me, if there is anything I can do.

internal/provider/data_source_gitlab_groups_test.go
```diff
@@ -0,0 +1,130 @@
...
+func TestAccDataSourceGitlabGroups_basic(t *testing.T) {
+	rInt := acctest.RandInt()
+	rInt2 := acctest.RandInt()
+
+	resource.ParallelTest(t, resource.TestCase{
+		ProviderFactories: providerFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccDataSourceGitlabGroupsConfig(rInt, rInt2),
+				Check: resource.ComposeTestCheckFunc(
```
armsnyder, line 23: For our provider tests, we perform setup prior to calling `resource.ParallelTest()`, to isolate the resource being tested. Example: https://github.com/gitlabhq/terraform-provider-gitlab/blob/2f06bac4d85ce66bbffdd6b9dc60d319782166be/internal/provider/resource_gitlab_project_approval_rule_test.go#L30

Could you make this change please? This should drastically reduce the number of steps and amount of TF config you need.

armsnyder (changes requested):
Thanks for contributing! The code looks good to me. Just one note on the tests.

internal/provider/data_source_gitlab_groups_test.go
l-with: Thank you for the suggestion. 
I need to create groups with specific names for testing the search of the data_source. 
With groups created by testAccCreateGroups, it is only possible to search for "acctest-group". The result contains also groups from parallel running tests and thus the number of results can not be verified.
To fix this I implemented a helper function testAccCreateGroupsWithPrefix and refactored the test with this function.

I hope that matches your intention.

internal/provider/data_source_gitlab_groups_test.go
```diff
@@ -0,0 +1,56 @@
...
+	testAccCreateGroupsWithPrefix(t, 42, prefixLotsOf)
+
+	resource.ParallelTest(t, resource.TestCase{
+		ProviderFactories: providerFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccDataSourceGitlabGroupsConfigSearchSort(prefixFoo),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("data.gitlab_groups.foos", "groups.#", "2"),
+					resource.TestCheckResourceAttr("data.gitlab_groups.foos", "groups.0.name", groupsFoo[0].Name),
```
armsnyder, line 27: Could you validate the group attributes here as well?

armsnyder (changes requested):

internal/provider/data_source_gitlab_groups_test.go
armsnyder: Thanks

internal/provider/data_source_gitlab_groups_test.go
l-with: I committed validating the group attributes.

armsnyder: Thanks, it looks good. 👍 Could you make sure the CI workflow is passing? There is a guide in the CONTRIBUTING.md, but tl;dr you can run `make reviewable` from the console.

l-with: I committed the change resulting from "make reviewable".
I also run "make testacc", but cannot run "make testacc SERVICE=gitlab-ee".

armsnyder (dismissed):

timofurrer (dismissed):
LGTM 🚢

timofurrer (approved):

PR #1254, opened 2022-09-20
Typofix: drop extra "lab"

demon:
## Description

GitLabLab isn't a thing, unless there's a dog who contributes.

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/docs/resources/application_settings.md
+++ b/docs/resources/application_settings.md
@@ -3,7 +3,7 @@
 page_title: "gitlab_application_settings Resource - terraform-provider-gitlab"
 subcategory: ""
 description: |-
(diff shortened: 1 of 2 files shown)


timofurrer (commented):
Good catch, Thanks 🎉 

This file is generated from the source file at 

https://github.com/gitlabhq/terraform-provider-gitlab/blob/2aa1c94e80d23d2e668330ea7b0e9402cc2edc09/internal/provider/resource_gitlab_application_settings.go#L16

by running `make reviewable` (you can use [GitPod](https://gitpod.io/from-referrer/), if you don't have the setup locally) - so that needs to be changed as well :)

timofurrer (approved):

PR #1268, opened 2022-10-02
fix(docs/pipeline_schedule_variable): Add missing resorce name in example

ondrejsika:
## Description

Just a small fix in docs example

### DIFF
--- a/docs/resources/pipeline_schedule_variable.md
+++ b/docs/resources/pipeline_schedule_variable.md
@@ -24,8 +24,8 @@ resource "gitlab_pipeline_schedule" "example" {
 }
 
 resource "gitlab_pipeline_schedule_variable" "example" {
-  project              = gitlab_pipeline_schedule.project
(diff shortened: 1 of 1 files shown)


docs/resources/pipeline_schedule_variable.md
```diff
@@ -24,8 +24,8 @@ resource "gitlab_pipeline_schedule" "example" {
 }
 
 resource "gitlab_pipeline_schedule_variable" "example" {
```
timofurrer, line 26: This documentation example originates at:

https://github.com/gitlabhq/terraform-provider-gitlab/blob/ee81e298d6d33ebbbaa762e2624920fbf6022f22/examples/resources/gitlab_pipeline_schedule_variable/resource.tf#L9

Do you mind changing it there and run `make reviewable` or `make generate` to re-generate this doc file here?

The PR looks good otherwise 🎉

timofurrer (changes requested):

PR #1276, opened 2022-10-11
Remove name field from update project options since it is not required

### LINKED ISSUE #1262: Maintainers cannot update projects
### GitLab Provider version

3.18.0

### GitLab version

GitLab Enterprise Edition 15.3.3-ee

### Terraform version

Terraform v1.3.0 on linux_amd64

_No response_

### Relevant log output

```shell
│ Error: PUT https://gitlab.example.com/api/v4/projects/2305: 403 {message: 403 Forbidden}
│ 
│   with gitlab_project.main,
│   on main.tf line 26, in resource "gitlab_project" "main":
│   26: resource "gitlab_project" "main" {
```

### Description

So I've reported [this bug](https://gitlab.com/gitlab-org/gitlab/-/issues/375033) (or at least I believe it to be a bug) on GitLab. It prevents project Maintainers to use the Edit Project API endpoint to update the project Name, even though Maintainers can do it on the GUI.

While this bug is not addressed, I'm facing an issue using the gitlab provider, because on [line 1266 of the gitlab project resource](https://github.com/gitlabhq/terraform-provider-gitlab/blob/main/internal/provider/resource_gitlab_project.go#L1266), the provider always passes the project name, no matter what the field the user is trying to change. 
In my specific case, I'm trying to change the `merge_requests_template` field of a project (which maintainers can do without problem, unlike the name field), but since we are passing the `name` everytime, I'm getting a `403 Forbidden`.

I think if we didn't pass in the name everytime, it would make the provider work correctly while the Gitlab issue is not fixed. I think it would be enough to just make sure that at least one field is present.

jcoelho93:
## Description

Closes #1262 
There is a [bug in GitLab](https://gitlab.com/gitlab-org/gitlab/-/issues/375033) that prevents project Maintainers to use this provider. Furthermore according to [GitLab documentation](https://docs.gitlab.com/ee/api/projects.html#edit-project), the name field is not required. There should be no use case where the Edit Project API is called without one of the fields required.

With this PR, maintainers would still not be able to update the name of a project, but they would be able to update other fields. Since the bug on GitLab only affects the name field

    - Create the resource
    - Update the attributes
    - Import the resource

### DIFF
--- a/internal/provider/resource_gitlab_project.go
+++ b/internal/provider/resource_gitlab_project.go
@@ -1331,10 +1331,12 @@ func resourceGitlabProjectUpdate(ctx context.Context, d *schema.ResourceData, me
 
 	if d.HasChange("approvals_before_merge") {
 		options.ApprovalsBeforeMerge = gitlab.Int(d.Get("approvals_before_merge").(int))
+		options.Name = gitlab.String(d.Get("name").(string))
 	}
 
 	if d.HasChange("wiki_enabled") {
 		options.WikiEnabled = gitlab.Bool(d.Get("wiki_enabled").(bool))
+		options.Name = gitlab.String(d.Get("name").(string))
 	}
 
 	if d.HasChange("snippets_enabled") {
@@ -1382,11 +1384,13 @@ func resourceGitlabProjectUpdate(ctx context.Context, d *schema.ResourceData, me
 	if d.HasChange("mirror") {
 		options.ImportURL = gitlab.String(d.Get("import_url").(string))
 		options.Mirror = gitlab.Bool(d.Get("mirror").(bool))
+		options.Name = gitlab.String(d.Get("name").(string))
 	}
 
 	if d.HasChange("mirror_trigger_builds") {
 		options.ImportURL = gitlab.String(d.Get("import_url").(string))
 		options.MirrorTriggerBuilds = gitlab.Bool(d.Get("mirror_trigger_builds").(bool))
+		options.Name = gitlab.String(d.Get("name").(string))
 	}
 
 	if d.HasChange("only_mirror_protected_branches") {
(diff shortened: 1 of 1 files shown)


internal/provider/resource_gitlab_project.go
```diff
@@ -1259,12 +1259,7 @@ func resourceGitlabProjectRead(ctx context.Context, d *schema.ResourceData, meta
 func resourceGitlabProjectUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client := meta.(*gitlab.Client)
 
-	// Always send the name field, to satisfy the requirement of having one
-	// of the project attributes listed below in the update call
-	// https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/lib/api/helpers/projects_helpers.rb#L120-188
-	options := &gitlab.EditProjectOptions{
-		Name: gitlab.String(d.Get("name").(string)),
-	}
+	options := &gitlab.EditProjectOptions{}
```
timofurrer, line 1262: Just "removing" it would probably lead to this bug again #1157 . WDYT? And it would also disallow for owners to change the name.

jcoelho93: You're right #1157 might happen again with this. But I don't see how owners would be disallowed to change the name, the field is still sent in line 1266. 

According to [this](https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/lib/api/helpers/projects_helpers.rb#L120-188), the fields that have to be passed with at least the `name` field are the following:
```
['only_allow_merge_if_all_status_checks_passed', 'approvals_before_merge', 'external_authorization_classification_label', 'import_url', 'issues_template', 'mirror_overwrites_diverged_branches', 'mirror_trigger_builds', 'mirror_user_id', 'mirror', 'only_mirror_protected_branches', 'requirements_access_level', 'wiki_enable']
```

What if we check if the request only has these fields, and in that case we add the name field explicitely?

jcoelho93: I've pushed a new commit with this change. What do you think? @timofurrer

jcoelho93:
Closing this PR in favor of !1294

PR #1280, opened 2022-10-16
Add group subgroups data source

### LINKED ISSUE #1177: Group sub-groups, descendent groups, and projects, data sources
### Feature Description

We would like to do some traversing of Gitlab sub-groups with Terraform. In order to be able to do this we would like to see  `gitlab_group_subgroups`, `gitlab_group_descendent`, and `gitlab_group_project` calls added to the provider.

API calls are available for these and they would allow us to manage other resources associated with the Gitlab groups.

- [ ] I would like to implement this myself 👷

jcoelho93:
## Description

Partially fixes #1177 
Adds the data source `gitlab_group_subgroups`

    - Create the resource
    - Update the attributes
    - Import the resource


internal/provider/resource_gitlab_project.go
```diff
@@ -1269,12 +1269,7 @@ func resourceGitlabProjectRead(ctx context.Context, d *schema.ResourceData, meta
 func resourceGitlabProjectUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client := meta.(*gitlab.Client)
 
-	// Always send the name field, to satisfy the requirement of having one
```
timofurrer, line 1272: 🚧 **Suggestion:** Can you please remove these changes in the project resource from this PR?

internal/provider/schema_gitlab_group.go
```diff
@@ -0,0 +1,162 @@
+package provider
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func gitlabGroupSchema() map[string]*schema.Schema {
```
timofurrer, line 7: ❓ **Question:**  what do you think about making use of this in the other group resources / data sources, too?

timofurrer (changes requested):
Looks almost ready to merge 🎉 I've left some comments - back to you 🏓

internal/provider/resource_gitlab_project.go
jcoelho93: Ups, mix up with branches. Removing

jcoelho93: Removed

internal/provider/schema_gitlab_group.go
jcoelho93: I think it makes sense, but we should do that on a different PR, so we don't increase the scope of this one. wdyt?

jcoelho93: @timofurrer

timofurrer:
Could you please rebase onto the latest main (see #1304)?

